PR #31, opened 2014-08-09
add patch for preserving js loading order

yathit:
This is patch for https://code.google.com/p/js-test-driver/issues/detail?id=409&q=order#makechanges 

Config file order is not respected is file contains "Object" in the filename


segrey:
It's already fixed in https://youtrack.jetbrains.com/issue/WEB-12963. This pull request can be closed.

PR #95, opened 2014-12-17
Basic Dart template macros (className, methodName, methodParams).

pq:
Basic macro building blocks for live templates (https://youtrack.jetbrains.com/issue/WEB-14603).

For example, with these in place, the following could be used to print method params:

`groovyScript("'\"' + _1.collect { it + ' = [\" + ' + it + ' + \"]'}.join(', ') + '\"'", methodParameters())`

Printing method name and class name are equally easy.


pq:
Manually merged.

PR #100, opened 2014-12-30
IDEA-134890: added support for hook annotations

SierraGolf:
ticket: https://youtrack.jetbrains.com/issue/IDEA-134890

this pull request adds support for cucumber hooks (```@Before```, ```@After```) for cucumber version 1.0.x, to 1.2.x. previously hooks defined in separate files outside of the step package weren't recognized by the logic which creates run configurations. now hooks are supported for running a single scenario, a single feature and multiple features.

I am not so much familiar with writing idea plugins. Actually it was pretty tough just to get everything up and running, so I could debug properly. I am unsure how to test my changes programmatically, so if anyone could help out I would be quite thankful.


SierraGolf:
this PR has been implemented by https://github.com/JetBrains/intellij-plugins/commit/6d91fc16a9e63b1f48d6ec677e10352c6c519d3f.

@avokin: btw. it is quite rude to copy/paste major parts from a pull request without any attributions or conversations with the original contributor.

PR #108, opened 2015-01-30
Replaced the deprecated format options with --plugin, which is the way t...

lazee:
The old --format configuration for Cucumber is deprecated a long time ago. We should change this plugin to use --plugin instead. I have used pretty here since that looks the best. You might want to consider keeping --monochrome as the default since it is only IntelliJ IDEA 14.x that supports colors. That needs to be tested. 

The optimal solution would be if this plugin could be configured in preferences ("Select output format", "Enable color output").

It is possible to override the program arguments manually for every run, but that is not a good solution. 

So, I hope you will consider fixing this.

NB! I haven't run the tests, since I'm not sure how that is done for a plugin.

### DIFF
--- a/cucumber-java/src/org/jetbrains/plugins/cucumber/java/run/CucumberJavaRunConfigurationProducer.java
+++ b/cucumber-java/src/org/jetbrains/plugins/cucumber/java/run/CucumberJavaRunConfigurationProducer.java
@@ -24,7 +24,7 @@ import org.jetbrains.annotations.Nullable;
  * @since 8/6/12
  */
 public abstract class CucumberJavaRunConfigurationProducer extends JavaRunConfigurationProducerBase<CucumberJavaRunConfiguration> implements Cloneable {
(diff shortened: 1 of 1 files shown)


lazee:
The way it will look with the new configuration

https://www.dropbox.com/s/61m0vwjyjpxet0t/Screenshot%202015-01-30%2009.43.23.png?dl=0

lazee:
With the plugin defaults:

https://www.dropbox.com/s/o82731uagh1udcs/Screenshot%202015-01-30%2009.44.04.png?dl=0

lazee:
Turns out the tests run but that I get an error in the IDE. I will investigate and create an alternative formatter.

lazee:
Closing PR since no-one at JetBrains boughter commenting.

avokin:
@lazee, sorry I missed the PR. By the way, the problem was solved in issue https://youtrack.jetbrains.com/issue/IDEA-135862

PR #124, opened 2015-02-17
Switch syntax is not automatically correctly indented

kasperpeulen:
see: https://youtrack.jetbrains.com/issue/WEB-15292


alexander-doroshko:
I fixed it in a bit different way and added tests. To find problems in your code run DartTypingTest against your version of DartBlock.

PR #129, opened 2015-02-21
WEB-15293 `case` and `on` are not suggested for code completion 

kasperpeulen:
https://youtrack.jetbrains.com/issue/WEB-15293

also added default and finally, and made sure that there is class name completion after `on`


alexander-doroshko:
Closing as code completion is now powered by the Dart Analysis Server

PR #132, opened 2015-02-24
WEB-11243: TypeHandler duplicates existing symbols

kasperpeulen:
see: https://youtrack.jetbrains.com/issue/WEB-11243

this is how it works now: http://i.imgur.com/5j6BA65.gif


alexander-doroshko:
Thank you for your pull request! I pushed the fix very similar to yours in some places, though not literally based on the pull request.

PR #133, opened 2015-02-24
WEB-15350 completion popup shows private variables

kasperpeulen:
https://youtrack.jetbrains.com/issue/WEB-15350


kasperpeulen:
this one doesn't pass testReference9, but this new pull request does: https://github.com/JetBrains/intellij-plugins/pull/138

PR #147, opened 2015-03-12
Dart: Add Shebang Detection

kaendfinger:
This adds shebang detection for Dart.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartFileType.java
+++ b/Dart/src/com/jetbrains/lang/dart/DartFileType.java
@@ -2,6 +2,7 @@ package com.jetbrains.lang.dart;
 
 import com.intellij.openapi.fileTypes.LanguageFileType;
 import com.intellij.openapi.fileTypes.ex.FileTypeIdentifiableByVirtualFile;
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Sorry I didn't notice the pull request earlier. It is recommended to file an issue in YouTrack together with the pull request, so everybody will be able to track the issue status and finally it will go to the release notes.
Unfortunately I have to reject this pull request. FileTypeIdentifiableByVirtualFile#isMyFileType must work instantly, it is called millions of times! And loading a file is an extremely costly operation.
By the way what is the use case for the problem you are trying to solve? Will it be enough for you to add your project-specific file names (or patterns) to the Settings (Preferences) | Editor | File Types | Dart?

kaendfinger:
Would I not have to add a file specific file name for every single script I have without a .dart extension? This is the problem I am trying to solve. I wish I could make this faster :(

nicity:
You can implement DartFileTypeDetector, similar to e.g.
GroovyHashBangFileTypeDetector

On Tue, Mar 17, 2015 at 8:10 PM, Kenneth Endfinger <notifications@github.com

kaendfinger:
Ok

PR #149, opened 2015-03-15
Dart Analysis Server usage bug fix- priority files can only be set on files in the analysis roots

jwren:
By adding files only called via the annotator, and by always removing them, if added, when the a file is closed, the correct set of priority files will always be in sync with the Dart Analysis Server.


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerAnnotator.java
```diff
@@ -91,6 +91,8 @@ public AnnotatorInfo collectInformation(@NotNull final PsiFile psiFile) {
 
     if (!DartAnalysisServerService.getInstance().serverReadyForRequest(module.getProject(), sdk)) return null;
 
+    DartAnalysisServerService.getInstance().virtualFileOpened(annotatedFile);
```
alexander-doroshko: I also added a flag not to prioritize files in case of batch inspection run (DartAnalysisServerGlobalInspectionTool), otherwise they never get out of the priority list.

PR #151, opened 2015-03-19
Change no-problems message in Dart Problems view, for @sethladd

jwren:


sethladd:
:+1:

alexander-doroshko:
:)
At first the message is shown for everyone for a pretty long time while analysis is in progress. I'm afraid there are users who wouldn't like the teaser before getting thousands of errors. I think we can set initial text to 'No problems found in Dart code yet' and if analysis is really done without errors - '...Happy day!'. Though when we implement filters to hide errors/warning/hints, the fact that user applied filters not always makes his day happy :)

jwren:
I don't have a preference, and don't disagree, but being playful is fun too.

sethladd:
Playful is good. (see: inbox when you have inbox zero :)

Agree that you don't want to show "Happy day" until analysis is complete. But when it's complete and there are no issues, say "Happy day!"

alexander-doroshko:
> when it's complete and there are no issues, say "Happy day!"
No objections here.
Though when we implement filters and View will be without errors&warnings because user simply filtered them out - do not show 'happy day' as well.

PR #156, opened 2015-03-30
Terminate loop over the Dart Analysis Server query getFixes, if it is timing out.

jwren:


alexander-doroshko:
We do not have to terminate the loop, we have a list of errors in hands and can show them in the editor. But we can stop asking for quick fixes as soon as some request timed out. I pushed it.

PR #176, opened 2015-04-24
[Dart] [WEB-15735] Resolve import prefixes to their exact declarations.

scheglov:


alexander-doroshko:
I'm closing it since we are switching to DAS driven reference resolution. By the way DAS has the same problem: both 'prefix' references are resolved to the same (first) prefix declaration.

PR #177, opened 2015-05-04
Reformat ReadMe for readability

stevemessick:
Added a bunch of newlines to make it easier to read.


alexander-doroshko:
Readme.txt was easy to read on [GitHub](https://github.com/JetBrains/intellij-plugins/tree/master/Dart) that splitted the test in lines itself.
Pushed. Need to make sure that it still looks nice on GitHub.
Probably the best solution would be to change readme to MD format. And add instructions for IntelliJ Community + Dart plugin setup.

PR #179, opened 2015-05-15
[WEB-16445] Fix for creating files in Quick Fixes.

scheglov:


Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -167,7 +169,9 @@ public boolean isAvailable(@NotNull final Project project, final Editor editor,
     final VirtualFile virtualFile = LocalFileSystem.getInstance().findFileByPath(FileUtil.toSystemIndependentName(fileEdit.getFile()));
     final ProjectFileIndex fileIndex = ProjectRootManager.getInstance(project).getFileIndex();
 
-    if (virtualFile == null || !fileIndex.isInContent(virtualFile)) return false;
+    if (fileEdit.getFileStamp() != -1) {
```
alexander-doroshko: Docs say that -1 means that a file should be created. But using nightly SDK 1.11.0-edge.161458 I see that fileEdit.getFileStamp() returns 0.

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -180,9 +184,23 @@ public boolean startInWriteAction() {
   @Override
   public void invoke(@NotNull final Project project, final Editor editor, final PsiFile file) throws IncorrectOperationException {
     final SourceFileEdit fileEdit = myChange.getEdits().get(0);
+    final String filePath = fileEdit.getFile();
```
alexander-doroshko: I get filePath="package:ProjectName/non_existing_file.dart". Shouldn't there be an absolute path instead?

alexander-doroshko: Also SourceChange.message="Create file 'package:ProjectName/non_existing_file.dart'". For me message like "Create file 'non_existing_file.dart'" looks better.
The later is true for creating file not in lib folder: current message contains full file path like "Create file 'C:\dart_projects\DartSample2\test\a\foopart.dart'". Wouldn't "Create file 'foopart.dart'" be better?

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
alexander-doroshko: Well, if file should be created in lib folder then fileEdit.getFileStamp() == 0, if in any other folder then it returns -1 as expected.

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -180,9 +184,23 @@ public boolean startInWriteAction() {
...
     final SourceEdit sourceEdit = fileEdit.getEdits().get(0);
 
-    final VirtualFile virtualFile = LocalFileSystem.getInstance().findFileByPath(FileUtil.toSystemIndependentName(fileEdit.getFile()));
+    // Create the file if it does not exist.
+    if (fileEdit.getFileStamp() == -1) {
+      try {
+        final File ioFile = new File(filePath);
```
alexander-doroshko: Do not use java.io.File when working with files that belong to the project. Such change is treated as external, undo doesn't work. Use VfsUtil.createDirectoryIfMissing and Virtualfile.createChildData().

alexander-doroshko:
General comment: please add a test to DartServerQuickFixTest. 
This test can be run only locally: remove abstract modifier, create corresponding JUnit run configuration and add -Ddart.sdk=path/to/your/local/dart-sdk to its VM Options.

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
scheglov: You're right.
This is another bug in the server.
Fixing it now.
https://codereview.chromium.org/1135073004

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
scheglov: Yes, probably.
We still provide full names if the client wants to show it.
https://codereview.chromium.org/1147733002

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
scheglov: Done.

alexander-doroshko: I'm not seeing the updated revision.

scheglov:
Updated.
The test does not work.
PTAL

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -1,5 +1,6 @@
 package com.jetbrains.lang.dart.validation.fixes;
 
+import com.google.common.io.Files;
```
alexander-doroshko: Unused import. Fixed.

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -180,9 +186,24 @@ public boolean startInWriteAction() {
...
   public void invoke(@NotNull final Project project, final Editor editor, final PsiFile file) throws IncorrectOperationException {
     final SourceFileEdit fileEdit = myChange.getEdits().get(0);
+    final String filePath = fileEdit.getFile();
     final SourceEdit sourceEdit = fileEdit.getEdits().get(0);
 
-    final VirtualFile virtualFile = LocalFileSystem.getInstance().findFileByPath(FileUtil.toSystemIndependentName(fileEdit.getFile()));
+    // Create the file if it does not exist.
+    if (fileEdit.getFileStamp() == -1) {
+      try {
+        final String directoryPath = VfsUtil.getParentDir(filePath);
```
alexander-doroshko: Will fail on Windows because of different slashes. Slash tolerant methods: PathUtil.getParentPath, PathUtil.getFileName. Fixed

Dart/src/com/jetbrains/lang/dart/validation/fixes/DartServerFixIntention.java
```diff
@@ -180,9 +186,24 @@ public boolean startInWriteAction() {
...
+        final VirtualFile directory = VfsUtil.createDirectoryIfMissing(directoryPath);
+        if (directory != null) {
+          final String fileName = VfsUtil.extractFileName(filePath);
+          directory.createChildData(this, fileName);
+        }
+      }
+      catch (IOException e) {
```
alexander-doroshko: I added CommonRefactoringUtil.showErrorHint to show red hint in case of failure

PR #181, opened 2015-05-18
Clarify dependency in ReadMe.txt

stevemessick:
@alexander-doroshko Minor edit to ReadMe.txt

### DIFF
--- a/Dart/ReadMe.txt
+++ b/Dart/ReadMe.txt
@@ -25,8 +25,8 @@ Prerequisites:
 
 4. Open intellij-community project, compile it.
    Open File | Project Structure | Modules | [+] | Import Module, select intellij-plugins/Dart/Dart-community.iml.
-   In the list, select community-main then click [+] at the bottom of the list to
-   add a dependency on the Dart-community module.
+   In the same Project Structure dialog open the Dependencies tab of the community-main module,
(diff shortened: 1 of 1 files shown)


Dart/ReadMe.txt
```diff
@@ -25,7 +25,8 @@ Prerequisites:
 
 4. Open intellij-community project, compile it.
    Open File | Project Structure | Modules | [+] | Import Module, select intellij-plugins/Dart/Dart-community.iml.
-   Open community-main module, add dependency on Dart module.
+   In the list, select community-main then click [+] at the bottom of the list to
+   add a dependency on the Dart-community module.
```
alexander-doroshko: There are a lot of [+] buttons. I'd specify that it is in the Dependencies tab. Different location on Mac (bottom) and Windows (right).
How about this:
In the same Project Structure dialog open Dependencies tab of the community-main module, click [+] to add a module dependency on the Dart-community module.

stevemessick: OK, I'm pushing another update which specifically calls out the presence of
two [+] buttons in that dialog and (hopefully) clarifies which to use.

stevemessick:
@alexander-doroshko PTAL I think I've addressed all our issues.

PR #183, opened 2015-05-18
Turn on Dart

stevemessick:
@alexander-doroshko Does it make sense to change this setting in the repos?

### DIFF
--- a/Dart/Dart-community.iml
+++ b/Dart/Dart-community.iml
@@ -144,5 +144,6 @@
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="library" name="Dart SDK" level="application" />
   </component>
 </module>
--- a/Dart/ReadMe.txt
+++ b/Dart/ReadMe.txt
@@ -4,8 +4,8 @@ Prerequisites:
 - Oracle JDK 1.6 or later (1.6 is preferred as we still must be compatible with its API)
 - Git command line client
 - IntelliJ IDEA Ultimate installed. The following bundled plugins are enabled:
-  UI Designer, Git Integration, I18n for Java, IntelliLang, JUnit, Plugin DevKit, Properties Support.
-  Additionally install 'Grammar-Kit' and 'Polymer & Web Components' plugins:
+  Git Integration, I18n for Java, IntelliLang, JUnit, Plugin DevKit, Properties Support, UI Designer.
+  Additionally install 'Dart', 'Grammar-Kit', and 'Polymer & Web Components' plugins:
   start IntelliJ IDEA, on Welcome screen click Configure | Plugins.
   Press 'Browse repositories...', find plugins and install them. Restart IDE.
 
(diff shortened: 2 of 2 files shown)


alexander-doroshko:
What for is it?
Although there are some Dart test data files in this module, it is still a Java module with all logic written in Java.
I guess you want to easily start dart_style original tests right from the test data. But note that this addition will require everyone setting up a project to make one more step: setup a Dart SDK before opening the IntelliJ Community + Dart plugin project.

stevemessick:
Does this affect everyone in the world, or everyone working on the Dart plugin? If the latter, then they have to set up the Dart SDK anyway. If the former, then we should either find a less obtrusive way to do this or delete the PR.

stevemessick:
I want to eliminate this line:
screen shot 2015-05-18 at 12 02 19 pm

I also think it would be nice to have syntax highlighting, like for html.

alexander-doroshko:
We have 3 IML files: Dart.iml affects everyone in JetBrains, Dart-plugin.iml affects only those who work with a single-module Dart-plugin project, Dart-community affects only those who work with IntelliJ Community + Dart Plugin project. So this change is safe (and requires readme update), but it raises the question: why do not we keep 3 IML files in sync.
You already have syntax highlighting even without Dart SDK enabled. Note that a lot of Dart test data files are syntactically incorrect so you'll never get them green.

stevemessick:
@alexander-doroshko PTAL I added some instructions to the readme about the sdk.
Also copied up the optional step of turning on internal dev tools.
I'll leave the issue of managing three iml files to those who know what they are doing.

alexander-doroshko:
Step about turning on internal dev tools is ok.
Installing Dart plugin to get syntax highlighting is ok too.
Still not sure about the need of enabling Dart support. Here are my thoughts.
Pro: No yellow bar when Dart test data file is open in the editor. Is there anything else? 
Con: The biggest thing IMO is that enabling Dart support enables Dart Analysis Server. So we get Dart Analysis tool window and other stuff from server. But the module doesn't contain valid Dart projects and most of Dart test data files are syntactically incorrect. What's the need to bother server with looking at invalid files? Tons of irrelevant errors and warnings in Dart Analysis tool window and in test data. Tons of errors instead of a yellow bar - is it a solution?
For me yellow bar doesn't seem to be a problem. This is not the place where I edit valid Dart projects.

stevemessick:
>The biggest thing IMO is that enabling Dart support enables Dart Analysis Server
I don't understand. Why does that matter?

If we don't include this PR then intellij-plugins/Dart/Dart_community.iml gets modified when the Dart plugin is enabled. If there is a way to remove that yellow bar without enabling Dart then I can leave Dart disabled.

If I were going to use IDEA to edit Dart code I would certainly want the plugin enabled.

alexander-doroshko:
Enabling Dart support is good when you work on Dart projects. Dart-community module doesn't contain Dart projects. It contains Dart files in test data, most of Dart files are not valid and syntactically incorrect - what's the benefit of having Dart support enabled for them?
At the moment there's no other way to remove yellow bar. I live with it for 1.5 years and it doesn't bother me as I feel Dart files in Dart plugin module just like an auxiliary test data, not as a primary code to concentrate on.
Yes, Dart-community.iml gets modified when the Dart support is enabled. You may keep it in a separate local CL like '~never commit'.
Sorry, I still do not see benefits of enabling Dart support for the Dart plugin module. I do not need Dart Analyzer to smell my test data. We may collect opinions of other teammates.

PR #185, opened 2015-05-19
Dart plugin: underline the files in the Project Structure when there are errors reported from the Dart Analysis Server

jwren:


alexander-doroshko:
The task seems to be not so trivial.
First thing that I noticed is that all red waives disappear as soon as you type any symbol in any file. Breakpoint in WolfTheProblemSolverImpl#doRemove helped to see that on any typing all red files are cleared and scheduled to be reanalyzed by the IDE HighlightingPass. And that's the main problem: on each typing all red files in the project are reanalyzed using IDE engine. This is not needed completely for us as we have server results ready, and this may take a lot of resources.
Unless we find a way of simple red waves managing we have to live without this feature.

PR #186, opened 2015-05-19
Fmt run

stevemessick:
@alexander-doroshko As you thought, very few tests pass. Only three files, actually. This gives us something to work toward, so I won't add it to the standard set of Dart tests yet.

### DIFF
--- a/Dart/testSrc/com/jetbrains/lang/dart/dart_style/DartStyleTest.java
+++ b/Dart/testSrc/com/jetbrains/lang/dart/dart_style/DartStyleTest.java
@@ -1,5 +1,6 @@
 package com.jetbrains.lang.dart.dart_style;
 
+import com.intellij.openapi.util.TextRange;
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Great, pushed!
Additionally I reformatted the code (Cmd+Alt+L) and fixed warnings reported by the IDE.
Also I made clickable diff view to show results for all failed subtests, not only the first one.
Finally I disabled the test case by adding an 'abstract' keyword. Otherwise it runs as a part of 'Dart tests' run configuration and also on build server (we do not want expected failures there).

May I ask you to be a bit more verbose in commit messages. People who screen daily activity in the repository get zero information seeing 'update', 'clean up' or even 'Make tests explicit'. They have to see affected files list or even go down to individual changes. At minimum commit message should clearly say what subsystem is affected. 'Cleanup DartStyleTest' is better. 'Added more test cases to DartStyleTest (comments and regression test data folders)' is perfect.

PR #187, opened 2015-05-20
Formatname

stevemessick:
@alexander-doroshko I'd like to see the path to the test file in the output. Some of the names are ambiguous and it is easier to find the original file when it's path is visible in the test output.

Is there really no way to keep this from running as part of the system tests other than making it abstract? Maybe I forgot to mention it but I prefer not to modify files unnecessarily. AFAIK I can't run the dart_style test as-is without modifying it; is there another way to run it?

### DIFF
--- a/Dart/testSrc/com/jetbrains/lang/dart/dart_style/DartStyleTest.java
+++ b/Dart/testSrc/com/jetbrains/lang/dart/dart_style/DartStyleTest.java
@@ -14,7 +14,7 @@ import java.io.File;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public abstract class DartStyleTest extends FormatterTestCase {
+public  class DartStyleTest extends FormatterTestCase {
 
   protected String getFileExtension() {
     return DartFileType.DEFAULT_EXTENSION;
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
It is possible to add an annotation at class or method level like this:
@Bombed(year = 2015, month = Calendar.NOVEMBER, day = 12, user = "Steve Messick")
With it corresponding tests won't be run on a build server. But still will be run locally when you run 'Dart tests' run configuration.
It is possible to modify 'Dart tests' run configuration to be based on pattern instead of 'all tests in module'.

stevemessick:
The Bombed annotation still doesn't do what I want: I want a test file that
runs when I specifically run it, and does not run when I run the project's
acceptance tests.

>It is possible to modify 'Dart tests' run configuration to be based
>on pattern instead of 'all tests in module'.

I'm not sure what this means, but it sounds promising. Or maybe not; if
everyone else likes having all tests run automatically then it might be a
bad idea.

alexander-doroshko:
Patterns look promising, but I'm not sure they support negation. Without it it is too boring to list all other tests in the module.
One more thought: may be just hardcode KNOWN_TO_FAIL_SUBTESTS right int the DartStyleTest, so that they are skipped out-of-the-box? We will be fixing one by one and removing them from the list.

stevemessick:
What's the syntax for KNOWN_TO_FAIL_SUBTESTS? I can't find any examples of
it (global via google search or locally via idea search and grep).

Is there a built-in tool similar to grep that searches all local intellij
files (including java sources in jar files)?

alexander-doroshko:
Sorry, it was just a suggestion of a constant name that we could introduce ourselves in the DartStyleTest and handle it ourselves in the main loop of the runTestInDirectory() method

alexander-doroshko:
if (KNOWN_TO_FAIL_SUBTESTS.contains(currentSubtestName)) continue;

stevemessick:
Haha! OK, I understand now.

Remember: everything I know about IDEA I learned from you. :)

PR #189, opened 2015-05-21
Change Dart cascade parser to build flatter tree

stevemessick:
Change cascade parser to build flat tree with VALUE_EXPRESSION as head, receiver as first child, and CASCADE_REFERENCE_EXPRESSIONs as subsequent children representing messages sent to the receiver.

Most of this was straight-forward. There's work to be done in DartSpacingProcessor (see TODO), which will bring it into compliance with dart_style.

@alexander-doroshko Perhaps now I can fix that bug I started to look at a few months ago. It will still require some work in DartBlock, but at least it is bounded now.

### DIFF
--- a/Dart/gen com/jetbrains/lang/dart/psi/impl/DartValueExpressionImpl.java
+++ b/Dart/gen com/jetbrains/lang/dart/psi/impl/DartValueExpressionImpl.java
@@ -11,21 +11,21 @@ import static com.jetbrains.lang.dart.DartTokenTypes.*;
 import com.jetbrains.lang.dart.psi.*;
 import com.jetbrains.lang.dart.util.DartPsiImplUtil;
 
-public class DartCascadeReferenceExpressionImpl extends DartReferenceImpl implements DartCascadeReferenceExpression {
+public class DartValueExpressionImpl extends DartExpressionImpl implements DartValueExpression {
 
-  public DartCascadeReferenceExpressionImpl(ASTNode node) {
+  public DartValueExpressionImpl(ASTNode node) {
     super(node);
   }
 
   public void accept(@NotNull PsiElementVisitor visitor) {
-    if (visitor instanceof DartVisitor) ((DartVisitor)visitor).visitCascadeReferenceExpression(this);
+    if (visitor instanceof DartVisitor) ((DartVisitor)visitor).visitValueExpression(this);
     else super.accept(visitor);
   }
 
   @Override
-  @Nullable
-  public DartExpression getExpression() {
-    return findChildByClass(DartExpression.class);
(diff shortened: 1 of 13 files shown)


Dart/gen com/jetbrains/lang/dart/psi/impl/DartValueExpressionImpl.java
```diff
@@ -0,0 +1,31 @@
+// This is a generated file. Not intended for manual editing.
```
alexander-doroshko: file in 'get com' folder? :)

Dart/src/com/jetbrains/lang/dart/psi/impl/DartReferenceImpl.java
```diff
@@ -140,9 +140,12 @@ public DartClassResolveResult resolveDartClass() {
       }
     }
     if (this instanceof DartCascadeReferenceExpression) {
-      DartReference[] children = PsiTreeUtil.getChildrenOfType(this, DartReference.class);
-      if (children != null && children.length == 2) {
-        return children[0].resolveDartClass();
+      PsiElement parent = this.getParent();
+      if (parent instanceof DartValueExpression) {
+        DartReference child = (DartReference) parent.getFirstChild();
```
alexander-doroshko: Dangerous unchecked cast.
In general, getFirstChild() may be null (not here of course), whitespace, comment (unlikely here, but still), throwExpression (which is not a DartReference).
Where possible try to use more specific methods (DartValueExpression.getExpressionList()) in this case instead of general ones like PsiElement.getFirstChild().

Dart/src/com/jetbrains/lang/dart/util/DartResolveUtil.java
```diff
@@ -468,12 +468,14 @@ public static DartReference getLeftReference(@Nullable final PsiElement node) {
...
-      final PsiElement parent = reference.getParent();
+      PsiElement parent = reference.getParent();
       if (parent instanceof DartCascadeReferenceExpression) {
-        DartReference[] references = PsiTreeUtil.getChildrenOfType(parent, DartReference.class);
-        DartReference result = references != null && references.length == 2 ? references[0] : null;
-        // if not starts with node
-        return !PsiTreeUtil.isAncestor(result, node, false) ? result : null;
+        parent = parent.getParent();
+        if (parent instanceof DartValueExpression) {
+          return (DartReference) parent.getFirstChild();
```
alexander-doroshko: the same: use DartValueExpression.getExpressionList() and check first entry type

alexander-doroshko:
Pushed with corresponding follow-up fixes

PR #191, opened 2015-05-26

stevemessick:
@alexander-doroshko This fixes some of the worst problems with cascade formatting. Nested cascades are now indented as they would by dartfmt. If multiple cascades call the same method they will be kept on the same line. At least for now, if they are already on separate lines the lines will not be combined, unlike dartfmt.

Also, adjacent strings are now formatted as dartfmt does it. I haven't done "prefix" + "suffix" yet. I need to change some default settings first.

I sent you an earlier PR that can be ignored. This supersedes it.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
@@ -3,14 +3,19 @@ package com.jetbrains.lang.dart.ide.formatter;
 import com.intellij.formatting.*;
 import com.intellij.formatting.templateLanguages.BlockWithParent;
 import com.intellij.lang.ASTNode;
+import com.intellij.openapi.util.Key;
+import com.intellij.psi.PsiElement;
 import com.intellij.psi.codeStyle.CodeStyleSettings;
 import com.intellij.psi.formatter.FormatterUtil;
 import com.intellij.psi.formatter.common.AbstractBlock;
 import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.tree.TokenSet;
+import com.intellij.psi.util.CachedValue;
 import com.jetbrains.lang.dart.DartFileType;
 import com.jetbrains.lang.dart.DartLanguage;
 import com.jetbrains.lang.dart.DartTokenTypes;
+import com.jetbrains.lang.dart.psi.DartCascadeReferenceExpression;
+import com.jetbrains.lang.dart.psi.DartValueExpression;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -30,6 +35,8 @@ public class DartBlock extends AbstractBlock implements BlockWithParent {
   private static final TokenSet LAST_TOKENS_IN_SWITCH_CASE = TokenSet.create(
     BREAK_STATEMENT, CONTINUE_STATEMENT, RETURN_STATEMENT
   );
+  private static final Key<DartBlock> DART_RECEIVER_KEY = Key.create("DART_RECEIVER_KEY");
+  private static final Key<DartBlock> DART_CASCADE_KEY = Key.create("DART_CASCADE_KEY");
 
   private final DartIndentProcessor myIndentProcessor;
   private final DartSpacingProcessor mySpacingProcessor;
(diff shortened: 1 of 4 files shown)


Dart/testData/dart_style/whitespace/cascades.stmt
```diff
@@ -15,6 +15,10 @@ list
   ..add("bar");
 <<<
 list..add("baz")..add("bar");
+>>> multiple cascades get the same line when the method names are the same
```
alexander-doroshko: This change makes test data different from what is there in dart_style original repo. It makes it difficult to port new changes from that repo.
I suggest to keep intellij-plugins/Dart/testData/dart_style 100% equivalent to dart_style/test and to update tests of DartFormatterTest instead.

Also as a rule all fixes are committed with working tests. So if we keep DartStyleTest disabled we need to add the case to DartFormatterTest (or update any its existing test data).
Other option is to enable DartStyleTest, but only those cases that pass. So with each next fix we'll enable more and more cases and some lucky day the test will fully pass.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
```diff
@@ -3,14 +3,19 @@
 import com.intellij.formatting.*;
 import com.intellij.formatting.templateLanguages.BlockWithParent;
 import com.intellij.lang.ASTNode;
+import com.intellij.openapi.util.Key;
+import com.intellij.psi.PsiElement;
```
alexander-doroshko: A bunch of unused imports. Before commit please make sure that your code doesn't add new warnings highlighted by IDE or warnings are justified.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
```diff
@@ -184,4 +221,16 @@ public BlockWithParent getParent() {
   public void setParent(BlockWithParent newParent) {
     myParent = newParent;
   }
-}
\ No newline at end of file
+
+  private static class DartCascadeBlock extends DartBlock {
```
alexander-doroshko: What is the purpose of these 2 classes?

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
```diff
@@ -70,11 +77,41 @@ public Spacing getSpacing(Block child1, @NotNull Block child2) {
...
+          receiverBlock = childBlock;
+          childBlock.setParent(this);
+        }
+        tlChildren.add(childBlock);
+      }
+    } else {
+      if (parent.getElementType() == CASCADE_REFERENCE_EXPRESSION) {
+        // Swizzle tree structure to make this block a child of the receiver block.
+        DartBlock receiver = parent.getUserData(DART_RECEIVER_KEY);
+        DartBlock cascade = parent.getUserData(DART_CASCADE_KEY);
```
alexander-doroshko: It seems to me that cascade==this after this line is executed, isn't it?

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
```diff
@@ -30,6 +35,8 @@
   private static final TokenSet LAST_TOKENS_IN_SWITCH_CASE = TokenSet.create(
     BREAK_STATEMENT, CONTINUE_STATEMENT, RETURN_STATEMENT
   );
+  private static final Key<DartBlock> DART_RECEIVER_KEY = Key.create("DART_RECEIVER_KEY");
```
alexander-doroshko: "user data" is a universal mechanism that is used when there are no other means to accomplish the task. User data driven logic is harder to read and understand, and also it is usually error-prone.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
```diff
@@ -70,11 +77,41 @@ public Spacing getSpacing(Block child1, @NotNull Block child2) {
       return EMPTY;
     }
     final ArrayList<Block> tlChildren = new ArrayList<Block>();
-    for (ASTNode childNode = getNode().getFirstChildNode(); childNode != null; childNode = childNode.getTreeNext()) {
-      if (FormatterUtil.containsWhiteSpacesOnly(childNode)) continue;
-      final DartBlock childBlock = new DartBlock(childNode, createChildWrap(childNode), createChildAlignment(childNode), mySettings);
-      childBlock.setParent(this);
-      tlChildren.add(childBlock);
+    ASTNode parent = getNode();
+    if (parent.getElementType() == VALUE_EXPRESSION) {
```
alexander-doroshko: I'm afraid I didn't get what this code do. Isn't DartBlock tree good enough after your changes in parser? How this code modifies DartBlock tree compared to what was here before?

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartIndentProcessor.java
```diff
@@ -125,9 +125,12 @@ public Indent getChildIndent(final ASTNode node) {
         (prevSiblingType == RPAREN || (prevSiblingType == ELSE && elementType != IF_STATEMENT))) {
       return Indent.getNormalIndent();
     }
-    if (prevSiblingType != DOT && elementType == DOT_DOT && parentType == CASCADE_REFERENCE_EXPRESSION) {
+    if (elementType == CASCADE_REFERENCE_EXPRESSION) {
       return Indent.getNormalIndent();
     }
+    if (elementType == OPEN_QUOTE && prevSiblingType == CLOSING_QUOTE && parentType == STRING_LITERAL_EXPRESSION) {
```
alexander-doroshko: Please update some existing DartFormatterTest to cover this case

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -404,9 +409,47 @@ private Spacing setBraceSpace(boolean needSpaceSetting,
...
+          if (name1.equals(name2)) {
+            return true;
+          }
+        }
+      }
+    }
+    return false;
+  }
+
+  private String getImmediateCallName(ASTNode callNode) {
```
alexander-doroshko: Just to make sure that we have the same project setup: does IDE show warning for you on this method name?

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
stevemessick: Makes debugging easier.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
stevemessick: It was intended to move the cascades to enable formatting. It turns out that it isn't even needed.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
stevemessick: I haven't trained myself to recognize warnings yet, I guess. There is one, I hadn't noticed it, and I don't even agree that what it is saying is good practice. But I'll change it anyway.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
alexander-doroshko: For easier debugging we usually override toString(), in this case in DartBlock class. Adding empty inheritor for this purpose is too heavy solution :)

stevemessick: Obviously, I disagree, but I'm not trying to teach you good programming style. It's a moot point since the code has been deleted.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
alexander-doroshko: I track warnings in the following way:
- Colored square in the editor top right shows file status. When it's yellow I try to make sure that warnings are justified. It is possible to suppress justified warning to make inspection skip it and show green file status
- Yellow stripes on the right gutter
- The fastest way to navigate between warnings is F2 / Shift+F2

Speaking about this warning I browsed internet a bit and noticed that most people recommend to make helper methods static. Special highlighting in IDE allows to distinguish static / dynamic methods calls. Seeing a static invocation I'm sure that this method doesn't need to read and won't change this instance state. One of more detailed answers [here](http://stackoverflow.com/questions/3346764/should-java-methods-be-static-by-default).

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartBlock.java
alexander-doroshko: As I understand the question is not in programming style but in the ease of debugging.
IDE provides a lot of ways to meet the goal without changing code:
- watches
- conditional breakpoints
- log evaluated expression instead of stopping on a breakpoint
- it shows toString() as a default object presentation
- the most advanced feature is 'Java Data Type Renderers' (in Preferences): it allows to show any expression instead of toString() as a default object presentation. For example it is possible to configure that for any object of DartBlock type debugger will show getNode().getElementType().toString() + ": " + getNode().getText()

PR #192, opened 2015-05-27

stevemessick:
@alexander-doroshko This fixes some of the worst problems with cascade formatting. Nested cascades are now indented as they would by dartfmt. If multiple cascades call the same method they will be kept on the same line. At least for now, if they are already on separate lines the lines will not be combined, unlike dartfmt.

Also, adjacent strings are now formatted as dartfmt does it. Default settings now allow line wrapping. Binary expressions are wrapped and indented to the continuation level. I added a test for ternary expression alignment but kept it disabled since it isn't working yet.

I added a couple elements to the settings dialog. They were already being used by the formatter, I think, but couldn't be modified by the user.


Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -371,14 +349,41 @@ else if (type1 == COLON) {
...
 
     if (elementType == VALUE_EXPRESSION && type2 == CASCADE_REFERENCE_EXPRESSION) {
-      //if (type1 == CASCADE_REFERENCE_EXPRESSION) {
-      //  if (((AbstractBlock)child1).getNode().getLastChildNode().getElementType() == CALL_EXPRESSION) {
-      //    // TODO if child1 message name == child2 message name, do not add line break
-      //  }
-      //}
+      if (type1 == CASCADE_REFERENCE_EXPRESSION) {
+        if (cascadesAreSameMethod(((AbstractBlock)child1).getNode(), ((AbstractBlock)child2).getNode())) {
+          return Spacing.createSpacing(0, 1, 0, mySettings.KEEP_LINE_BREAKS, mySettings.KEEP_BLANK_LINES_IN_CODE);
```
alexander-doroshko: The 2nd argument (maxSpaces) should probably be zero. Dart Style doesn't allow spaces before cascade when it is on the same line.
The same about line 361 below.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -371,14 +349,41 @@ else if (type1 == COLON) {
...
 
+    if (type1 == CLOSING_QUOTE && type2 == OPEN_QUOTE && elementType == STRING_LITERAL_EXPRESSION) {
+      ASTNode sib = node1;
+      int preserveNewline = 0;
+      // Adjacent strings on the same line should not be split.
+      while ((sib = sib.getTreeNext()) != null) {
+        // Comments are handled elsewhere.
+        if (sib.getElementType() == WHITE_SPACE) {
+          String ws = sib.getText();
+          if (ws.contains("\n") || ws.contains("\r")) {
```
alexander-doroshko: Just FYI: line breaks are always normalized in Document instances and AST/PSI. You'll never meet '\r', only '\n'.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -371,14 +349,41 @@ else if (type1 == COLON) {
...
+        }
+      }
       return addLineBreak();
     }
 
+    if (type1 == CLOSING_QUOTE && type2 == OPEN_QUOTE && elementType == STRING_LITERAL_EXPRESSION) {
+      ASTNode sib = node1;
+      int preserveNewline = 0;
+      // Adjacent strings on the same line should not be split.
+      while ((sib = sib.getTreeNext()) != null) {
```
alexander-doroshko: I tried to comment out the whole 'while' statement, but all tests still pass. This code looks not trivial, please make sure it is covered by tests.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -371,14 +349,41 @@ else if (type1 == COLON) {
...
+          if (ws.contains("\n") || ws.contains("\r")) {
+            preserveNewline++;
+            break;
+          }
+          continue;
+        }
+        break;
+      }
+      // Adjacent strings on separate lines should not include blank lines.
+      return Spacing.createSpacing(0, 1, preserveNewline, true, 0);
```
alexander-doroshko: Shouldn't the first argument be 1?
I guess there are many more places in this class where the first argument of Spacing.createSpacing() should be equal to the second one, but 0 and 1 are passed instead.

Dart/testData/formatter/CascadeSame.dart
```diff
@@ -0,0 +1,3 @@
+m(list) {
+  list.add("first").add("second");
```
alexander-doroshko: did you mean cascade here?

alexander-doroshko:
Pushed!
I didn't do any additional commits, please take a look at my comments where you agree with them.

P.S. Please a bit more specific commit messages. Like 'Continuation indent in Dart binary expressions' instead of 'More formatter work'. Thank you!

PR #193, opened 2015-05-29

stevemessick:
@alexander-doroshko There's not much here but I want to sync up before switching repositories.

I changed a few things to match your review comments. Max spaces, newline checking.

I'm starting to work on detailed spacing. Type arguments, is-not test, named args so far. I have not seen UI elements to control these (but they may exist) so there's no checking for preferences. I'll add that if I find the correct preferences -- but I think our first priority is to stop undoing what dart_style does.

Quite a few test files needed minor updates. I'm still trying to figure out how to make {} appear on a single line when appropriate, and not break anything. Also still thinking about the conditional (ternary) operator. I may have to adjust the DartBlock tree to handle it correctly.


Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
```diff
@@ -384,6 +388,25 @@ else if (type1 == REFERENCE_EXPRESSION) {
...
 
+    if (elementType == NAMED_ARGUMENT) {
+      if (type1 == COLON) {
+        return addSingleSpaceIf(true);
+      }
+      if (type2 == COLON) {
+        return noSpace();
+      }
+    }
+    if (elementType == TYPE_ARGUMENTS && type1 == LT || type2 == GT) {
```
alexander-doroshko: class A < B > is covered partly by this 'if' (and probably that's not what you expected to be fixed by this code).
At first I was afraid that type2 == GT check is dangerous for relational operator. It's kind of luck that it is not affected. I think it would be good to make code easier to understand by adding one more check in addition to type2 == GT, for example type2 == GT && elementType == SOMETHING.

alexander-doroshko:
Pushed.
2 more cases pass in DartStyleTest.
I'm afraid some of the tweaks done in DartSpacingProcessor are not covered with tests. Feel free to update any input test data file by adding spaces where they are not expected and removing where they should be - output file should probably remain inchanged in these places.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartSpacingProcessor.java
stevemessick: That's a typo, which should be: (type1 == LT || type2 == GT)
All type args should be formatted with minimal spaces, as in: 
abstract class HashMap<K, V> implements Map<K, V>
I'll fix it. It currently permits a space before '<' which should be tightened up, too.

stevemessick: Sorry, example got munged.
`abstract class HashMap<K, V> implements Map<K, V>`

PR #194, opened 2015-05-29

stevemessick:
@alexander-doroshko I didn't have your latest changes to DartStyleTest when I made the other pull request. Since the changes to spacing already cause some dart_style tests to pass I need to update that. Here you are!


alexander-doroshko:
I already pushed the same :)

PR #195, opened 2015-06-01

stevemessick:
@alexander-doroshko Most of the interesting changes are in DartSpacingProcessor.

We need to carefully consider if KEEP_LINE_BREAKS should default to false. Changing it causes several previously broken dart_style tests to succeed. I'm not sure what the user expectations are, though, so I didn't change the default.

I wonder if CALL_PARAMETERS_LPAREN_ON_NEXT_LINE is misnamed. It does not seem to do what I thought it would do from its name.

25 new dart_style tests pass, including all tests run by expressions3().

Empty blocks are still not formatted correctly (according to dart_style). I didn't change conditional operator handling yet, either.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartIndentProcessor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartIndentProcessor.java
@@ -61,6 +61,9 @@ public class DartIndentProcessor {
       if (elementType == LBRACE || elementType == RBRACE || elementType == LBRACKET || elementType == RBRACKET) {
         return Indent.getNoneIndent();
       }
+      if (elementType == TYPE_ARGUMENTS) {
+        return Indent.getNoneIndent();
+      }
       return Indent.getNormalIndent();
     }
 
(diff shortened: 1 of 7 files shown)


alexander-doroshko:
Pushed.
I do not know user expectations about default value of KEEP_LINE_BREAKS as well.
What I think is that default value of this option is less important than the default value of RIGHT_MARGIN. You have already set it to 80 unlike all other languages where IntelliJ default is 120. The argument is that [Style Guide](https://www.dartlang.org/articles/style-guide/) recommends to go this way.
Personally for me the value of 80 sounds very 1990'th when monitors were small. In IDEA project we use 140 and the longest line still takes only half width of my monitor. I have no problem reading wide code, my eyes do not get tired. For me it is more important to see more code without scrolling.
The problem is that it is not easy to find the place in Settings where user may change right margin for Dart.
Well, that's a subject for a big separate discussion with wider audience. We have some time before the next major release and may collect feedback about defaulting to 80 during EAP period.

As for CALL_PARAMETERS_LPAREN_ON_NEXT_LINE - I didn't play with it, no idea. Check Java formatter implementation if you want.

PR #196, opened 2015-06-02
More work on Dart formatting.

stevemessick:
@alexander-doroshko There's a problem with wrapping and indenting arguments. I left a TODO in DartIndentProcessor explaining what's going on, but I don't know how to fix it yet. Having more context in the indenter, like the wrapper has with childWrap, might help.

One dart_style test that previously was "passing" no longer is. In fact, it was passing because the cascade wrapping logic was incorrect. I'm still not convinced cascade receivers are being handled correctly, but it is certainly better (as in: more tests pass).

DartStyleTest itself had some issues with the tests in 'regression'. Fixing that also caused more tests to pass, which is nice.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartWrappingProcessor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/formatter/DartWrappingProcessor.java
@@ -38,9 +38,10 @@ public class DartWrappingProcessor {
           return createWrap(true); // Allow first arg to wrap.
         }
         if (!mySettings.PREFER_PARAMETERS_WRAP && childWrap != null) {
+          // Not used; PREFER_PARAMETERS_WRAP cannot be changed in the UI.
           return Wrap.createChildWrap(childWrap, WrappingUtil.getWrapType(mySettings.CALL_PARAMETERS_WRAP), true);
         }
-        return Wrap.createWrap(WrappingUtil.getWrapType(mySettings.CALL_PARAMETERS_WRAP), true);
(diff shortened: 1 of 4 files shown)


Dart/src/com/jetbrains/lang/dart/ide/formatter/DartWrappingProcessor.java
```diff
@@ -38,9 +38,10 @@ Wrap createChildWrap(ASTNode child, Wrap defaultWrap, Wrap childWrap) {
           return createWrap(true); // Allow first arg to wrap.
         }
         if (!mySettings.PREFER_PARAMETERS_WRAP && childWrap != null) {
+          // Not used; PREFER_PARAMETERS_WRAP cannot be changed in the UI.
```
alexander-doroshko: If you think it should be in UI feel free to add it there

Dart/testSrc/com/jetbrains/lang/dart/dart_style/DartStyleTest.java
```diff
@@ -384,7 +366,7 @@ public void testTop_level() throws Exception {
     runTestInDirectory("comments");
   }
 
-  @Bombed(year = 2015, month = Calendar.NOVEMBER, day = 12, user = "alexander.doroshko",
+  @Bombed(year = 2015, month = Calendar.MAY, day = 12, user = "alexander.doroshko",
```
alexander-doroshko: This is the date until which the test is not run on the build server. So it should be some future date when we plan to understand how to fix the test. So far I have no idea why the test behaves completely different on TeamCity server. Some subtests that fail locally pass on server and vice versa. Reverting to November.

PR #197, opened 2015-06-04
Fix up empty Dart block formatting.

stevemessick:
@alexander-doroshko Work on formatting empty blocks. Add UI options to keep simple (empty) blocks on same line, which are enabled by default to be consistent with dart_style.

There are a bunch of warnings related to 'lineFeeds = 0' that should be ignored. I think the code is clear the way it is written. If it needs to be modified later, this form will probably be easier to work with than the 'proper' style according to the warning. (I would have used a switch if the type was an enum but it isn't.)

One of the tests has an empty block (for a mixin) that does not collapse to a single line. That's because the option to remove lines is not set by default in DartFormatterTest. This is a Good Thing. :)

I extended the conditional operator test but it is still not being run. No progress on that front.


alexander-doroshko:
Agree, 'if statement with identical branches' is justified here.
Hint: you can suppress the warning, press Alt+Enter and Right arrow to see options.

alexander-doroshko:
For this example IDE adds line break, but Dart Style doesn't:
foo() {
  {}
}

alexander-doroshko:
Two more degenerate examples: empty try and empty finally.
try {} finally {}

alexander-doroshko:
More real-life example fails: function expression with empty body.
var a = () {};

Dart/src/com/jetbrains/lang/dart/ide/formatter/settings/DartLanguageCodeStyleSettingsProvider.java
```diff
@@ -50,8 +50,12 @@ public CommonCodeStyleSettings getDefaultCommonSettings() {
     defaultSettings.BINARY_OPERATION_WRAP = CommonCodeStyleSettings.WRAP_AS_NEEDED;
     defaultSettings.ASSIGNMENT_WRAP = CommonCodeStyleSettings.WRAP_AS_NEEDED; // TODO Remove if not needed.
     defaultSettings.ARRAY_INITIALIZER_WRAP = CommonCodeStyleSettings.WRAP_AS_NEEDED;
-    defaultSettings.TERNARY_OPERATION_WRAP = CommonCodeStyleSettings.WRAP_AS_NEEDED;
+    defaultSettings.TERNARY_OPERATION_WRAP = CommonCodeStyleSettings.WRAP_ON_EVERY_ITEM;
```
alexander-doroshko: WRAP_ON_EVERY_ITEM is not a valid value for this option. You get million of NPEs when you open Wrapping and Braced tab of the Dart settings. Did you notice them in debug console?

I spent half an hour hunting for the answer. And here it is: see the code where [TERNARY_OPERATION_WRAP](https://github.com/JetBrains/intellij-community/blob/9eb2ba1e8ce1d0b47e8f12f7c403712cb9039e56/platform/lang-impl/src/com/intellij/application/options/codeStyle/WrappingAndBracesPanel.java#L146) option is declared.
[CodeStyleSettingsCustomizable#WRAP_VALUES](https://github.com/JetBrains/intellij-community/blob/9eb2ba1e8ce1d0b47e8f12f7c403712cb9039e56/platform/lang-api/src/com/intellij/psi/codeStyle/CodeStyleSettingsCustomizable.java#L256) array contains possible values of TERNARY_OPERATION_WRAP option. And there is no WRAP_ON_EVERY_ITEM, but instead there is WRAP_AS_NEEDED | WRAP_ON_EVERY_ITEM which is equal to 5.

PR #198, opened 2015-06-05
More work on Dart formatter

stevemessick:
@alexander-doroshko Most of the work is driven by review comments. I wanted to have a comprehensive test of all empty blocks. Not sure I've got it, but it is close. Had to do a little work on the logic of spacing elements at top-level and class member level.

I didn't add UI affordances for spaces around 'finally'. For consistency, they should be there, but I'd rather consistently remove items.


alexander-doroshko:
> I didn't add UI affordances for spaces around 'finally'. For consistency, they should be there, but I'd rather consistently remove items.

You mean 'Catch on new line' and 'Finally on new line' settings? Yep, either both or none. I see no harm in having both like done for Java.

PR #200, opened 2015-06-10
More formatter tweaks.

stevemessick:
@alexander-doroshko Lots of small changes that make a bunch of dart_style tests work. I think 35 new tests are passing, while one that used to pass no longer does. We're well over 50% passing now.

The default value for parameter alignment is changed. By default, alignment is allowed but not required, so this makes sense, but users can change it. There are two settings that are adjusted in DartStyleTest. These can all be changed by the user. Leaving the defaults as-is makes sense if the majority of users will be IntelliJ customers who want to start working with Dart. If the majority are Dart users who are new to IntelliJ then perhaps the default should be changed to the values used in DartStyleTest. I will keep adding adjustments to DartStyleTest as I discover the need and leave it up to JetBrains engineering/product mgmt to decide the defaults.

I'm done trying to make conditional operators format the way dart_style does it. I don't think we can do it with this framework. Parameter lists are not looking like they will work quite right, either.


alexander-doroshko:
> We're well over 50% passing now.

Great!

> If the majority are Dart users who are new to IntelliJ then perhaps the default should be changed to the values used in DartStyleTest.

I think defaults should match Dart Style. I expect a lot of user complaints about right margin for Dart not matching right margin set in the IDE for all other languages (especially as Dart-specific setting is not that easy to find). As for the other settings, I think users do not care that much about defaults. Well, EAPs that will start soon will help us to understand user expectations.

> I'm done trying to make conditional operators format the way dart_style does it. Parameter lists are not looking like they will work quite right, either.

Can you please send me a summary of your investigation. I'll take a look myself and ask formatter framework guru.

PR #204, opened 2015-06-22
Work on wrapping long lines.

stevemessick:
@alexander-doroshko Work in progress, but what's done so far seems to be working. The commented code in DartSpacingProcessor is due to be revisited.


alexander-doroshko:
@stevemessick Pushed.
I see a need in refactoring as I wrote previously. Otherwise we'll have to multiply ugly code with 'user data'.

PR #205, opened 2015-06-23
More attention-to-detail work.

stevemessick:
@alexander-doroshko Adjust formatting of switch and if statements to match dart_style a little better. Also eliminate extra spaces in type params.


alexander-doroshko:
pushed

PR #207, opened 2015-07-01
Basic implementation for using Analysis Server for resolution.

scheglov:


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -81,6 +89,7 @@
   private final List<String> myPriorityFiles = new ArrayList<String>();
 
   @NotNull private final Queue<CompletionInfo> myCompletionInfos = new LinkedList<CompletionInfo>();
+  @NotNull private final Map<String, List<NavigationRegion>> myNavigationData = new THashMap<String, List<NavigationRegion>>();
```
alexander-doroshko: Map is updated and accessed from different threads. Aren't you afraid of ConcurrentModificationException or even worse: failure without exception?

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -22,9 +30,36 @@
...
+      //}
+      final VirtualFile refVirtualFile = refPsiFile.getVirtualFile();
+      if (refVirtualFile != null) {
+        if (isIdentifier(reference)) {
+          Project project = reference.getProject();
+          final int refOffset = reference.getTextOffset();
+          final String refPath = refVirtualFile.getPath();
+          final List<NavigationRegion> regions = DartAnalysisServerService.getInstance().getNavigation(refPath);
+          for (NavigationRegion region : regions) {
+            if (region.containsInclusive(refOffset)) {
```
alexander-doroshko: shouldn't we check exact regions match?

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -22,9 +30,36 @@
...
+              NavigationTarget target = region.getTargetObjects().get(0);
+              return getResultForNavigationTarget(project, target);
```
alexander-doroshko: We can return a list of target PsiElements from this method, so we do not have to ignore other region.getTargetObjects(). BTW what are cases when this list is greater than 1? Duplicate declarations?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -283,12 +307,26 @@ public static DartAnalysisServerService getInstance() {
     return ServiceManager.getService(DartAnalysisServerService.class);
   }
 
+  /**
+   * Returns {@link NavigationRegion}s for the given file.
+   * Empty if no regions.
+   */
+  @NotNull
+  public List<NavigationRegion> getNavigation(String file) {
+   List<NavigationRegion> sourceRegions = myNavigationData.get(file);
```
alexander-doroshko: DAS uses native slashes in file paths, but IDE always uses forward slashes in its model. So this code didn't work on Windows. I fixed it in a follow up commit.

alexander-doroshko:
Pushed. At first glance it looks like a good Proof Of Concept. Of course there are a lot of things to fix and improve, I'll write later.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Yes, I should a synchronization here and in getNavigation().

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: Yes, we should.
I'm moving in this direction.
With this change to the server https://github.com/dart-lang/sdk/commit/7e194d8fb5960f6249e87b5648dc67c1e928cc64 we can start doing this.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: A part may be used by more then one library, so have different imports / surrounding classes.

lib1.dart
--------
library lib;
part 'test.dart';
class A {}

lib2.dart
--------
library lib;
part 'test.dart';
class A {}

test.dart
--------
part of lib;
main() {
  new A();
}

  Here "A" may be resolved to two different elements "A" - in lib1.dart or in lib2.dart files.
  Probably not very useful though.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Or just create map as Collections.synchronizedMap(new THashMap<String, List<NavigationRegion>>()) - it's the same as adding 2 synchronized blocks.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
alexander-doroshko: Yep, and I guess there is also a lot of examples of invalid code with duplicate declarations when reference resolution is ambiguous. Does server return several regions for the reference in this case?
In any case it's easy for us to support multi-resolve.

scheglov: Currently DAS doesn't return several targets.
But it could.
I will make a tweak to use all targets.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerAnnotator.java
```diff
@@ -88,6 +74,27 @@ public AnnotatorInfo collectInformation(@NotNull final PsiFile psiFile, @Nullabl
     return new AnnotatorInfo(psiFile.getProject(), annotatedFile.getPath());
   }
 
+  public static boolean serverReadyForRequest(@NotNull PsiFile psiFile) {
+    if (psiFile instanceof DartExpressionCodeFragment) return false;
+
+    final Module module = ModuleUtilCore.findModuleForPsiElement(psiFile);
+    if (module == null) return true;
```
alexander-doroshko: Is 'return true' intentional here? To preserve logic that was before here should be 'return false'.

scheglov: No, it wasn't intentional.
Just a mistake, sorry.

alexander-doroshko: No worries, I've pushed the fix.

PR #208, opened 2015-07-06
Resolve using analysis server

scheglov:
With updating navigation information on document changes lets us to keep resolution for all references after the changed position valid. This eliminates flashing between user edits and new navigation notifications.

We also check whether notifications we receive are for the most current document state, otherwise we ignore it, because we know that eventually a new one will come, and also attempt to apply any such out-of-sync navigation notification will cause losing resolution, semantic highlighting, flashing, etc.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -1,5 +1,7 @@
 package com.jetbrains.lang.dart.analyzer;
 
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import com.google.dart.server.*;
 import com.google.dart.server.generated.AnalysisServer;
 import com.google.dart.server.generated.types.*;
@@ -18,6 +20,9 @@ import com.intellij.openapi.application.ApplicationNamesInfo;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.editor.EditorFactory;
+import com.intellij.openapi.editor.event.DocumentAdapter;
+import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.fileEditor.FileEditorManagerAdapter;
@@ -82,9 +87,10 @@ public class DartAnalysisServerService {
   private final DartServerRootsHandler myRootsHandler = new DartServerRootsHandler();
   private final Map<String, Long> myFilePathWithOverlaidContentToTimestamp = new THashMap<String, Long>();
   private final List<String> myPriorityFiles = new ArrayList<String>();
+  private final Set<String> myFilePathsWithUnsentChanges = Sets.newHashSet();
 
   @NotNull private final Queue<CompletionInfo> myCompletionInfos = new LinkedList<CompletionInfo>();
-  @NotNull private final Map<String, List<NavigationRegion>> myNavigationData = new THashMap<String, List<NavigationRegion>>();
+  @NotNull private final Map<String, List<PluginNavigationRegion>> myNavigationData = new THashMap<String, List<PluginNavigationRegion>>();
 
   private final AnalysisServerListener myAnalysisServerListener = new AnalysisServerListenerAdapter() {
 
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -403,6 +432,7 @@ public void run() {
         }
       }
     }
+    myFilePathsWithUnsentChanges.clear();;
```
alexander-doroshko: Wouldn't it be more correct to clear this set in updateFilesContent() or even in UpdateContentConsumer.onResponse() inside updateFilesContent()?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -82,9 +87,10 @@
   private final DartServerRootsHandler myRootsHandler = new DartServerRootsHandler();
   private final Map<String, Long> myFilePathWithOverlaidContentToTimestamp = new THashMap<String, Long>();
   private final List<String> myPriorityFiles = new ArrayList<String>();
+  private final Set<String> myFilePathsWithUnsentChanges = Sets.newHashSet();
 
   @NotNull private final Queue<CompletionInfo> myCompletionInfos = new LinkedList<CompletionInfo>();
-  @NotNull private final Map<String, List<NavigationRegion>> myNavigationData = new THashMap<String, List<NavigationRegion>>();
+  @NotNull private final Map<String, List<PluginNavigationRegion>> myNavigationData = new THashMap<String, List<PluginNavigationRegion>>();
```
alexander-doroshko: Contents of this map is never removed, it may lead to incorrect data. For example if file is closed, info about it remains in the map forever, but in fact it may easily become outdated.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -94,9 +100,23 @@ public void computedErrors(@NotNull final String file, @NotNull final List<Analy
...
+          pluginRegions.add(new PluginNavigationRegion(region));
+        }
+        // Put PluginNavigationRegion(s).
+        synchronized (myLock) {
+          myNavigationData.put(file, pluginRegions);
+        }
+        // Force (re)highlighting.
         final VirtualFile virtualFile = LocalFileSystem.getInstance().findFileByPath(file);
         if (virtualFile != null) {
           Project[] projects = ProjectManager.getInstance().getOpenProjects();
```
alexander-doroshko: myRootsHandler.getTrackedProjects() will be better

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -997,6 +1027,36 @@ public void run() {
     return true;
   }
 
+  private void updateInformationFromServer(DocumentEvent e) {
+    final Document document = e.getDocument();
+    final VirtualFile file = FileDocumentManager.getInstance().getFile(document);
+    if (file == null) {
+      return;
+    }
+    final String filePath = file.getPath();
```
alexander-doroshko: Targets in this file that correspond to regions in other files also become outdated. May be should be handled as well.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Though probably it's not a problem, as we have to reimplement Find Usages anyway

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Sorry, it is exactly inside updateFilesContent(). GitHub diff view is not ideal.

alexander-doroshko:
Pushed.
Please reformat code before commit, it also optimizes imports.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -94,9 +100,23 @@ public void computedErrors(@NotNull final String file, @NotNull final List<Analy
...
+        if (myFilePathsWithUnsentChanges.contains(file)) {
+          return;
+        }
+        // Convert NavigationRegion(s) into PluginNavigationRegion(s).
+        List<PluginNavigationRegion> pluginRegions = new ArrayList<PluginNavigationRegion>(regions.size());
+        for (NavigationRegion region : regions) {
+          pluginRegions.add(new PluginNavigationRegion(region));
+        }
+        // Put PluginNavigationRegion(s).
+        synchronized (myLock) {
```
alexander-doroshko: as we talked it is safer to use separate lock for this map

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Actually you're right, it is better to put it into UpdateContentConsumer.onResponse() inside updateFilesContent().

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Done.
Although I'm not sure what to do now, you said that you pushed it already.

alexander-doroshko: Separate pull request or just push it yourself (though to do it you need to have another Community+Dart project configured, based on original intellij-plugins repo)

PR #209, opened 2015-07-07
Fixes for review comments for DAS based resolver.

scheglov:


alexander-doroshko:
Pushed. GitHub repo mirror is synced with the main repo once an hour.

PR #210, opened 2015-07-07
Parse null-aware operators.

stevemessick:
@alexander-doroshko Work in progress. Parser accepts null-aware operators. Not enabled in analysis server yet; see DartAnalysisServerService. Resolution needs work, maybe other stuff does too.


Dart/src/com/jetbrains/lang/dart/lexer/Dart.flex
```diff
@@ -244,6 +247,7 @@ HEX_NUMBER = 0 [Xx] {HEX_DIGIT}*
 <YYINITIAL, LONG_TEMPLATE_ENTRY> "&&"               { return AND_AND; }
 <YYINITIAL, LONG_TEMPLATE_ENTRY> "@"                { return AT; }
 <YYINITIAL, LONG_TEMPLATE_ENTRY> "#"                { return HASH; }
+<YYINITIAL, LONG_TEMPLATE_ENTRY> "?."               { return QUEST_DOT; }
```
alexander-doroshko: duplicate of line 237

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -897,7 +897,7 @@ public void println(String str) {
         final boolean ok = runInPooledThreadAndWait(new Runnable() {
           @Override
           public void run() {
-            server.analysis_updateOptions(new AnalysisOptions(true, true, true, false, false, true, false));
+            server.analysis_updateOptions(new AnalysisOptions(true, true, true, false /*true*/, false, true, false));
```
alexander-doroshko: Are there reasons not to enable it?

alexander-doroshko:
Can you please take a look at DartTokenTypesSets class. I guess some token sets should be updated.

PR #211, opened 2015-07-08
Optimize search for navigation regions, library refs resolution.

scheglov:


Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -46,9 +45,9 @@ public static boolean isServerDrivenResolution() {
...
           final int refLength = reference.getTextLength();
-          final List<PluginNavigationRegion> regions =
-            DartAnalysisServerService.getInstance().getNavigation(refVirtualFile);
-          for (PluginNavigationRegion region : regions) {
+          final List<PluginNavigationRegion> regions = DartAnalysisServerService.getInstance().getNavigation(refVirtualFile);
+          PluginNavigationRegion region = findRegionAtOffset(regions, refOffset);
+          if (region != null) {
             final int regionOffset = region.getOffset();
             final int regionLength = region.getLength();
             if (regionOffset == refOffset && regionLength == refLength) {
```
alexander-doroshko: may be move exact boundaries check to findRegionAtOffset()?

scheglov: Good idea.
Will do and update the PR.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -189,6 +185,33 @@ else if (!lValue && component.isGetter()) {
     });
   }
 
+  /**
+   * Find the region with the given offset in the given list of sorted regions.
+   * Returns the found region or null.
+   */
+  @Nullable
+  private static PluginNavigationRegion findRegionAtOffset(List<PluginNavigationRegion> regions, int offset) {
```
alexander-doroshko: O(ln) is better that O(n) :)
Did you see not zero time for this search in profiler? If needed we can make it O(1) by introducing an array of regions with regions.getLast().getOffset() size that would map offset to region.

alexander-doroshko:
pushed

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: Before this change DartResolver.resolve() was 11% of total time.

Now DartResolver.resolve() is about 6% of total time when I type several characters with small delays and then backspace them one by one with small delays. And findRegion() itself is invisible now.

  Here is break down of the current DartResolver.resolve() state.

screen shot 2015-07-08 at 10 08 36 am

PR #212, opened 2015-07-08
Semantic highlighting using Analysis Server.

scheglov:
The new DAS based implementation has highlighting for untyped variables, type parameters, import prefixed, function type aliases and enums. No escape sequences highlighting yet, it will require some DAS work first with a new DAS client snapshot. Also note references, like [A], are highlighted in documentation comments.

The screenshot is not ideal, I've submitted a CL to update DAS to support separate highlighting for functions declaration/reference and static/instance getters/setters declarations.

screen shot 2015-07-08 at 2 18 44 pm


Dart/resources/META-INF/plugin.xml
```diff
@@ -143,6 +143,7 @@
 
     <resolveScopeProvider implementation="com.jetbrains.lang.dart.resolve.DartResolveScopeProvider"/>
 
+    <!--<externalAnnotator language="Dart" implementationClass="com.jetbrains.lang.dart.analyzer.DartAnalysisServerColorAnnotator"/>-->
```
alexander-doroshko: Does it work for Dart embedded in HTML as well?

scheglov: Currently DAS only reports error notification for Dart in HTML.
So, no, it does not work.

OTOH, Dart in HTML is used pretty rarely.

alexander-doroshko: > Dart in HTML is used pretty rarely.
I thought so too but users say differently: https://youtrack.jetbrains.com/issue/WEB-16912

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerColorAnnotator.java
```diff
@@ -0,0 +1,147 @@
...
+    if (type.equals(HighlightRegionType.BUILT_IN)) {
+      return DartSyntaxHighlighterColors.DART_KEYWORD;
+    }
+    if (type.equals(HighlightRegionType.CLASS)) {
+      return DartSyntaxHighlighterColors.DART_CLASS;
+    }
+    if (type.equals(HighlightRegionType.CONSTRUCTOR)) {
+      return DartSyntaxHighlighterColors.DART_CONSTRUCTOR_DECLARATION;
+    }
+    if (type.equals(HighlightRegionType.DYNAMIC_TYPE)) {
```
alexander-doroshko: The question is about DAS behavior: in case of a field of dynamic type do you think it is correct that DYNAMIC_TYPE highlighting wins? I'd expect field to be highlighted as field regardless of its type.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerColorAnnotator.java
```diff
@@ -0,0 +1,147 @@
...
+        final TextAttributesKey textAttributes = TextAttributesKey.createTextAttributesKey(key);
+        final TextRange textRange = new TextRange(region.getOffset(), region.getOffset() + region.getLength());
+        holder.createInfoAnnotation(textRange, null).setTextAttributes(textAttributes);
+      }
+    }
+  }
+
+  @Nullable
+  @Override
+  public Object collectInformation(@NotNull PsiFile file, @NotNull Editor editor, boolean hasErrors) {
```
alexander-doroshko: I'm thinking about combining DartAnalysisServerColorAnnotator with DartAnalysisServerAnnotator.
collectInformation() will be as in DartAnalysisServerAnnotator: it performs checks and sends files updated content.
doAnnotate() will wait both for errors and for updated highlighting data for the current file.
In this case we'll get correct file highlighting in one cycle of highlighting daemon.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -875,7 +916,7 @@ private void startServer(@NotNull final DartSdk sdk) {
       final DebugPrintStream debugStream = new DebugPrintStream() {
         @Override
         public void println(String str) {
-          //System.out.println("debugStream: " + str);
+          System.out.println("debugStream: " + str);
```
alexander-doroshko: comment it back :)

Dart/resources/META-INF/plugin.xml
scheglov: OK, I see.
Well, we will need to extend DAS to support HTML.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerColorAnnotator.java
scheglov: DYNAMIC_TYPE is reported for both declaration/reference of local variables/parameters and only for declarations of fields. We discussed this internally and decided that fields should be always highlighted as fields, even in declarations.

alexander-doroshko: What are arguments for having DYNAMIC_TYPE for not typed local vars and parameters instead of LOCAL_VARIABLE_DECLARATION, LOCAL_VARIABLE and PARAMETER?

scheglov: We can infer/propagate local variables/parameter types in DAS, but sometimes we are unable to do this. In this case in method like

main(p) {
  p.unknownMethod();
}

  We wan't get enough indication that we call a method of a dynamic parameter "p".
  So, at some point we decided that we want to highlight variables of "dynamic" type.

  That said, we are not sure now whether we want to continue highlighting variables or unresolved member references, such as "unknownMethod" in "p.unknownMethod()". Maybe both, and let users to decide what to highlight...

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerColorAnnotator.java
scheglov: I'm sorry, I don't quite understand.

Do you mean copying DartAnalysisServerAnnotator.collectInformation() into DartAnalysisServerColorAnnotator?

Or do you mean copying DartAnalysisServerColorAnnotator semantic highlighting implementation into DartAnalysisServerAnnotator?

alexander-doroshko: Anyway it is out of scope of this pull request. Just to think about.
I like how DartAnalysisServerAnnotator works.
It sends updated content in collectInformation().
It asks and waits for errors in doAnnotate(). When errors are received we are sure that they are actual at the moment.
Then it applies errors in apply().

Semantic highlights are different as they are not returned based on client request, but notify listener instead. So I'm thinking how to mimic behavior of server.analysis_getErrors() that has a callback.

Not for this pull request anyway. I'm pushing it.

PR #213, opened 2015-07-08
Null-aware operators get syntax highlighting and completion.

stevemessick:
@alexander-doroshko I think this completes support for [DEP 0009]
(https://github.com/dart-lang/dart_enhancement_proposals/blob/master/Accepted/0009%20-%20Null-aware%20Operators/proposal.md)

* Address previous review comments.
* Enable null-aware operator parsing.
* Add syntax highlighting: ??= assignment, ?? operator, ?. message/dot.
* Clean up completion for elvis operator.

### DIFF
--- a/Dart/testSrc/com/jetbrains/lang/dart/completion/reference/DartReferenceCompletionInLibraryRootTest.java
+++ b/Dart/testSrc/com/jetbrains/lang/dart/completion/reference/DartReferenceCompletionInLibraryRootTest.java
@@ -54,6 +54,10 @@ public class DartReferenceCompletionInLibraryRootTest extends DartCompletionTest
     doTest();
   }
 
+  public void testNonNullRef() throws Throwable {
+    doTest();
+  }
+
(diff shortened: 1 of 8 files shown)


Dart/testSrc/com/jetbrains/lang/dart/completion/reference/DartReferenceCompletionInLibraryRootTest.java
```diff
@@ -54,6 +54,10 @@ public void testMixin1() throws Throwable {
     doTest();
   }
 
+  public void testNonNullRef() throws Throwable {
```
alexander-doroshko: Next time when you'll need to write test for completion take a look at DartCompletionTest. It looks more powerful and easier to write and maintain.

alexander-doroshko:
Awesome! Support seems to be complete now.
The only thing I'd like to ask is to add a test to DartResolveTest. It can be as simple as testEnum() and other simple tests in this class.

PR #214, opened 2015-07-09
Add resolution test for null-aware message send.

stevemessick:
@alexander-doroshko Resolution test as requested.


alexander-doroshko:
Thanks!

PR #215, opened 2015-07-10
New DAS client. Improved semantic highlighting.

scheglov:
Now as DAS is extended to support more fine grain semantic region types, in this CL we support it in the new DartAnalysisServerColorAnnotator.

Also updates to adopt to the new DAS client package name.
New SDK is required.


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -837,7 +837,7 @@ public void onError(final RequestError error) {
       final boolean ok = runInPooledThreadAndWait(new Runnable() {
         @Override
         public void run() {
-          server.edit_format(filePath, selectionOffset, selectionLength, consumer);
+          server.edit_format(filePath, selectionOffset, selectionLength, -1, consumer);
```
alexander-doroshko: I think this is correct parameter. Please check that it will work when you change margin in Preferences | Editor | Code Style | Dart | Wrappijng and Braces | Right margin.
CodeStyleSettingsManager.getSettings(project).getCommonSettings(DartLanguage.INSTANCE).RIGHT_MARGIN

Dart/src/com/jetbrains/lang/dart/highlight/DartColorsAndFontsPage.java
```diff
@@ -179,8 +218,7 @@ public String getDemoText() {
            "var <top.level.var.decl>topLevelVar</top.level.var.decl> = new SomeClass(null).<inherited.call>g</inherited.call>();\n\n" +
            "<top.level.func.decl>main</top.level.func.decl>() {\n" +
            "  <top.level.func.call>print</top.level.func.call>(<top.level.var.call>topLevelVar</top.level.var.call>);\n" +
-           "}\n"
-      ;
+           "}\n";
```
alexander-doroshko: Good time to update the whole text!
Now it is way too long, has incorrectly colored regions and at the same time doesn't demonstrate all options.

alexander-doroshko:
Pushed, but didn't test much.
Will take another look next week.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: OK

My initial goal was just to keep status quo here.
Later we may start providing the real line length.

scheglov: https://github.com/JetBrains/intellij-plugins/pull/216

PR #219, opened 2015-07-20
Use new flag for enabling new DAS semantic highlight.

scheglov:
Also, new client with new subscriptions.
DAS versions >= 1.9.0 and < 2.0.0 is required.


alexander-doroshko:
What does useAnalysisHighlight2 change in DAS behavior?

> DAS versions >= 1.9.0 and < 2.0.0 is required

Isn't it too strict limitation for the range of supported SDKs?

scheglov:
We have existing DAS clients who rely of analysis.highlight notification.
And we cannot replace its results without breaking them.
So, instead we make it optional, when a client specifies "--useAnalysisHighlight2", it will get new highlighting results with high enough granularity.

I will let @jwren to comment about the SDK range.

jwren:
I don't see any issue here, >=1.9 because that is new functionality is after 1.9.0

< 2.0.0 because 2.0 would presumably be a breaking change.

For DAS's that don't have yet have the support, they will continue working in the normal way.

alexander-doroshko:
@scheglov @jwren Supported SDK range [1.9, 2.0) is fine. The problem is that DAS version is not equal to SDK version. I'm afraid with minimal supported DAS version 1.9 it doesn't work with any SDK available now. According to the errors that I get SDK version -> DAS version table looks like this:

SDK version | DAS version
------------ | -------------
1.12.0-edge | 1.8.0
1.12.0-dev.3.1 | 1.7.0
1.11.1 | 1.7.0
1.10 | 1.6.0

alexander-doroshko:
@scheglov @jwren So what do we have to do with the problem? I'm afraid this pull request completely breaks Dart support.

jwren:
@alexander-doroshko 
You are right Alex that if we land it, and people grab the bleeding edge version of the Dart IJ Plugin, and grab the released version of Dart SDK they would be blocked.

However, as you pointed out we aren't releasing the bleeding edge version of the IJ plugin any time soon.  I think it may be wise for us to update the logic that checks the version of the DAS and update it to 1.9 (DartSdkUpdateChecker.java), in the event that a developer (basically us), points to a version that we aren't compatible with.

For changes that are more additive to the DAS and plugin we should, have backwards compatible logic, but in other situations like this one, we should increment the DartSdkUpdateChecker.

If we did need to do a roll of WS 1.0.4 in a few weeks, then we would simply revert any PR that increments the value in DartSdkUpdateChecker.

alexander-doroshko:
I don't remember saying that we aren't releasing Dart plugin from master branch. WebStorm 11 EAP is going to be announced today, and I'm going to release correspondng version of the Dart plugin for IntelliJ IDEA 15 EAP. They go directly from master branch. I think at this moment we shouldn't block everyone who's using SDK earlier than yesterday's 1.12.0-dev.4.0 (the only available SDK with DAS version 1.9.0). 
Are you saying we **have** to restrict EAPers to use 1.12.0-dev.4.0 or later?

jwren:
It would be nice to not slow down work by us having to maintain a queue of changes in to the plugin repo.

If people are grabbing an EAP version, is it so crazy that the grab a bleeding edge (non-dev) Dart SDK?

alexander-doroshko:
At the moment of this pull request there were zero SDKs available with 1.9.0 DAS version, that's why I didn't accept the pull request immediately.
Now we have 1.12.0-dev.4.0 that seem to work.
So, do you think it's ok that WebStorm 11 will require Dart SDK 1.12+ ?

jwren:
> So, do you think it's ok that WebStorm 11 will require Dart SDK 1.12+ ?

Yes, it will guarantee that people will get bug fixes that that been put into the DAS.

alexander-doroshko:
Ok, I pushed this PR and updated DartAnalysisServerService.MIN_SDK_VERSION to 1.12.

PR #221, opened 2015-07-21
Use DAS in DartUriElementBase.

scheglov:


Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
```diff
@@ -53,6 +57,25 @@ public PsiReference getReference() {
     final String uri = ((DartUriBasedDirective)parent).getUriString();
     final int uriOffset = ((DartUriBasedDirective)parent).getUriStringOffset();
 
+    if (DartResolver.isServerDrivenResolution()) {
+      final PsiFile refPsiFile = getContainingFile();
+      final int refOffset = getTextOffset();
+      final int refLength = getTextLength();
+      final PluginNavigationRegion region = DartResolver.INSTANCE.findRegion(refPsiFile, refOffset, refLength);
+      if (region != null) {
+        final PluginNavigationTarget target = region.getTargets().get(0);
```
alexander-doroshko: can region.getTargets() be empty? I'd add a check for safety.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -129,6 +123,16 @@ else if (leftReference != null) {
   }
 
   @Nullable
+  public PluginNavigationRegion findRegion(final PsiFile refPsiFile, final int refOffset, final int refLength) {
```
alexander-doroshko: can be static

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
```diff
@@ -53,6 +57,25 @@ public PsiReference getReference() {
     final String uri = ((DartUriBasedDirective)parent).getUriString();
     final int uriOffset = ((DartUriBasedDirective)parent).getUriStringOffset();
 
+    if (DartResolver.isServerDrivenResolution()) {
+      final PsiFile refPsiFile = getContainingFile();
+      final int refOffset = getTextOffset();
+      final int refLength = getTextLength();
+      final PluginNavigationRegion region = DartResolver.INSTANCE.findRegion(refPsiFile, refOffset, refLength);
+      if (region != null) {
```
alexander-doroshko: region is always null for me (SDK 1.12.0-dev.3.1). The problem is that in case of import statement like this 
`import "dart:core" as core;`
IDE reference range is `"dart:core"` only, but navigation region returned by server is `import "dart:core"`.

Dart/src/com/jetbrains/lang/dart/psi/impl/DartFileReference.java
```diff
@@ -0,0 +1,93 @@
...
+import com.intellij.openapi.util.TextRange;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.PsiReference;
+import com.intellij.util.IncorrectOperationException;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Reference to a file in import or export directive.
```
alexander-doroshko: and in `part` directive as well :)

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
```diff
@@ -53,6 +57,25 @@ public PsiReference getReference() {
...
+        final PluginNavigationTarget target = region.getTargets().get(0);
+        final String targetPath = target.getFile();
+        final VirtualFile targetVirtualFile = LocalFileSystem.getInstance().findFileByPath(targetPath);
```
alexander-doroshko: Reference should be there even if it can't be resolved. Code that is searching for target file should be moved to DartFileReference.resolve()

Dart/src/com/jetbrains/lang/dart/psi/impl/DartFileReference.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @NotNull
+  @Override
+  public String getCanonicalText() {
+    return myUri;
+  }
+
+  @Override
+  public PsiElement handleElementRename(final String newElementName) throws IncorrectOperationException {
+    throw new IncorrectOperationException("Can't rename file");
```
alexander-doroshko: This completely blocks file rename/move functionality. `return myElement` is better - file will be moved/renamed, but uri will become invalid.
Would be great to update uri on file move/rename. With IDE-driven resolution it works perfectly. I wouldn't like to loose the feature.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -129,6 +123,16 @@ else if (leftReference != null) {
...
+    final VirtualFile refVirtualFile = refPsiFile.getVirtualFile();
```
alexander-doroshko: to get virtual file from PsiElement or PsiFile use DartResolveUtil.getRealVirtualFile(). I do not know exactly in which cases it works better than just PsiFile.getVirtualFile(), but for safety I prefer to use it.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
```diff
@@ -39,25 +39,19 @@ public static boolean isServerDrivenResolution() {
...
-            final Project project = reference.getProject();
-            final List<PsiElement> result = new SmartList<PsiElement>();
-            for (PluginNavigationTarget target : region.getTargets()) {
-              final PsiElement targetElement = getElementForNavigationTarget(project, target);
-              if (targetElement != null) {
-                result.add(targetElement);
-              }
-            }
-            return result;
+      final int refOffset = reference.getTextOffset();
```
alexander-doroshko: isIdentifier method is unused now and can be removed

alexander-doroshko:
Review completed. In general it's ok. I can push like this or wait for the updates first, as you want.

The most difficult part is rename handling. bindToElement() in anonymous class inside DartUriElementBase.java is doing related work. I'm not sure if/how we can use info from server to update uri after file rename/move.

scheglov:
I will do the changes, hopefully everything except bindToElement(), update the PR and write another message when it is ready.

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: Done.
IDEA did not make a proposal for this.
I used to use it to know when a method can be made static :-)

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
scheglov: Right.
To support this feature we updated DAS to return `"dart:core"` instead of `import "dart:core"` as a region.

Dart/src/com/jetbrains/lang/dart/psi/impl/DartFileReference.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
alexander-doroshko: Yep, looks like a bug in IDEA inspection :)

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
scheglov: I'm sorry, but I don't understand.
What should I return if it cannot be resolved?

There were another branches in this method that `return PsiReference.EMPTY_ARRAY` when they cannot resolve an URI.

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
alexander-doroshko: I think we should check SDK version and search for a different region among navigation data for older SDKs

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java
scheglov: Done.

Dart/src/com/jetbrains/lang/dart/psi/impl/DartUriElementBase.java
alexander-doroshko: It's kind of a general rule: return references where references are expected, see if they are resolved only when asked to resolve.
For our case it is not that important to follow this rule, so it's probably ok to return EMPTY_ARRAY.
But another reason to move file search to resolve() is that we shouldn't do extra work unless we are asked for. I suggest to pass file path to DartFileReference and to search for file (and remember result) in its resolve().

scheglov: Ah... Now I get it, thanks.
Done.

scheglov:
The PR is updated.
PTAL

alexander-doroshko:
Pushed.
Additionally I cached null resolve result in DartFileReference and unblocked rename (with having incorrect uri as a result).
Two things left: support older SDKs and update uri reference after move/rename. If it is not something to fix quickly please file issues in YouTrack in order not to forget these cases.

scheglov:
URIs: https://youtrack.jetbrains.com/issue/WEB-17381

SDKs: @jwren clarified this.

PR #222, opened 2015-07-21
Add 'Reanalyze Sources' action.

scheglov:


jwren:
LGTM and tested on my linux box

alexander-doroshko:
Typical mistake that every IntelliJ developer makes at least once a year :).

Developers who work on some framework support think that everyone in the world uses only this framework in the IDE and nothing more. But what will Scala developer think seeing "Reanalyze Sources" action with "Reanalyze all sources" description? :)
Action visibility and availability are controlled in overridden update() method.

But I've got another idea: may be add a button to the Dart Analysis tool window instead of main menu? "Dart Analysis tool window visible" is equivalent to "Reanalyze Dart Sources action applicable". And it seems to be a natural place for this action. DartProblemsViewPanel#fillRightToolbarGroup seems to be the right place to do it.

alexander-doroshko:
Also if the action is declared in plugin.xml (even if not added to any group by default) then user is able to add it anywhere he wants in Preferences | Menus and Toolbars. So checking action applicability is still a good idea.

scheglov:
OK, I've added a check like DartStyleAction does.
Also added an action to DartProblemsViewPanel.

PTAL

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
```diff
@@ -218,6 +218,9 @@ dartium.checked.mode.label=Start Dartium in &checked mode
 dart.project.generate.sample.content.label=&Generate sample content\:
 dart.project.templates.loading.message=Loading Dart project templates...
 
+dart.reanalyze.action.name=Reanalyze Sources
```
alexander-doroshko: I'll rename it to Reanalyze Dart Sources

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
```diff
@@ -14,6 +19,13 @@ public DartProblemsViewPanel(Project project) {
   @Override
   protected void fillRightToolbarGroup(DefaultActionGroup group) {
     super.fillRightToolbarGroup(group);
+    group.add(new AnAction(DartBundle.message("dart.reanalyze.action.name"), DartBundle.message("dart.reanalyze.action.description"),
```
alexander-doroshko: Just FYI: if action is declared in plugin.xml it is possible to get it from ActionManager by id.

Dart/resources/META-INF/plugin.xml
```diff
@@ -209,6 +209,10 @@
             text="Dart File" description="Create new Dart file">
       <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFromTemplate"/>
     </action>
+    <action id="Dart.Reanalyze" class="com.jetbrains.lang.dart.ide.actions.ReanalyzeAction"
+            text="Reanalyze Sources" description="Reanalyze all sources">
+      <add-to-group group-id="BuildMenu" relative-to-action="CompileProject" anchor="after"/>
```
alexander-doroshko: Do you think it is important to have the action in main menu by default? Isn't it enough to have it in tool window?
Personally I'd prefer not to add it to main menu by default. Users can do it themselves in Preferences | Menu and Toolbars

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
scheglov: Thank you, good to know.
I thought I still have to patch enablement somehow, because probably we want to run Reanalyze from the view ignoring what file is selected. Hm... But maybe we can check hosting view in update()? I don't see any related key in CommonDataKeys though.

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
scheglov: SGTM

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
alexander-doroshko: 'ToolWindowManager.getInstance(project).getToolWindow(DartProblemsViewImpl.TOOLWINDOW_ID) != null'
looks like a valid and simple implementation of isEnabled()

scheglov: Hm... Does it guarantee that the action which update() is now executing is hosted by this tool window?

alexander-doroshko: I think action availability and behavior shouldn't differ whether it is invoked for Dart Analysis tool window, from main menu, from Find Action or from any custom location. If DAS works then action is available. Files open in editor or whatever else do not matter. Does it sound correct?
If so then just simplify ReanalyzeAction.update() and use PluginManager to add this action to tool window.

alexander-doroshko:
Pushed.
Just curious: what are typical use cases for this action? When something goes wrong at server side? If so, how does it look like for users?

jwren:
Correct: when something goes wrong on the server side.  Basically it means that resolution information fell out of sync, and this is a refresh.  Sometimes it is because of the DAS (attempts to make it faster/quicker more incremental) may introduce an bug.  Or it could be a package that the DAS depends on such as the watcher package.

scheglov:
Sorry that I don't answer timely.
I was working on another feature in DPI.

And because it wasn't something I'm blocking on, because you had already explained me how to it, I didn't feel it as time pressing. I feel bad that we don't have a more formal way for you to say "This feature requires more work" or for me to say "OK, I'm done with working on the review comments, so if you don't have additional comments, please push it."

Specifically for this PR, I agree that maybe my initial approach with adding it to the main menu is not quite good idea. Maybe just having it in the Dart Analysis panel is enough. In Dart Editor we also have it in the context menu for a project. So, maybe it should be done this way in DPI too. OTOH, I've got some experience in writing IJ actions :-)

I will create another PR to update this feature.

alexander-doroshko:
No worries, I pushed minor follow up commit, I think all is good now. Probably the only thing to improve is action icon, but that's a question to our designers anyway.

PR #223, opened 2015-07-22
Formatnullops

stevemessick:
@alexander-doroshko Some small changes to formatting for null-aware operators. Also improves tests.


Dart/src/com/jetbrains/lang/dart/DartTokenTypesSets.java
```diff
@@ -137,7 +137,9 @@
   );
 
   TokenSet LOGIC_OPERATORS = TokenSet.create(
-    OR_OR, AND_AND
+    OR_OR, AND_AND,
+    // Strictly speaking, this isn't a logical operator, but should be formatted the same.
+    QUEST_QUEST
```
alexander-doroshko: If it is not logic operator we can handle it separately in DartSpactingProcessor without being added to this TokenSet

PR #224, opened 2015-07-22
Add a less-strict form of testing the dart_style test suite.

stevemessick:
@alexander-doroshko I made the original DartStyleTest into an abstract class with Strict (original) and Lenient (new) subclasses. As expected, line splitting stands out as the weakest area of the formatter, and it is also the area least likely to be correctly handled. There are three tests that pass in strict mode that fail in lenient mode, so I'll see about fixing them before diving into line splitting again.


alexander-doroshko:
Awesome! Pushed.
Tests passing in lenient mode - strange indeed. Might be something wrong with the test itself.
When fixed I suggest to introduce KNOWN_TO_FAIL_LENIENT and KNOWN_TO_FAIL_STRICT_ONLY sets. Thus we'll be sure no more strange tests appear.

PR #226, opened 2015-07-22
Improve code completion presentation - icons and parameters.

scheglov:


Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -27,8 +41,130 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
+      return AllIcons.Nodes.Class;
+    }
+    else if (elementKind.equals(ElementKind.ENUM)) {
+      return AllIcons.Nodes.Enum;
+    }
+    else if (elementKind.equals(ElementKind.ENUM_CONSTANT) || elementKind.equals(ElementKind.FIELD)) {
+      return AllIcons.Nodes.Field;
+    }
+    else if (elementKind.equals(ElementKind.COMPILATION_UNIT)) {
+      return PlatformIcons.FILE_ICON;
```
alexander-doroshko: Is it a Dart file? DartIcons.Dart_file would be better in this case

PR #227, opened 2015-07-23
Use DAS for 'Find Usages'.

scheglov:


Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandlerFactory.java
```diff
@@ -0,0 +1,35 @@
...
+import com.intellij.find.findUsages.FindUsagesHandler;
+import com.intellij.find.findUsages.FindUsagesHandlerFactory;
+import com.intellij.psi.PsiElement;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class DartServerFindUsagesHandlerFactory extends FindUsagesHandlerFactory {
+  @Override
+  public boolean canFindUsages(@NotNull PsiElement element) {
+    return true;
```
alexander-doroshko: Hey! It breaks all Find Usages in all JetBrains IDEs in all languages and frameworks!
findUsagesHandlerFactory extension point doesn't have lang attribute, so you have to care about applicability yourself.
Additionally I suggest to keep it switched off by default via DartResolver#isServerDrivenResolution, I think both resolution and find usages should be powered by the same engine.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -742,6 +739,80 @@ public void run() {
...
+        while ((resultSet = mySearchResultSets.poll()) != null) {
+          if (!resultSet.id.equals(searchId)) continue;
+          for (final SearchResult searchResult : resultSet.results) {
+            processor.process(searchResult);
+          }
+          if (resultSet.isLast) return;
+        }
+        // wait for more results
+        try {
+          mySearchResultSets.wait();
```
alexander-doroshko: ProgressManager.checkCancelled() is extremely important for user experience (instant reaction on Escape/Cancel button). See how it is done in addCompletions()

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        if (psiFile != null) {
+          final int offset = location.getOffset();
+          final int length = location.getLength();
+          processor.process(new UsageInfo(psiFile, offset, offset + length));
+        }
+        return true;
+      }
+    };
+    // Send the search request and listen for results.
+    ApplicationManager.getApplication().runReadAction(new Runnable() {
```
alexander-doroshko: ReadAction surrounding a long operation that has wait() inside may lead to deadlock or unresponsive UI. It's better to use multiple short read actions where it is needed instead of a long one for the whole method.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -742,6 +739,80 @@ public void run() {
...
+                 ": " +
+                 (System.currentTimeMillis() - t0) +
+                 "ms");
+        return;
+      }
+
+      searchId = searchIdRef.get();
+    }
+
+    while (true) {
```
alexander-doroshko: In some cases this loop becomes infinite. For example take simple Dart file and search for function usages:
`foooo(){}`
In this case server gives zero notifications via `computedSearchResults()`. Here's full trace:
`
debugStream: 1437741190475 => {"id":"22","method":"search.findElementReferences","params":{"file":"C:/flex_projects/untitled5/dart/bin/dart.dart","offset":0,"includePotential":true},"clientRequestTime":1437741190475}`
`
debugStream: 1437741190480 <= {"id":"22","result":{}}
`

Any ideas how to fix it?

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandlerFactory.java
alexander-doroshko: Here should be some smart filter. For example we shouldn't try to search for 'abstract' keyword.
Probably only DartNamedElement and DartUriElements are eligible, you should know better about server capabilities, what can it search for. Code injection in comments should somehow work too, but regular plain comment should not be searched for usages.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Yes.
According to the DAS spec `If no element was found at the given location, this field will be absent, and no results will be reported via the search.results notification.`

I've added check for this into DaASS.

As for searching of top-level functions, this is another bug.
I forgot to wrap convert `VirtualFile.getPath()` into `FileUtil.toSystemDependentName`.
So, it works on MAC, but not on Windows.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Fixed.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandler.java
scheglov: Fixed.

We need ReadAction here only for getting the `element's` surrounding file path.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandlerFactory.java
scheglov: Done.

scheglov:
The PR is updated.

PR #228, opened 2015-07-24
Insert argument list with parameter names after completion.

scheglov:
It would be nice to propose variables from the context for arguments, but DAS does not have this feature. Hopefully yet.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -41,6 +41,7 @@ import com.intellij.psi.PsiManager;
 import com.intellij.psi.impl.source.resolve.ResolveCache;
 import com.intellij.util.Alarm;
 import com.intellij.util.ArrayUtil;
+import com.intellij.util.Consumer;
 import com.intellij.util.Processor;
 import com.intellij.util.concurrency.Semaphore;
 import com.intellij.util.net.NetUtils;
@@ -253,7 +254,7 @@ public class DartAnalysisServerService {
     }
   }
 
-  public void addCompletions(@NotNull final String completionId, @NotNull final CompletionSuggestionProcessor processor) {
+  public void addCompletions(@NotNull final String completionId, @NotNull final Consumer<CompletionSuggestion> consumer) {
     while (true) {
       ProgressManager.checkCanceled();
 
@@ -267,7 +268,7 @@ public class DartAnalysisServerService {
           if (!completionInfo.myCompletionId.equals(completionId)) continue;
 
           for (final CompletionSuggestion completion : completionInfo.myCompletions) {
-            processor.process(completion);
+            consumer.consume(completion);
           }
 
           if (completionInfo.isLast) return;
@@ -1368,10 +1369,6 @@ public class DartAnalysisServerService {
     public int getOffset() {
       return offset;
     }
-  }
-
-  public interface CompletionSuggestionProcessor {
-    void process(CompletionSuggestion suggestion);
   }
 
   /**
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -1,14 +1,18 @@
 package com.jetbrains.lang.dart.ide.completion;
 
+import com.intellij.codeInsight.AutoPopupController;
 import com.intellij.codeInsight.completion.*;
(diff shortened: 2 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -253,7 +253,7 @@ private void forceFileAnnotation(String file) {
     }
   }
 
-  public void addCompletions(@NotNull final String completionId, @NotNull final CompletionSuggestionProcessor processor) {
+  public void addCompletions(@NotNull final String completionId, @NotNull final Processor<CompletionSuggestion> processor) {
```
alexander-doroshko: yep, Processor is better. Consumer is even better when you never need to stop processing in the middle. By convention boolean returned by process() means "process further" whereas false means stop processing.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -93,33 +113,34 @@ private static LookupElement createLookupElement(Project project, CompletionSugg
...
-      //});
+      // insert handler
+      lookup = lookup.withInsertHandler(new InsertHandler<LookupElement>() {
+        @Override
+        public void handleInsert(InsertionContext context, LookupElement item) {
+          final TemplateManager templateManager = TemplateManager.getInstance(project);
+          final Template template = templateManager.createTemplate("", "");
+          // Fill the template with the parameter names.
+          final List<String> parameterNames = suggestion.getParameterNames();
+          if (parameterNames != null) {
```
alexander-doroshko: it seems to be more accurate to check `if (parameterNames != null)` earlier and simply not to add insertHandler if not needed

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -93,33 +113,34 @@ private static LookupElement createLookupElement(Project project, CompletionSugg
...
-      //lookup = lookup.withInsertHandler(new InsertHandler<LookupElement>() {
-      //  @Override
-      //  public void handleInsert(InsertionContext context, LookupElement item) {
-      //    context.getDocument().insertString(context.getStartOffset(), "()");
-      //  }
-      //});
+      // insert handler
+      lookup = lookup.withInsertHandler(new InsertHandler<LookupElement>() {
+        @Override
+        public void handleInsert(InsertionContext context, LookupElement item) {
```
alexander-doroshko: As a further improvement take a look at AutoPopupController#autoPopupParameterInfo. It shows parameter info tooltip - very handy when we are filling parameters after selecting method from completion list.

alexander-doroshko:
General comment.
Filling or not arguments placeholders after selecting method from completion list: this appears to be one of the differences in IntelliJ IDEA vs Eclipse style. I've found request to implement the same in Java (https://youtrack.jetbrains.com/issue/IDEABKL-5496), the feature is being discussed since 2008 but so far IntelliJ goes another way.
Speaking about Dart - I'm not sure the feature adds much value in case of optional parameters, even for Eclipse-style lovers. Usually you need to pass only some of optional params, so if template adds placeholders for all of them you have to do a lot of extra keystrokes deleting them. And even in case of standart parameters, I'm not sure this template saves any keystrokes.
In IntelliJ world we usually have Parameter Info Tooltip shown automatically after selecting method from completion list (insert handler is as simple as AutoPopupController#autoPopupParameterInfo).
May be we should hold a wider discussion before enabling the feature.

scheglov:
The PR is updated to insert only parenthesis and requesting displaying parameters popup.

PTAL

alexander-doroshko:
Does it handle method with zero parameters correctly? I see that in Java it inserts () but caret is placed after closing ')', not inside, and no tooltip.

scheglov:
No, it didn't.
I've updated the PR to move put cursor after parenthesis if there are no parameters.

alexander-doroshko:
Pushed.
Found some things to improve. Please file issues for cases that can't be fixed soon.

1. if you already have `foo();` and use completion to change it to `fooBar();` insert handler adds extra parentheses: `fooBar()();`. Instead if opening paren is already there (even if with spaces between method name and opening paren), completion should not add extra `()`, but simply move caret after existing `(`. Similar logic is already done for Java.

2. Do not insert parentheses if we are completing code after show/hide in import directive.

3. There are some other cases where we shouldn't insert parentheses for functions/methods. For example when completion argument of a function that expects Function at this place. There might be other cases, you should know better as your Dart experience is better.

PR #229, opened 2015-07-28
Group usages by unit/class members.

scheglov:


Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandler.java
```diff
@@ -64,6 +66,19 @@ public boolean processInReadAction(SearchResult result) {
...
+    final String locationFilePath = location.getFile();
+    final PsiFile psiFile = findPsiFile(context, locationFilePath);
+    if (psiFile != null) {
+      final int offset = location.getOffset();
+      PsiElement result;
+      if ((result = PsiTreeUtil.findElementOfClassAtOffset(psiFile, offset, DartReference.class, false)) != null) {
+        return result;
+      }
+    }
+    return null;
```
alexander-doroshko: We loose some usages this way. At least from doc comments where references are not parsed. May be some other cases as well.
May be if we failed to find a DartReference matching region we can take any deepest PsiElement that includes the region.

alexander-doroshko:
Not related to this pull reuest, just curious: what values can SearchResult.getKind() have?

Dart/resources/META-INF/plugin.xml
```diff
@@ -153,6 +153,8 @@
     <lang.findUsagesProvider language="Dart" implementationClass="com.jetbrains.lang.dart.ide.findUsages.DartFindUsagesProvider"/>
     <usageTypeProvider implementation="com.jetbrains.lang.dart.ide.findUsages.DartUsageTypeProvider"/>
     <findUsagesHandlerFactory implementation="com.jetbrains.lang.dart.ide.findUsages.DartServerFindUsagesHandlerFactory"/>
+    <fileStructureGroupRuleProvider implementation="com.jetbrains.lang.dart.ide.findUsages.UnitMemberGroupRuleProvider"/>
+    <fileStructureGroupRuleProvider implementation="com.jetbrains.lang.dart.ide.findUsages.ClassMemberGroupRuleProvider"/>
```
alexander-doroshko: We have 25 group rule providers for different languages in our code base, among them 8 are called `*Class*GroupRuleProvider`. So to make developer's life easier I suggest to call all Dart specific classes starting with 'Dart'. DartUnitMemberGroupRuleProvider, DartClassMemberGroupRuleProvider. The rule is more important for extensions and classes that extend Platform classes. But I prefer to have Dart word always, it's much easier to guess the class name this way when you are looking for something.

alexander-doroshko:
Just a thought. I see that server provides SearchResult#getPath. When we switch to server powered fund usages we may use this info for grouping if it makes sense.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartComponentUsageGroup.java
```diff
@@ -0,0 +1,138 @@
...
+import com.intellij.usages.UsageView;
+import com.jetbrains.lang.dart.psi.DartComponent;
+import com.jetbrains.lang.dart.psi.DartComponentName;
+import org.jetbrains.annotations.NonNls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+
+class DartComponentUsageGroup implements UsageGroup, DataProvider {
```
alexander-doroshko: Will PsiNamedElementUsageGroupBase super class help to save a few lines?

scheglov:
About SearchResult.getKind() - see https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#L3533

alexander-doroshko:
Thanks, I should have noticed SearchResultKind class myself. One of values is DECLARATION. In which cases are declarations reported as usages?

scheglov:
DECLARATION results are returned for variable-like elements - local variables, parameters and fields.
There are usually not many results, and declaration represents significant part of interesting information about these elements.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandler.java
scheglov: Good catch.
Actually, we don't need to search for any specific PsiElement, just and is OK.
We can then climb up to a class or unit members.

Ouch... Documentation comments are not part of elements they comment.
So, references in comments are all grouped in class or unit instead of methods/classes/etc.
image

Dart/resources/META-INF/plugin.xml
scheglov: OK, I will rename them.

See also "Smurf Naming Convention" http://blog.codinghorror.com/new-programming-jargon/ :-)

scheglov:
Yes, we use SearchResult#getPath in Dart Editor because we want it to be a dumb editor, it does not parse Dart code at all. So, we don't know otherwise enclosing elements and cannot built a tree.

OTOH, in IDEA we do parse Dart code. Also, I don't know if it is possible to group results without having PsiElement(s) anyway, it seems that we are asked for CommonDataKeys.PSI_ELEMENT to do this. In Dart Editor we have our own UI panel, but IDEA's one is nicer.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartServerFindUsagesHandler.java
alexander-doroshko: Yes, unfortunately parser is not perfect in this area. I was told it is easy to fix, but so far I didn't do it. There's a special logic somewhere that finds doc comment for component, so while parser not fixed we need to write a helper that finds component for doc comment.

Dart/resources/META-INF/plugin.xml
alexander-doroshko: Yep, all class names related to Smurf support start with Smurf. :) Otherwise it would be difficult to find anything in the repository that contains support for more than 100 technologies.

Dart/src/com/jetbrains/lang/dart/ide/findUsages/DartComponentUsageGroup.java
scheglov: PsiNamedElementUsageGroupBase looks pretty much the same, but there are two issues with it.

1. It implements calcData(), but does not implement TypeSafeDataProvider, so calcData() is not used. This means that Preview does not work.

2. If I make it implement TypeSafeDataProvider, it still does not work well. It does not highlight names, but instead just a first token.
image
image

So, at least for now, DartComponentUsageGroup seems to provide better functionality.

scheglov:
The PR is updated.
PTAL

alexander-doroshko:
There was a problem with dartUnitMemberGroupingRule.java file name in the pull request that lead to compilation failure. We do not know reliable way to rename only case in file name, so the first person who noticed compilation failure renamed the class to DartUnitMemberUsageGroupingRule.

scheglov:
Thank you (and that person too!) for fixing these problems.

PR #230, opened 2015-07-28
'Sort Members in Dart File' action.

scheglov:


yole:
There is already an action for sorting the members in a file: Code | Rearrange Code. It provides an API (com.intellij.psi.codeStyle.arrangement.Rearranger) allowing language plugins to integrate with it. Please consider using the standard action instead of adding a Dart-specific one.

alexander-doroshko:
@yole Dart team from Google develops Dart Analysis Server: it provides everything that is needed to make smart IDE out of a plain text editor. In our Dart support we already use a lot of Server driven services like errors/warnings highlighting and quick fixes. There's no need to do the same work twice (at server side and at IDE side) so we are moving forward to using server driven code completion, reference resolution, find usages, etc. So server driven code rearrangement is one of these steps in Dart support.

yole:
So how exactly does this require adding a separate action for sorting the members? If the API we have isn't flexible enough to allow running the logic on the server, we should extend the API, instead of cluttering the menu with multiple actions that do the same thing.

alexander-doroshko:
I think honest answer is because this is a quick and easy working solution.
Menu shouldn't become cluttered because of limited action availability/visibility.

alexander-doroshko:
I've filed a task to intergate Sort Dart Members action to the IntelliJ standard Rearrange Code: https://youtrack.jetbrains.com/issue/WEB-17538

PR #231, opened 2015-07-28
Invocation completion improvements - parenthesis, identifier.

scheglov:
Fixes for https://github.com/JetBrains/intellij-plugins/pull/228 review comments.

1. Use ParenthesesInsertHandler to reuse the same logic as Java uses.

2. Check whether suggestion's kind is INVOCATION, for import directives IDENTIFIER is returned, so we know when we should not insert arguments.

3. The same INVOCATION / IDENTIFIER works also for using functions as arguments, but we need to wait for isLast suggestions notification.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -259,19 +259,15 @@ public class DartAnalysisServerService {
       ProgressManager.checkCanceled();
 
       synchronized (myCompletionInfos) {
-        // todo in case of several AnalysisServerListenerAdapter.computedCompletion() invocations for one completion (only the last
-        // invocation has isLast==true) it looks like each next List<CompletionSuggestion> also contains all items that were already
-        // given in previous invocations. If it is by design we should optimize cycle and handle only the last in queue.
-
         CompletionInfo completionInfo;
         while ((completionInfo = myCompletionInfos.poll()) != null) {
           if (!completionInfo.myCompletionId.equals(completionId)) continue;
+          if (!completionInfo.isLast) continue;
 
           for (final CompletionSuggestion completion : completionInfo.myCompletions) {
             consumer.consume(completion);
           }
-
-          if (completionInfo.isLast) return;
+          return;
         }
 
         try {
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -259,19 +259,15 @@ public void addCompletions(@NotNull final String completionId, @NotNull final Co
...
 
       synchronized (myCompletionInfos) {
-        // todo in case of several AnalysisServerListenerAdapter.computedCompletion() invocations for one completion (only the last
-        // invocation has isLast==true) it looks like each next List<CompletionSuggestion> also contains all items that were already
-        // given in previous invocations. If it is by design we should optimize cycle and handle only the last in queue.
-
         CompletionInfo completionInfo;
         while ((completionInfo = myCompletionInfos.poll()) != null) {
           if (!completionInfo.myCompletionId.equals(completionId)) continue;
+          if (!completionInfo.isLast) continue;
```
alexander-doroshko: Did you mean `break` here? Looks like we need to get out of inner loop and to myCompletionInfos.wait() a bit more for the next notification.

alexander-doroshko:
It is by design that each following addCompletions() invocation contains suggestions that were already given in previous invocations? What is the rationale behind this? Clients who need this can collect suggestions themself.
If this is not something that can be changed at server side, may be instead of waiting for the last we'd better filter out duplicates ourselves? This will result in showing first suggestions faster to the user.

scheglov:
Yes, it is by design.

The rationale is that if the unit in which completion is requested is not resolved, then only AST based suggestions are sent, such as scope local elements, top-level elements (they are cached). And later, after resolution of the unit, the set of suggestions is refined.

Is it possible in IDEA to update CompletionResultSet? It looks to me that it is add-only.

alexander-doroshko:
It isn't possible to remove whatever from CompletionResultSet, but we can try to update DartLookupElement behavior when we get updated info. What exactly can be changed when we get CompletionSuggestion for the same element again? Can it be a different presentation or CompletionSuggestionKind? In the latter case we need to update InsertHandler behavior: we can pass some flag to already shown DartLookupElement that will change its behavior.

scheglov:
The server specification does not put any limitations on possible changes between the first and the completion sets. So, in general we need to replace them completely.

Maybe @danrubel can decide whether we want to add some guaranties about changes.

PR #232, opened 2015-07-29
Integration with the new analyzedFiles notification from the Dart Analysis Server, removing the need for the PubListPackageDirsAction

jwren:


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -331,6 +345,148 @@ public LibraryDependenciesResult(@Nullable final String[] libraries,
...
+        finally {
+          if (!modifiableModel.isDisposed()) {
+            modifiableModel.dispose();
+          }
+        }
+      }
+    }
+  }
+
+  private static Library createDartPackagesLibrary(@NotNull final Project project,
```
alexander-doroshko: similar code is in DartFileListener#updatePackagesLibraryRoots, please try to reuse it

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -331,6 +345,148 @@ public LibraryDependenciesResult(@Nullable final String[] libraries,
...
+            // if there is a pubspec, skip this contentRoot
+            if (contentRoot.findChild(PubspecYamlUtil.PUBSPEC_YAML) != null) continue;
+            affectedModules.add(module);
+          }
+        }
+      }
+
+      final Library library = createDartPackagesLibrary(project, rootsToAddToLib);
+
+      for (final Module module : affectedModules) {
```
alexander-doroshko: Outdated dependencies of module on library are not removed. Consider reusing DartFileListener#updateDependenciesOnDartPackagesLibrary. The only difference is module applicability check.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -331,6 +345,148 @@ public LibraryDependenciesResult(@Nullable final String[] libraries,
...
+      }
+
+
+      final Set<Module> affectedModules = new THashSet<Module>();
+      final Module[] modules = ModuleManager.getInstance(project).getModules();
+      for (final Module module : modules) {
+        if (DartSdkGlobalLibUtil.isDartSdkGlobalLibAttached(module, sdk.getGlobalLibName())) {
+          for (final VirtualFile contentRoot : ModuleRootManager.getInstance(module).getContentRoots()) {
+            // if there is a pubspec, skip this contentRoot
+            if (contentRoot.findChild(PubspecYamlUtil.PUBSPEC_YAML) != null) continue;
```
alexander-doroshko: Do you think it will be better to check for pubspec.yaml file presence not only in the content root, but anywhere at any nested level in any of module content roots? 
`FilenameIndex.getVirtualFilesByName(module.getProject(), PUBSPEC_YAML, module.getModuleContentScope())`
If no pubspec file in module at all then we think that this is google internal module. Is there any better way to detect google internal project?

alexander-doroshko:
Current code updates library roots and module-on-lib dependencies even if nothing changed since previous library configuration. Suggested to reuse code from DartFileListener performs necessary checks and doesn't change project model when not needed.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -331,6 +345,148 @@ public LibraryDependenciesResult(@Nullable final String[] libraries,
...
+        }, ModalityState.defaultModalityState());
+      }
+    }
+    catch (Exception e) {
+      LOG.warn(e);
+    }
+  }
+
+  private void doConfigureImportedLibraries(@NotNull final Collection<String> libraries) {
+    for (final Project project : myRootsHandler.getTrackedProjects()) {
```
alexander-doroshko: Is there any chance to get information what library file is used in what project?
For example: let us have 2 projects with completely different set of dependencies. If we open any one of them - 'Dart Packages' library for it is configured correctly. If we open both projects simultaneously then we get both projects reconfigured so that Dart Packages library of each of them contains a superset.
**And even worse:** for each individual project `fileIndex.isInContent(vFile)` will return `false` for all files from another project. As a result Dart Packages lib for 1st project will include all content roots of the 2nd project and vice versa. We need to sort it out somehow.

alexander-doroshko:
Would you like me to add an option to the Registry to control if we want to subscribe to GeneralAnalysisService.ANALYZED_FILES or not? Google internal users will have to switch it on once in their life and then forget about it. I believe they will also be able to work with standard pubspec-based projects as well even with this option enabled as they should have server driven reference resolution enabled as well. The option can be called like 'dart.projects.without.pubspec' or suggest any other name.

May be at some point we will switch this mechanism on for everyone instead of my DartFileListener. But at the moment it seems too risky. And the main known problem is when multiple projects are open.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -331,6 +345,148 @@ public LibraryDependenciesResult(@Nullable final String[] libraries,
...
+        ApplicationManager.getApplication().runWriteAction(new Runnable() {
+          public void run() {
+            doConfigureImportedLibraries(rootsToAddToLib);
+          }
+        });
+      }
+    });
+  }
+
+  private static void runInEventDispatchThread(final Runnable runnable) {
```
alexander-doroshko: Notification always comes in ServerResponseReaderThread, so it is safe to proceed without thread check.

More important is that there's no need in blocking ServerResponseReaderThread and waiting until EDT does project configuration. We do not need results immediately, so it is safe to call invokeLater() instead of invokeAndWait().

One more important knowledge is that (except some special cases) project model change should happen in non-modal state (instead of defaultState() that can be modal if, for example, Project Structure dialog is open).

And the last but not the least: you need to use index. That means that the code must not be run in Dumb mode when indexes are not ready.

Al these problems are solved by using `DumbService.getInstance(project).smartInvokeLater()`. I already use it in DartFileListener.

alexander-doroshko:
Review completed.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -104,6 +113,11 @@
   private final AnalysisServerListener myAnalysisServerListener = new AnalysisServerListenerAdapter() {
 
     @Override
+    public void computedAnalyzedFiles(List<String> directories) {
```
alexander-doroshko: Looks like a misleading parameter name in server API. The list contains file paths, not directories.

jwren: Changed to "files"

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: Adding a TODO to this effect. It is very similar, but different.  It will take a bit of rework to use DartLibInfo.  I don't want to potentionally break the code before I leave on a leave, and you leave on vacation.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: Changed DartFileListener#updateDependenciesOnDartPackagesLibrary to take a Module[] instead of a Project, and code is now reused.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: Done.  As for a different way, not at this time, but I have ideas for the future.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: Good question.  1 project should be sufficient, but it is nice to have more than one for other kinds of projects, which is why I recommend ultimate.

alexander-doroshko: One project with a lot of modules of different types is ok. Several open projects where only one contains Dart is also ok (DartRootsHandler.getTrackedProjects() will return only one in this case). More than one open project - so far no idea how to configure correctly. I'm afraid we have to skip auto-configuration in this case for now.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: I iterated bit here until I finally figured out the thread check.  I kept getting exceptions.  I will put in a TODO.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: Correct.

jwren:
Would you like me to add an option to the Registry to control if we want to subscribe to GeneralAnalysisService.ANALYZED_FILES or not?

Good question.  I think we eventually want to be in a place were users don't have to mess with the registries.  Can the registries be set with a file or entry in a file, like IDEA does for the JVM (ideaxx.vmoptions)?

alexander-doroshko:
I'e already added it. You may use it like this:
```
Registry.is("dart.projects.without.pubspec", false)
```
I think it is a better place for secret option than configuration file in installation.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: `SwingUtilities.isEventDispatchThread()` is always `false` isn't it?

PR #233, opened 2015-07-30
Use Quick Assist from DAS.

scheglov:


Dart/src/com/jetbrains/lang/dart/assists/DartQuickAssistIntention.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public void invoke(@NotNull Project project, Editor editor, PsiFile file) throws IncorrectOperationException {
+    AssistUtils.applySourceChange(project, sourceChange);
+  }
+
+  @Override
+  public boolean isAvailable(@NotNull Project project, Editor editor, PsiFile file) {
+    if (!(file instanceof DartFile)) {
+      return false;
```
alexander-doroshko: Will it work in HTML some day?

scheglov: Some day - yes.

Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
```diff
@@ -0,0 +1,59 @@
...
+      @Override
+      public void run() {
+        final List<SourceFileEdit> fileEdits = sourceChange.getEdits();
+        for (SourceFileEdit fileEdit : fileEdits) {
+          final String filePath = FileUtil.toSystemIndependentName(fileEdit.getFile());
+          final VirtualFile file = LocalFileSystem.getInstance().findFileByPath(filePath);
+          if (file != null) {
+            final Document document = FileDocumentManager.getInstance().getDocument(file);
+            if (document != null) {
+              applySourceEdits(document, fileEdit.getEdits());
```
alexander-doroshko: Can quick assist contain LinkedEditGroups? Should we start a template like we do for DartQuickFix?

scheglov: Yes, some of the quick assist have LInkedEditGroups.
I was going to switch to using templates in a subsequent PR.

alexander-doroshko: Ok, please make sure to extract common code to reuse.

alexander-doroshko:
Just FYI: standard way of 'Surround with' assists in IntelliJ is not an intention action, but Cmd+Alt+T on selection. There are already some done for Dart. So now we have 2 different ways of doing the same thing.

scheglov:
Well, I don't know what would be the ideal solution for this.
We could filter out corresponding "Surround with" assists we receive from DAS.

alexander-doroshko:
We need to declare one IntentionAction via plugin.xml and the rest can be added add dynamically. It can extend HighPriorityAction to make sure that the first one is at the top. By the way isn't 20 intentions too much?
The reason why one should be declared using standard way is that user has ability to switch it off, but no ability to switch it on back (Right arrow opens this submenu):
image

scheglov:
OK
https://github.com/JetBrains/intellij-plugins/pull/239

PR #235, opened 2015-08-04
Fix comment detection during indenting

stevemessick:
@alexander-doroshko Fix for WEB-17494, at least some of the cases. Since we don't have full source for all the problematic code I can't be certain everything is fixed.


alexander-doroshko:
Great, pushed!
Can you please check if other 3 usages of FormatterUtil.hasPrecedingSiblingOfType in Dart plugin are justified?

PR #236, opened 2015-08-04
Implement DartImportOptimizer.

scheglov:
Also includes new DAS client to access the new functionality.


alexander-doroshko:
DAS issue: duplicate directives are not deleted.
And one question: do you think it is expected for user to swipe out directives with unresolved URIs?

alexander-doroshko:
I've also added DartOptimizeImportsTest.

scheglov:
1. CL for removing duplicate imports https://codereview.chromium.org/1277983002
2. Yes, both Eclipse and IDEA remove unresolved imports in Java files on Organize Imports.

PR #237, opened 2015-08-05
WEB-17605. Use selection offset / length from CompletionSuggestion.

scheglov:
It seems to me that ideally we should remove these "for ()", "import as " completions and replace them with Live Templates. We need to provide contexts thought in order to not suggest using "import" inside method bodies :-)


alexander-doroshko:
In what cases is `suggestion.getSelectionLength() > 0`?

alexander-doroshko:
I also added DartServerCompletionTest.testCaretPlacementInFor(). If you know cases with non-zero selection please add one more test.

scheglov:
I've looked through the DAS sources and cannot find a case when `selectionLength != 0`.
I will ask the completion author.

PR #241, opened 2015-08-06
Fix for updating PluginNavigationRegion(s).

scheglov:
It helps to keep navigation information in sync with the source after user inserted/deleted some portion of the file, but before new navigation notification is received from DAS.

With this PR we get correct navigation after pressing quickly the SPACE key and Navigate Declaration before `items;` in the following example:
```
main() {
  var items = <String>[];
  items;
}
```

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -20,6 +20,8 @@ import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.editor.EditorFactory;
+import com.intellij.openapi.editor.event.DocumentAdapter;
 import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.fileEditor.*;
 import com.intellij.openapi.progress.ProgressIndicator;
@@ -377,15 +379,12 @@ public class DartAnalysisServerService {
         }
       });
 
-    /*
-
     EditorFactory.getInstance().getEventMulticaster().addDocumentListener(new DocumentAdapter() {
       @Override
       public void beforeDocumentChange(DocumentEvent e) {
         updateInformationFromServer(e);
       }
     });
-    */
 
     registerQuickAssistIntentions();
   }
@@ -936,8 +935,7 @@ public class DartAnalysisServerService {
 
         @Override
         public void onError(final RequestError error) {
-          if (RequestErrorCode.FORMAT_WITH_ERRORS.equals(error.getCode()) ||
-              RequestErrorCode.FORMAT_INVALID_FILE.equals(error.getCode())) {
+          if (RequestErrorCode.FORMAT_WITH_ERRORS.equals(error.getCode()) || RequestErrorCode.FORMAT_INVALID_FILE.equals(error.getCode())) {
             LOG.info(getShortErrorMessage("edit_format()", filePath, error));
           }
           else {
@@ -1342,11 +1340,11 @@ public class DartAnalysisServerService {
         final int eventOffset = e.getOffset();
         final int deltaLength = e.getNewLength() - e.getOldLength();
         for (PluginNavigationRegion region : regions) {
-          if (region.offset <= eventOffset && eventOffset <= region.offset + region.length) {
-            region.length += deltaLength;
-          }
-          else if (region.offset >= eventOffset) {
+          if (eventOffset <= region.offset) {
             region.offset += deltaLength;
+          }
+          else if (region.offset < eventOffset && eventOffset <= region.offset + region.length) {
+            region.length += deltaLength;
           }
           for (PluginNavigationTarget target : region.getTargets()) {
             if (target.file.equals(filePath) && target.offset >= eventOffset) {
@@ -1418,6 +1416,11 @@ public class DartAnalysisServerService {
       for (NavigationTarget target : region.getTargetObjects()) {
         targets.add(new PluginNavigationTarget(target));
       }
+    }
+
+    @Override
+    public String toString() {
+      return "PluginNavigationRegion(" + offset + ", " + length + ")";
     }
 
     public int getOffset() {


alexander-doroshko:
But I commented out updateInformationFromServer() call yesterday to make tests pass. Need to sort it our somehow.

scheglov:
It seems that the test `testErrorsAfterEOF` is failing even if the listener does nothing.

I've attempted to fix this by removing it in Disposable, but it is invoked _after_ the failure.

```
    System.out.println("++++++++++++++++++++++++");
    EditorFactory.getInstance().getEventMulticaster().addDocumentListener(myDocumentListener);
```

```
    Disposer.register(ApplicationManager.getApplication(), new Disposable() {
      public void dispose() {
        System.out.println("-------------------------");
        EditorFactory.getInstance().getEventMulticaster().removeDocumentListener(myDocumentListener);
        stopServer();
      }
    });
```

```
++++++++++++++++++++++++

Listeners leaked for interface com.intellij.openapi.editor.event.DocumentListener:
[com.jetbrains.lang.dart.analyzer.DartAnalysisServerService$2@7bd3705d]

	at org.junit.Assert.fail(Assert.java:88)
	at com.intellij.testFramework.EditorListenerTracker.checkListenersLeak(EditorListenerTracker.java:66)
	at com.intellij.testFramework.fixtures.impl.HeavyIdeaTestFixtureImpl.tearDown(HeavyIdeaTestFixtureImpl.java:135)
	at com.intellij.testFramework.fixtures.impl.CodeInsightTestFixtureImpl$22.run(CodeInsightTestFixtureImpl.java:1253)
	at com.intellij.util.ui.UIUtil.invokeAndWaitIfNeeded(UIUtil.java:2308)
	at com.intellij.testFramework.fixtures.impl.CodeInsightTestFixtureImpl.tearDown(CodeInsightTestFixtureImpl.java:1237)
	at com.intellij.testFramework.fixtures.CodeInsightFixtureTestCase.tearDown(CodeInsightFixtureTestCase.java:61)
	at com.intellij.testFramework.UsefulTestCase.defaultRunBare(UsefulTestCase.java:358)
	at com.intellij.testFramework.UsefulTestCase$4.run(UsefulTestCase.java:434)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:302)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:738)
	at java.awt.EventQueue.access$300(EventQueue.java:103)
	at java.awt.EventQueue$3.run(EventQueue.java:699)
	at java.awt.EventQueue$3.run(EventQueue.java:697)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$1.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:708)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:358)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:242)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:161)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:150)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:146)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:138)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:91)

-------------------------
```

alexander-doroshko:
EditorListenerTracker.checkListenersLeak checks that amount of EditorListener is the same at the moment of startUp() and tearDown(). Our case seems to be legal but I see no way to tell it ti test framework. I temporary disabled listener addition in unittest mode, but when we'll be writing tests for server driven resolution we need to sort it out in a better way.
Probably it'd be better to add listener on server startup and remove on server shutdown. Though it doesn't solve the problem with tests.

alexander-doroshko:
`myNavigationData` map is never cleared and objects are never removed from it. May be we'd better keep info for myVisibleFiles only?

alexander-doroshko:
pushed

scheglov:
OK, see https://github.com/JetBrains/intellij-plugins/pull/242 for clearing data for invisible files.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1342,12 +1340,12 @@ private void updateInformationFromServer(DocumentEvent e) {
...
         for (PluginNavigationRegion region : regions) {
-          if (region.offset <= eventOffset && eventOffset <= region.offset + region.length) {
-            region.length += deltaLength;
-          }
-          else if (region.offset >= eventOffset) {
+          if (eventOffset <= region.offset) {
             region.offset += deltaLength;
           }
+          else if (region.offset < eventOffset && eventOffset <= region.offset + region.length) {
+            region.length += deltaLength;
```
alexander-doroshko: @scheglov With typing inside the region most likely resolution target will change. I think removing the region would be better, would you agree?

Also some more cases are not handled. When selection is deleted in the editor DocumentEvent range and region may intersect, for example:
`eventOffset < region.offset < eventOffset + eventLength < region.offset + region.length`
`region.offset < eventOffset < region.offset + region.length < eventOffset + eventLength`
Will you agree that in case of any intersection we should delete the region from `myNavigationData`.

I noticed this when was working on highlighting. I'm going to add the same logic for `myHighlightData`.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1342,12 +1340,12 @@ private void updateInformationFromServer(DocumentEvent e) {
...
+          }
           for (PluginNavigationTarget target : region.getTargets()) {
             if (target.file.equals(filePath) && target.offset >= eventOffset) {
               target.offset += deltaLength;
```
alexander-doroshko: @scheglov I think if target region intersects with DocumentEvent range we should remove such target (and corresponding region if no more valid targets, or keep in mind that targets array may be empty). Agree?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: 1. I agree, typing in a region most probably changes its resolution, unless it's typing a letter and backspacing it 100 ms later. So, yes, any intersecting regions should be deleted.

2. Highlighting is visible, and it would be less ideal to just remove highlighting regions. For example when user changes the name of a variable in its declaration, we have a special highlighting region, and it should be nice to keep this highlighting, not just turn it into black-on-white, because any other variable name will have the same highlighting.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: I think removing target it less interesting.
Jumping to a place that had some element declaration, not now doesn't is usually not harmful.
But checking every navigation region and target would be probably cause some performance hit.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Ok, I'll update the code and add tests. For navigation - delete if ranges intersect. For highlighting: update if do not intersect or if DocumentEvent range is inside region. Delete if intersect like in my previous examples.

scheglov: SGTM

PR #243, opened 2015-08-10
DartRenameHander - integrate 'Rename' refactoring from DAS.

scheglov:


scheglov:
The PR was updated.
PTAL

alexander-doroshko:
Before I start looking at it please make sure to add tests.

scheglov:
What level of tests do you think we need?

UI is kind of hard to test, so DartRenameHandler itself is out.
Unit tests for the status/ package?

Test just ServerRenameRefactoring - initialize with error and OK, update name with errors and OK, get changes and apply them (in test, because DartRenameHandler does it from UI)?

alexander-doroshko:
We usually do not test UI automatically. Somewhere in code we have `if (isUnitTestMode()) doWithDefaultOptions() else showDialogOrAskForUserInput()`
Tests for renaming usually look like this: configure multi-file project using test data from TestName/before folder, rename something, check all files content using TestName/after folder. For example see DartMoveTest.

scheglov:
I've added tests for DartRenameHandler.
PTAL

alexander-doroshko:
It's safe to push as it is disabled for now, so I'm pushing it.
I'm afraid I didn't spend enough time on reading and testing each line of code yet. So only a couple of general comments so far:
- @Nullable/@NotNull for all parameters, return values and for most of fields greatly helps to understand and use the new API
- a bit more tricky tests would be great like rename across multiple files, rename across hierarchy, constructor + class rename, references in doc comments update. At the same time we do not have the goal to test all DAS functionality, the main purpose of our tests is to cover IDE-side code.

Some problems found during manual testing:
- It is possible to rename symbol from SDK and 3rd party package (in Pub global cache). Shouldn't it be blocked by server?
- I don't know why, but sometimes it doesn't allow to rename symbol from SDK. That's expected, but the error dialog that is shown is not user friendly. Could be a tooltip like when you try to rename Java String class.
- "Search in comments and strings" check box in the DartRenameDialog does not work.
- All dynamic (not strict) usages are silently renamed although they might be not related to the renamed symbol at all.
- Undo message is not user friendly, for example "Undo Rename Top-Level Function" (it is shown if rename affected multiple files).
- Questionable DAS behavior: rename constructor from invocation `new Foo()`. It behaves differently depending on the explicit default constructor presence in class Foo. If there's no explicit constructor - class is renamed. If there is explicit constructor - class is not renamed, but default constructor is changed to named constructor `Foo.newName`. I'd expect class to be renamed in both cases. When I see only `new Foo()` I do not want to jump into class Foo declaration and check explicit constructor presence, but I want refactoring to be predictable.

scheglov:
> It's safe to push as it is disabled for now, so I'm pushing it.
> I'm afraid I didn't spend enough time on reading and testing each line of
> code yet. So only a couple of general comments so far:
>
>    - @Nullable/@NotNull <https://github.com/NotNull> for all parameters,
>    return values and for most of fields greatly helps to understand and use
>    the new API
>
>   OK, will do.


>
>    - a bit more tricky tests would be great like rename across multiple
>    files, rename across hierarchy, constructor + class rename, references in
>    doc comments update. At the same time we do not have the goal to test all
>    DAS functionality, the main purpose of our tests is to cover IDE-side code.
>
>   OK, will do.



>
>
> Some problems found during manual testing:
>
>    - It is possible to rename symbol from SDK and 3rd party package (in
>    Pub global cache). Shouldn't it be blocked by server?
>
>   Probably.

  In Dart Editor we receive all the changes, and then warn the user that
the changes for files outside of the workspace will not be performed.

  How does it work in IDEA?
  Will return LocalFileSystem.getInstance().findFileByPath(filePath)
a non-null file?



>
>    - I don't know why, but sometimes it doesn't allow to rename symbol
>    from SDK. That's expected, but the error dialog that is shown is not user
>    friendly. Could be a tooltip like when you try to rename Java String class.
>
>   I was not able to reproduce this, but yes, "initial status" is probably
better to show as a tooltip.
  Any hint how to show it in IDEA?



>
>    - "Search in comments and strings" check box in the DartRenameDialog
>    does not work.
>
>   Yes, it should be removed.
  DAS does not provide this feature.


>
>    - All dynamic (not strict) usages are silently renamed although they
>    might be not related to the renamed symbol at all.
>
>   Right.
  These changes are marked as "potential" by DAS.
  Client are expected to show preview to the user and group these changes
separately.

  Can something like this be implemented in IDEA?
[image: Inline image 1]



>
>    - Undo message is not user friendly, for example "Undo Rename
>    Top-Level Function" (it is shown if rename affected multiple files).
>
>   Hm...
  AFAIK, it should be displayed even if the refactoring changes only one
file.
[image: Inline image 2]

  Do you mean that it should say "Rename Parameter 'name' to 'newName'"?

  Note, that IDEA uses gerund:
[image: Inline image 3]
...and for example OSX Finder uses names of commands, verbs:
[image: Inline image 4]


>
>    - Questionable DAS behavior: rename constructor from invocation new
>    Foo(). It behaves differently depending on the explicit default
>    constructor presence in class Foo. If there's no explicit constructor -
>    class is renamed. If there is explicit constructor - class is not renamed,
>    but default constructor is changed to named constructor Foo.newName.
>    I'd expect class to be renamed in both cases. When I see only new Foo()
>    I do not want to jump into class Foo declaration and check explicit
>    constructor presence, but I want refactoring to be predictable.
>
>   OK, I see the problem.
  We convert an offset into the PsiElement, and use it then to send
"declaration" offset to the server.
  But the problem that that in "class AAA{} main() {new AAA()}" navigation
from "new AAA()" is to the class declaration, even though DAS understand,
that it is a constructor rename.
  I will fix it by sending original offsets.

-- 
Konstantin Scheglov
Software Engineer
Google, Inc.

alexander-doroshko:
>>    - It is possible to rename symbol from SDK and 3rd party package (in Pub global cache). Shouldn't it be blocked by server?

>  Probably.
In Dart Editor we receive all the changes, and then warn the user that
the changes for files outside of the workspace will not be performed.
How does it work in IDEA?
Will return LocalFileSystem.getInstance().findFileByPath(filePath) a non-null file?

Virtual file will be not-null even if it is not in project. Looks like only `ProjectRootManager.getFileIndex().isInContent()` are allowed to be modified.
Also make sure to clear read-only status if needed using `FileModificationService`.

> I was not able to reproduce this, but yes, "initial status" is probably
better to show as a tooltip.
Any hint how to show it in IDEA?

`HintManagerImpl.getInstanceImpl().showEditorHint()`. We already use it in `AbstractDartFileProcessingAction`

> These changes are marked as "potential" by DAS.
Client are expected to show preview to the user and group these changes
separately.
Can something like this be implemented in IDEA?

Yes.  DartRenameDialog has Preview button (that doesn't work at the moment). We may force preview (that allows to select usages to rename) after pressing Rename button if there are unsure usages. 

> AFAIK, it should be displayed even if the refactoring changes only one
file.

Right. But when you undo multi-file change you see a dialog with this message as well.

> Do you mean that it should say "Rename Parameter 'name' to 'newName'"?

Yes I mean messages like in Java.

scheglov:
Most of the comments are addressed in https://github.com/JetBrains/intellij-plugins/pull/253.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/ServerRefactoring.java
```diff
@@ -0,0 +1,216 @@
...
+    RefactoringOptions options = getOptions();
+    final boolean success = DartAnalysisServerService.getInstance()
+      .edit_getRefactoring(kind, file, offset, length, validateOnly, options, new GetRefactoringConsumer() {
+        @Override
+        public void computedRefactorings(List<RefactoringProblem> initialProblems,
+                                         List<RefactoringProblem> optionsProblems,
+                                         List<RefactoringProblem> finalProblems,
+                                         RefactoringFeedback feedback,
+                                         SourceChange _change,
+                                         List<String> potentialEdits) {
```
alexander-doroshko: If I add `Thread.sleep(1000);` here to emulate slow response from server DartRenameDialog disables Refactor button forever without any error message after typing a new name. @scheglov can you please take a look at it?

FYI: Registry key now enables server driven resolution/find usages/rename in one shot, so you do not have to tweak plugin.xml to test it.

scheglov: I don't know :-(

```java
    myRefactoring.setListener(new ServerRefactoring.ServerRefactoringListener() {
      @Override
      public void requestStateChanged(final boolean hasPendingRequests, @NotNull final RefactoringStatus optionsStatus) {
        System.out.println("hasPendingRequests: " + hasPendingRequests);
        ApplicationManager.getApplication().invokeLater(new Runnable() {
          @Override
          public void run() {
            System.out.println("hasPendingRequests2: " + hasPendingRequests);
            myHasPendingRequests = hasPendingRequests;
            myOptionsStatus = optionsStatus;
            validateButtons();
          }
        });
      }
    });
```

  This code outputs only
```
hasPendingRequests: true
hasPendingRequests2: true
hasPendingRequests: false
```
  Where is `hasPendingRequests2: false`?
  Why this `invokeLater` is not called?
  Does it have something to do with throwing `ConfigurationException` from `canRun`?

  It seems to work as expected when I replace `ApplicationManager.getApplication().invokeLater` with `SwingUtilities.invokeLater`.

alexander-doroshko: Got it. Need to use Application.invokeLater with modality argument. 
In this case by default you schedule your Runnable to be run in non-modal state, e.g. after Rename dialog closes. ModalityState current() should work I think.

scheglov: `ModalityState.current()` does not work - we call `ServerRefactoringListener` not on the UI thread.
But `ModalityState.any()` works.

scheglov: See https://github.com/JetBrains/intellij-plugins/pull/272

PR #245, opened 2015-08-10
Use overridden elements from DAS.

scheglov:


scheglov:
The PR was updated. PTAL

alexander-doroshko:
Please add required tests.

scheglov:
Isn't it already covered by the existing test for DartMethodOverrideMarkerProvider?

alexander-doroshko:
Only tests that call `DartTestUtils.configureDartSdk()` with `true` as the last argument use real Analysis Server. All these tests are in `com.jetbrains.dart.analysisServer` package.

scheglov:
I've added tests.
PTAL

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -479,6 +488,21 @@ public static DartAnalysisServerService getInstance() {
...
   }
 
+  /**
+   * Returns {@link OverrideMember}s for the given file.
+   * Empty, but not null.
+   */
+  @NotNull
+  public List<OverrideMember> getOverrideMembers(@NotNull final VirtualFile file) {
+    synchronized (myOverrideData) {
+      List<OverrideMember> regions = myOverrideData.get(file.getPath());
```
alexander-doroshko: toSystemDependentName

Dart/src/com/jetbrains/lang/dart/ide/marker/DartServerOverrideMarkerProvider.java
```diff
@@ -0,0 +1,176 @@
...
+                                                return DartBundle.message("overrides.method.in", name, superClass.getName());
+                                              }
+                                              return DartBundle.message("implements.method.in", name, superClass.getName());
+                                            }
+                                          }, new GutterIconNavigationHandler<PsiElement>() {
+      @Override
+      public void navigate(MouseEvent e, PsiElement elt) {
+        List<DartComponent> superComponents = Lists.newArrayList();
+        if (superclassComponent != null) {
+          superclassComponent.add(superclassComponent);
```
alexander-doroshko: "Coded by my iPhone" with its f..ing autocompletion? :).

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -169,6 +170,14 @@ public void computedNavigation(String file, List<NavigationRegion> regions) {
     }
 
     @Override
+    public void computedOverrides(String file, List<OverrideMember> overrides) {
+      synchronized (myOverrideData) {
+        myOverrideData.put(file, overrides);
+      }
+      forceFileAnnotation(file);
```
alexander-doroshko: When we switch to DAS driven features we'll get daemon restarted and caches dropped 4 times on each typing. I'm not sure it's good. Need to think about it.

alexander-doroshko:
General questions:

- Will DAS provide 'overridden' info for operators? (would be great to have line markers for them too)

- Will DAS provide info for classes? We do not show line markers for super classes/interfaces (only for subclasses), but we could use this info in DartGotoSuperHandler

- Will DAS provide info for subclasses, overriding/implementiong methods/operators? Currently WebStorm is good enough itself wrt line markers pointing up the hierarchy, but not performant enough for line markers pointing down (aka slow like markers, DartImplementationsMarkerProvider). Would be great to switch our lame logic to DAS based one.

alexander-doroshko:
I'd suggest not to switch to DAS driven 'override' markers at the moment for 2 reasons:

- IntelliJ way is to show all super methods, not only the direct one. For example if foo() is implemented in all 3 classes A->B->C then line marker 'up' for A.foo() should suggest to jump both to B.foo() and C.foo() (see how it worked previously in Dart or in Java). Currently DAS gives information only about direct parent. Can this be changed?

- Line markers have almost the same functionality as "Go To Super" (Cmd+U) and "Go To Implementation" (Cmd+Alt+B). Difference is that Go To Super works also for classes where line markers  "up" are not shown. So switching to DAS driven logic should happen in 2 places simultaneously: `DartServerOverrideMarkerProvider and DartGotoSuperHandler`. When 'down' info appears: `DartImplementationsMarkerProvider and DartInheritanceIndex.DefinitionsSearchExecutor`

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Notifications often come quickly one after another.
So, we could queue restart requests for unique files and perform them with some delay, like 25 ms.

scheglov:
1. Yes, DAS provides "overridden" information for operators.
image

2. No, DAS does not have superclass notifications. I think we could use 
  BTW, it seems that the name is not quite correct. It says "Choose Subclass of BBB", but shows superclasses. Also, why does it show only direct superclass and implemented interfaces? It seems inconsistent with displaying overridden methods, which you ask to display for all the superclasses.
image

3. The notification for "down" classes and members are not implemented yet in DAS.

4. OK, you want to show all the super methods, not just the direct one. How about interfaces? Should methods of implemented interfaces be included? Only directly implemented interfaces, or also interfaces implemented by direct and indirect superclasses, as well as by other interfaces? Basically, do we need the whole set of "super" members or just some subset of it?

  It DAS we designed overridden members to be a quick indicator that there is a super member, and to use Type Hierarchy to go up or down.
image

5. About "Go To Super" - I cannot find this action in IDEA.
image

alexander-doroshko:
> Yes, DAS provides "overridden" information for operators.

Great! Now I see that it gives this info, but we do not use it for line markers yet.

> BTW, it seems that the name is not quite correct. It says "Choose Subclass of BBB", but shows superclasses.

Dart line markers and Go To Super / Go To Implementations are buggy. This is one of the bugs. See how it works for Java, we should do the same.

> Also, why does it show only direct superclass and implemented interfaces? It seems inconsistent with displaying overridden methods, which you ask to display for all the superclasses.

You are right, behavior is a bit different for Go To superclass and Go To supermethod. I'm afraid I do not know why it is done this way, but I see that it works this way for Java as well. When we do not have strong reasons to do in a different way I think we should do the same as in Java. But this inconsistency looks strange for me indeed, it is quite difficult to jump to grangparent (too many keypresses), so I'm going to talk to Java guys about it. This is the case where we can do not as in Java and suggest the whole set of superclasses/interfaces if we think it is better for UX.

> OK, you want to show all the super methods, not just the direct one. How about interfaces? Should methods of implemented interfaces be included? Only directly implemented interfaces, or also interfaces implemented by direct and indirect superclasses, as well as by other interfaces? Basically, do we need the whole set of "super" members or just some subset of it?

I think we should do as in Java where the whole set of super members is suggested to jump to.

> About "Go To Super" - I cannot find this action in IDEA.

It's on your screenshot: Go To Super Method (up) and Go To Implementations (down). Also it's under Navigate main menu item. Actions work when caret is on the class/method name (may be when caret is anywhere on the declaration line as well - it depends on implementation). Default shortcuts on Mac are Cmd+U and Cmd+Alt+B, but you seem to use non-default keymap. I see a cosmetical issue: action Go To Super Method doesn't change its name to Go To Super Class when needed, but it still works fine. I've filed an [issue](https://youtrack.jetbrains.com/issue/IDEA-144485), feel free to vote

alexander-doroshko:
Looks like I could be wrong saying that line marker 'up' and Go To Super Method show full hierarchy of super methods. Playing more with Java I see that only direct parent(s) are suggested. Please play a bit yourself with Java. If so, then probably current 'override' info from DAS is sufficient and we only need to enable line markers for operators and to switch DartGotoSuperHandler to DAS.

scheglov:
Yes, it seems that IDEA does not allow to go to super-super methods.
image

And, in the same way, it does not allow you to go to the super-super class or super-super interfaces.
Only to the direct superclass and interfaces that the class directly implements.
image

As I can see, we have the same behavior for super methods with this PR.
I don't think that it is the best one, but at least it is consistent.

I think we could implement `DartGotoSuperHandler ` using `search.getTypeHierarchy` with a (new) request flag for requesting only superclasses. It currently returns all the super classes/interfaces. But we can determine direct ones if we decide to display less information then we could :-)

Operators are not easy to implement.
By some reason they are not `DartComponent`(s) as methods.
This would require more changes than I'd like to make in one PR.

What if we make `DartServerOverrideMarkerProvider` disabled by default, land the PR and continue working on it in subsequent PRs?

alexander-doroshko:
sounds good

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: I will update `computedOverrides` instead, as we do for all the other notifications.

Dart/src/com/jetbrains/lang/dart/ide/marker/DartServerOverrideMarkerProvider.java
scheglov: Ouch! It updates the file itself.
Fixed.

scheglov:
I've updated the PR.
The tests are failing now though, because the feature is disabled.
Should I comment out the tests then?

alexander-doroshko:
yes, just make the test class abstract

scheglov:
done

alexander-doroshko:
DartAnalysisServerHighlightingTest#testErrorsAfterEOF now hangs for me and I get an exception:
```
java.lang.AssertionError
	at com.intellij.codeInsight.daemon.impl.FileStatusMap.assertAllowModifications(FileStatusMap.java:179)
	at com.intellij.codeInsight.daemon.impl.FileStatusMap.markAllFilesDirty(FileStatusMap.java:170)
	at com.intellij.codeInsight.daemon.impl.DaemonCodeAnalyzerImpl.restart(DaemonCodeAnalyzerImpl.java:506)
	at com.jetbrains.lang.dart.analyzer.DartAnalysisServerService.forceFileAnnotation(DartAnalysisServerService.java:270)
	at com.jetbrains.lang.dart.analyzer.DartAnalysisServerService.access$500(DartAnalysisServerService.java:77)
	at com.jetbrains.lang.dart.analyzer.DartAnalysisServerService$1.computedOverrides(DartAnalysisServerService.java:178)
	at com.google.dart.server.internal.BroadcastAnalysisServerListener.computedOverrides(BroadcastAnalysisServerListener.java:123)
	at com.google.dart.server.internal.remote.processor.NotificationAnalysisOverridesProcessor.process(NotificationAnalysisOverridesProcessor.java:42)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processNotification(RemoteAnalysisServerImpl.java:639)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processResponse(RemoteAnalysisServerImpl.java:667)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.access$1(RemoteAnalysisServerImpl.java:665)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl$ServerResponseReaderThread.run(RemoteAnalysisServerImpl.java:151)
Exception in thread "ServerResponseReaderThread" java.lang.AssertionError: null
	at com.intellij.testFramework.LoggedErrorProcessor.processError(LoggedErrorProcessor.java:56)
	at com.intellij.testFramework.TestLogger.error(TestLogger.java:67)
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:130)
	at com.jetbrains.lang.dart.analyzer.DartAnalysisServerService$6.logError(DartAnalysisServerService.java:403)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl$ServerResponseReaderThread.run(RemoteAnalysisServerImpl.java:167)
```

scheglov:
Hm...
I had to add `((CodeInsightTestFixtureImpl)myFixture).canChangeDocumentDuringHighlighting(true);` to the `DartServerOverrideMarkerProviderTest`. I don't know if this indicates a real problem or just a peculiarity of test fixtures.

alexander-doroshko:
I had to add canChangeDocumentDuringHighlighting(true) to some other Dart server tests as well. It doesn't look like a real problem. The need in it may go away if we optimize multiple forceFileAnnotation() calls.
Side note: forceFileAnnotation() calls clearCache(). I guess we need to clear cache on navigation data only, but not highlightoing/overrides?

PR #246, opened 2015-08-14
Experimental code to format using dartfmt

stevemessick:
@denofevil @alexander-doroshko This is NOT a real pull request. Please do not merge this code; you will break the build.

I'd just like some feedback from JetBrains about the viability of this approach to formatting code. I have not figured out how to restrict it to only run during paste; any help would be appreciated. More importantly, uncommitted/committed documents are causing problems. See the comment on line 68 of DartPreFormatProcessor.

Would using postFormatProcessor be any better?

I suspect we will need to stop using the external formatter if the text to be formatted exceeds some value. TBD.

One more thing:
FileUtil.toSystemDependentName(myFile.getVirtualFile().getPath());
That would be the preferred way to get the path to the file to be formatted, I think. It doesn't work during paste because the file content has not had the pasted text added to it yet. Any way around that? I added a lot of otherwise useless code to create a temporary file to give to the formatter.

And yes, there is an outstanding request to enhance the formatter API to specify a region of text to be formatted. No estimate on when that will be implemented.


Dart/src/com/jetbrains/lang/dart/ide/formatter/DartFormattingModelBuilder.java
```diff
@@ -19,8 +23,15 @@ public FormattingModel createModel(@NotNull final PsiElement element, @NotNull f
     // element can be DartFile, DartEmbeddedContent, DartExpressionCodeFragment
     final PsiFile psiFile = element.getContainingFile();
     final ASTNode rootNode = psiFile instanceof DartFile ? psiFile.getNode() : element.getNode();
-    final DartBlock rootBlock = new DartBlock(rootNode, null, null, settings);
-    return new DocumentBasedFormattingModel(rootBlock, element.getProject(), settings, psiFile.getFileType(), psiFile);
+    final DartBlock rootBlock;
+    Object wasFormatted = rootNode.getUserData(DartPreFormatProcessor.FORMAT_MARK);
+    if (wasFormatted == DartPreFormatProcessor.FORMAT_MARKER) {
```
stevemessick: Yes, but they have not been included in this experimental code. Some unit tests also cause that condition.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private CodeStyleSettings mySettings;
+  private PsiFile myFile;
+  private Project myProject;
+  private int selectionOffset;
+
+  @NotNull
+  @Override
+  public TextRange process(@NotNull ASTNode element, @NotNull TextRange range) {
+    // TODO return range if not in a paste operation
```
alexander-doroshko: Do you think we should use server logic for Paste only? I'd say either we try to substitute IDE's formatter everywhere or nowhere. And there are dozens of places where IDE's formatter is called:
- explicit reformat action (with selection or without)
- paste
- reformat before commit if corresponding check box is selected (should reformat only VCS changed text)
- all logic that modifies PSI, including:
-- quick fixes
-- intention actions
-- code generation
- dozens of places that modify files via document like live templates, and other.

stevemessick: AFAIK we can only distinguish explicit invocation from all other invocations. We have no way to tell if the formatter was invoked from paste, refactoring, or whatever -- according to email I got from Dennis after I made this PR. Initially, I wanted to start small and get some idea of performance, but I don't think that is possible, so that TODO is invalid.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.io.*;
+import java.util.Collection;
+import java.util.List;
+
+public class DartPreFormatProcessor implements PreFormatProcessor {
+  public static Key<Boolean> FORMAT_MARK = new Key<Boolean>("FORMAT_MARK");
+  public static Boolean FORMAT_MARKER = Boolean.TRUE;
+
+  private CodeStyleSettings mySettings;
```
alexander-doroshko: Extensions like PreFormatProcessor must not contain any fields and must not remember any state. Instance of this DartPreFormatProcessor is created once per application and is kept in memory forever. Project might be closed, but with these fields we get tons of memory leaked. This class should work like a helper. All needed arguments are passed to process() method and after returning from process() there must be no trace that it was ever called.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+  private PsiFile myFile;
+  private Project myProject;
+  private int selectionOffset;
```
alexander-doroshko: `CodeStyleSettings` and `int` are relatively harmless to remember, but `PsiFile` and `Project` is something that we must always be aware about: its leakage extremely affects memory allocation.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
stevemessick: That's good info and will be really useful to know when I start working on a production version, if I ever do. However, the purpose of this PR is to figure out why the code doesn't work. For example, DartServerQuickFixTest.testCreateMethodInAnotherFile() produces totally mangled code when this patch is applied. I have already tried a number of tricks Dennis suggested in email and those did not help. I believe you were on the distro list of that email if you need to know what he suggested.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+    myProject = psiElement.getProject();
+    mySettings = CodeStyleSettingsManager.getSettings(myProject);
+    final FileViewProvider viewProvider = myFile.getViewProvider();
+    final Document document = viewProvider.getDocument();
+    if (document == null) return range;
+
+    final String code = reformat(document.getText(), range);
+    if (code == null) return range;
+    final RangeMarker marker = document.createRangeMarker(range.getStartOffset(), range.getEndOffset(), true);
+    document.replaceString(0, document.getTextLength(), code);
```
alexander-doroshko: I'm afraid this RangeMarker can't survive with any reasonable range after total document text replacement.
The task to keep range after reformat doesn't look trivial. Probably PsiElementPointers can help, after reformatting all PsiElementPointers should survive. Althouh PsiElements change on reparse, in general PSI tree remains the same (because only whitespaces are changed), so this is where PsiElementPointer can survive and provide updated range. That's just a guess.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+
+    element.putUserData(FORMAT_MARK, FORMAT_MARKER);
+    PsiDocumentManager documentManager = PsiDocumentManager.getInstance(myProject);
+    // Without this commit, an error is thrown by some tests.
+    // With it, some files get garbled.
+    documentManager.commitDocument(document);
```
alexander-doroshko: What files and how are they garbled?
By Platform API design PSI must be valid after returning from process(). As you modify Document you must call commitDocument() to reparse PSI for the updated text. So calling commitDocument() is valid here and if it causes any problems they should be addressed.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
stevemessick: Perhaps I can get the analysis server API changed to return the modified range after reformatting. The formatter keeps track of it but the API doesn't include the range argument.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
stevemessick: Please see my earlier reply that mentions DartServerQuickFixTest.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
alexander-doroshko: Sounds promising. As far as I understand this will allow us to easily reformat selection even though server can reformat the whole file only, am I right?

stevemessick: That would require some work in the formatter.
https://github.com/dart-lang/dart_style/issues/92

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
alexander-doroshko: Yep, I see the test failing, and also can reproduce the same in runtime. Sample code 
```
main() {
  Iterable b = doSomething(2+2==4, 1+1, 2*2.0, "", [1], {1:2}, #*, null);
}
```
Quick fix starts a live template, template asks to reformat range, after DartPreFormatProcessor range for template variable becomes broken which seems to cause the mangled code.
image
I'm afraid nobody can answer why that happens without some time (may be hours) of debugging. This might be a bug in template engine, template misuse, bug in PreformatProcessor or anything else. As an experiment we can try to use PostFormatProcessor instead of PreFormatProcessor.

Dart/src/com/jetbrains/lang/dart/ide/formatter/DartPreFormatProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public String reformat(String code, TextRange range) {
+    final DartAnalysisServerService.FormatResult formatResult;
+    try {
+      // While running a paste operation the contents of the VFS are not in sync with the document.
+      // Put the document content in a temporary file to be reformatted.
+      String parentPath = parentFilePath();
+      if (parentPath == null) return null;
+      final File tempFile = FileUtil.createTempFile(new File(parentPath), "fmt", ".dart");
```
alexander-doroshko: Disk operations are potentially slow and look like an overhead for this task. DAS can work with content that is not saved on disk. DartAnalysisServerService#updateFilesContent sends current contents of all unsaved Documents (modified since previous sync) to the server. Other options is to extend server API so that its edit_format() could also accept file contents instead of a file path.

stevemessick: That's why I initially only wanted to run the formatter on paste operations. The problem is that the formatter requires a disk file. It was designed as a unix shell command.

alexander-doroshko: Formatter integrated into Analysis server doesn't require file to be saved, the same as all other server-based operations like highlighting and reference resolution. when you right click Dart file and select Reformat With Dart Style - file is not saved. It is enough to send actual content to the server, and that's what updateFilesContent() method does.

alexander-doroshko:
In general this is a good try. May be not that many lines from this PR will survive but I believe some day we'll switch to DAS-based formatter completely (not only for pasting or other individual cases). If DAS API will help us to work with selected ranges - that'd be awesome.

Mangled code in case of quick fix that runs a live template is a problem, but I believe that committing a document in PreFormatProcessor is allowed. I hope we'll be able to find the cause.

PostFormatProcessor or PreFormatProcessor - I do not know. May be no difference. May be one of them will appear to be more convenient.

> I suspect we will need to stop using the external formatter if the text to be formatted exceeds some value. TBD.

Formatting big file (15 000 LOC) file takes reasonable time when formatting the whole file. But we'll probably need better performance when formatting small parts of it. I hope when selection formatting will be done at server side then small part formatting of a big file will be faster then the whole file formatting.

Temporary files or saving whatever should not be needed, updateFilesContent() should do the trick. 

Performance may potentially become a problem, but we'll be able to measure it only when everything is implemented.

There's one problem that I do not know solution for yet. IDE's formatter model is used not only for formatting code. It is also used for indent calculation:
- caret placement after Enter
- caret placement when pressing End on empty line
- 'smart backspace'
- and may be other cases that I do not remember now

If we change formatter model to a single block covering the whole file and doing nothing then we loose all this functionality. At the moment I do not know a solution.

stevemessick:
Thanks for the careful review. I'll keep experimenting and see what I can come up with. I tried updateFilesContent() but that didn't help. Next up is to see what post-format processing does. I suspect nothing will change but am curious to see if I'm wrong. And I agree that we'll have to have more information about when to run the external formatter versus using the internal one. It has been a very long time since I've looked at the java reflection API but that might have something we can use.

alexander-doroshko:
After Konstantin's [commit](https://github.com/JetBrains/intellij-plugins/pull/249) the problem of mangled code seems to go away (he did some changes wrt live templates started by quick fixes).

PR #247, opened 2015-08-17
Use continuation indent for incomplete statements

stevemessick:
@denofevil @alexander-doroshko A newline at the end of an incomplete statement should indent the next line at continuation level if no other rule applies.


stevemessick:
@denofevil Thanks for catching that. I fixed the code (and reformatted) and added several more tests.

PR #248, opened 2015-08-18
Improve wrapping of long class declarations.

stevemessick:
@alexander-doroshko @denofevil The bulk of the changes have to do with wrapping *implements* and *with* clauses properly. Getting the keywords themselves where they belong was extra hard. See the comment on line 380 of DartSpacingProcessor. Could that be a bug? It doesn't make sense.


alexander-doroshko:
pushed

PR #251, opened 2015-08-25
Use the version of the DAS client that never blocks on sending requests.

scheglov:
Also call ProgressManager.checkCanceled() while waiting for responses.

Should we convert all the server APIs invocations to using ProgressManager.checkCanceled()?


alexander-doroshko:
Pushed.
Yes, everywhere where we wait for server response we should checkCancelled() unless we already do.

PR #252, opened 2015-08-25
Improve Dart string wrapping

stevemessick:
@alexander-doroshko Typing <return> in the middle of a string will insert proper syntax to make two concatenating strings (similar to java).

Fixes https://youtrack.jetbrains.com/issue/WEB-12608

This may have a merge conflict in DartIndentProcessor with another formatter-related pull request that has not yet been merged. I'm not sure how clever git is; there actually is no conflict.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
@@ -1,0 +1,56 @@
+package com.jetbrains.lang.dart.ide.editor;
+
+import com.intellij.codeInsight.editorActions.enter.EnterHandlerDelegateAdapter;
+import com.intellij.lang.ASTNode;
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.editor.actionSystem.EditorActionHandler;
+import com.intellij.openapi.util.Ref;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.tree.IElementType;
+import com.jetbrains.lang.dart.DartTokenTypes;
+import org.jetbrains.annotations.NotNull;
+
+public class DartEnterInStringHandler extends EnterHandlerDelegateAdapter {
+
+  @Override
+  public Result preprocessEnter(@NotNull final PsiFile file,
+                                @NotNull final Editor editor,
+                                @NotNull final Ref<Integer> caretOffsetRef,
+                                @NotNull final Ref<Integer> caretAdvanceRef,
+                                @NotNull final DataContext dataContext,
+                                final EditorActionHandler originalHandler) {
+    int caretOffset = caretOffsetRef.get().intValue();
+    int caretAdvance = caretAdvanceRef.get().intValue();
+    PsiElement psiAtOffset = file.findElementAt(caretOffset);
+    if (psiAtOffset == null || psiAtOffset.getTextOffset() > caretOffset) {
+      return Result.Continue;
+    }
+    Document document = editor.getDocument();
+    ASTNode token = psiAtOffset.getNode();
+    IElementType type = token.getElementType();
+    if (type == DartTokenTypes.REGULAR_STRING_PART) {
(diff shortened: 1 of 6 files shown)


Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
```diff
@@ -0,0 +1,56 @@
...
+    PsiElement psiAtOffset = file.findElementAt(caretOffset);
+    if (psiAtOffset == null || psiAtOffset.getTextOffset() > caretOffset) {
+      return Result.Continue;
+    }
+    Document document = editor.getDocument();
+    ASTNode token = psiAtOffset.getNode();
+    IElementType type = token.getElementType();
+    if (type == DartTokenTypes.REGULAR_STRING_PART) {
+      token = token.getTreeNext();
+      type = token.getElementType();
```
alexander-doroshko: `getTreeNext()` may return `null`

stevemessick: Even when type == DartTokenTypes.REGULAR_STRING_PART? I thought it would at least have an error node following. How can I write a test that produces null there?

Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
```diff
@@ -0,0 +1,56 @@
...
+      // At this point we might have type == TokenType.ERROR_ELEMENT if the string is unterminated.
+      // We're ignoring that case since we don't have a rich enough set of signals to deduce where or
+      // even if an additional quote should be added. Since strings are by default terminated it should be rare.
+    }
+    if (type == DartTokenTypes.CLOSING_QUOTE) {
```
alexander-doroshko: this handler doesn't work for strings with interpolation if caret is before interpolation:
`print("one<caret>, $a, three");`

Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
```diff
@@ -0,0 +1,56 @@
...
+      // The final effect is to insert matching close-quote, newline, indent, matching open-quote.
+      char literalStart = token.getText().charAt(0);
```
alexander-doroshko: There are not only single-character quotes in Dart.
I guess raw single quoted strings `r'foo'`, `r"foo"` should be handled the same as normal single quoted ones.
For triple quoted strings (both raw and normal) we shouldn't add quotes at the end/beginning of the line as these strings are multiline already.
DartPsiImplUtil#unquoteDartString might be of help (at least to reuse constants).

Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
alexander-doroshko: Well, you are right, may be in this case null can't appear. Personally I prefer to add an extra null-check rather than think about data flow and hope that everyone who will be updating lexer/parser/DartEnterInStringHandler in future won't forget to add null-check when null become real.

alexander-doroshko:
Pushed.
The most major issue found is that multi-line strings are corrupted on Enter.

Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
```diff
@@ -0,0 +1,56 @@
...
+public class DartEnterInStringHandler extends EnterHandlerDelegateAdapter {
+
+  @Override
+  public Result preprocessEnter(@NotNull final PsiFile file,
+                                @NotNull final Editor editor,
+                                @NotNull final Ref<Integer> caretOffsetRef,
+                                @NotNull final Ref<Integer> caretAdvanceRef,
+                                @NotNull final DataContext dataContext,
+                                final EditorActionHandler originalHandler) {
+    int caretOffset = caretOffsetRef.get().intValue();
```
alexander-doroshko: I've added language check as the first thing to do
`    if (file.getLanguage() != DartLanguage.INSTANCE && !HtmlUtil.isHtmlFile(file)) return Result.Continue;`
and also added a test for Dart in HTML

PR #253, opened 2015-08-27
Improvements for the Rename refactoring integration.

scheglov:
1. @NotNull and @Nullable annotations.
2. More tests.
3. Rewritten dialog.
4. Use isInContent() and prepareVirtualFilesForWrite().
5. The fix for renaming constructors.

What is not done:
1. Any handling of potential changes. I don't know how to preview it.
2. Undo/Redo message - we get it from DAS as SourceChange.getMessage(), should be changed there.


scheglov:
rebased

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartRenameHandler.java
```diff
@@ -55,14 +65,7 @@ public void invoke(@NotNull Project project, @NotNull PsiElement[] elements, Dat
...
   public boolean isAvailableOnDataContext(DataContext dataContext) {
-    final PsiElement element = CommonDataKeys.PSI_ELEMENT.getData(dataContext);
-    if (element instanceof DartNamedElement) {
-      return true;
-    }
-    if (element instanceof DartComponent) {
-      return true;
-    }
-    return false;
+    return true;
```
alexander-doroshko: This extension point doesn't support `lang` attribute in `plugin.xml`, so this `return true` breaks renaming in all languages in all JetBrains IDEs. I'll fix.

PR #254, opened 2015-08-31
Correctly handle raw and interpolated strings.

stevemessick:
@alexander-doroshko The title says it all, I think.


alexander-doroshko:
Pushed. 
Enter in triple-quoted strings also seem to work better now (i.e. doing nothing). Can you please add tests for it? The only questionable thing about Enter in multiline strings is whether we want to use standard indent or may be return `Result.DefaultSkipIndent` in this case.

I've found some tricky cases with interpolation that could be improved:
`var a = "$x and <caret> also $y";`
`var a = "see <caret>$y";`
`var a = "see <caret> also ${y}";`

PR #255, opened 2015-09-01
Add more tests and simplify logic of string wrapping

stevemessick:
@alexander-doroshko This addresses review comments by simplifying and generalizing the logic, and adding additional tests. It is much more sensible now IMO.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
@@ -31,35 +31,28 @@ public class DartEnterInStringHandler extends EnterHandlerDelegateAdapter {
     Document document = editor.getDocument();
     ASTNode token = psiAtOffset.getNode();
     IElementType type = token.getElementType();
+    if (type == DartTokenTypes.RAW_TRIPLE_QUOTED_STRING) {
+      return Result.DefaultSkipIndent; // Multiline string gets no indent
+    }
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/ide/editor/DartEnterInStringHandler.java
```diff
@@ -31,37 +31,30 @@ public Result preprocessEnter(@NotNull final PsiFile file,
...
+    }
     if (type == DartTokenTypes.RAW_SINGLE_QUOTED_STRING) {
       char quote = token.getText().charAt(1);
       breakString("r" + quote, String.valueOf(quote), caretOffsetRef, caretAdvanceRef, document);
       return Result.Default;
     }
-    if (type == DartTokenTypes.REGULAR_STRING_PART) {
-      token = token.getTreeNext();
-      if (token == null) return Result.Continue; // Can't happen with current grammar.
+    while ((token = token.getTreeParent()) != null) {
```
alexander-doroshko: Do you think we can optimize it by checking not more than 2 parents up the hierarchy?

stevemessick: I think we have to do something. The current code is buggy when trying to
split embedded interpolations.

var b = "see ${'${a}'} and more";

If you type a Return when the cursor is before the 'a' in '{a}' you get
incorrect code. It should insert a newline and indent but no additional
quotes.

PR #256, opened 2015-09-02
Hook up dartfmt for experimenting

stevemessick:
@alexander-doroshko Here's another attempt at using dartfmt as the incremental formatter. Unfortunately, my git-fu is weak and I ruined the previous branch so I decided to start over. There are still some useful comments in the old PR so I have not deleted it yet.

This passes the same tests that pass without it. (I reported two new failures earlier.) Notably, as you mentioned, the test that previously mangled the code now works. So does one that should have failed (DartServerQuickFixTest.testCreateClass()), so I wonder if this formatter is even being called by the tests. (I just checked with the debugger. The breakpoint in DartPreFormatProcessor was not hit so I guess that test no longer uses the formatter.)

You can ignore DartPostFormatProcessor. I have not tested it, but I haven't deleted it since this is still work-in-progress.

I tried the Extract Method refactoring in the runtime IDE and it did hit the breakpoint, so at least that works. The performance is horrible. It took about 7 seconds to extract the method when *Bar* is selected in this code.
```dart
library foo;

part 'eee.dart';

main() {
  Iterable b = new Bar().doSomething(2+2==4, 1+1, 2*2.0, "", [1], {1:2}, #*, null);
}
```

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -59,6 +59,8 @@
     <lang.implementationTextSelectioner language="Dart"
                                         implementationClass="com.jetbrains.lang.dart.ide.DartImplementationTextSelectioner"/>
     <lang.formatter language="Dart" implementationClass="com.jetbrains.lang.dart.ide.formatter.DartFormattingModelBuilder"/>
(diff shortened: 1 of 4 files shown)


alexander-doroshko:
Sorry, didn't have time to play today.

PR #257, opened 2015-09-02
Make double-click select words in line doc comments

stevemessick:
@alexander-doroshko Fixes https://youtrack.jetbrains.com/issue/WEB-17046

### DIFF
--- a/Dart/testSrc/com/jetbrains/lang/dart/ide/editor/DartWordSelectionTest.java
+++ b/Dart/testSrc/com/jetbrains/lang/dart/ide/editor/DartWordSelectionTest.java
@@ -1,0 +1,56 @@
+package com.jetbrains.lang.dart.ide.editor;
+
+import com.intellij.testFramework.fixtures.CodeInsightTestUtil;
+import com.jetbrains.lang.dart.DartCodeInsightFixtureTestCase;
+import com.jetbrains.lang.dart.util.DartTestUtils;
+
+public class DartWordSelectionTest extends DartCodeInsightFixtureTestCase {
+
+  @Override
+  public void setUp() throws Exception {
+    super.setUp();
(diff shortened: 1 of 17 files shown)


alexander-doroshko:
I think it should be testable. There must be tests for other `ExtendWordSelectionHandler`'s to look at.

stevemessick:
If you know where to put the test I'll be glad to add one. I don't know how to test double-click's in this system.

alexander-doroshko:
Just a guess based on extension name: test similar to one of `*WordSelectionTest` could do the trick.

alexander-doroshko:
Note that your extension affects not only double click but also incremental 'word selection' (Alt+Up / Alt+Down) - very handy IDE feature!

stevemessick:
@alexander-doroshko PTAL I added tests for selecting words in all kinds of comments. Most of the tests are not in scope of this base change, but apparently they weren't being tested.

Dart/testSrc/com/jetbrains/lang/dart/ide/editor/DartWordSelectionTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.intellij.testFramework.fixtures.CodeInsightTestUtil;
+import com.jetbrains.lang.dart.DartCodeInsightFixtureTestCase;
+import com.jetbrains.lang.dart.util.DartTestUtils;
+
+public class DartWordSelectionTest extends DartCodeInsightFixtureTestCase {
+
+  @Override
+  public void setUp() throws Exception {
+    super.setUp();
+    DartTestUtils.configureDartSdk(myModule, getTestRootDisposable(), true);
```
alexander-doroshko: I'll change the last argument to `false`, we do not need real Dart SDK and real Analysis Server for these tests

alexander-doroshko:
Pushed, thank you for the tests!

PR #258, opened 2015-09-02
Use DAS search.getTypeHierarchy to implement 'Goto super class/member'.

scheglov:


alexander-doroshko:
Can you please add a test (disabled so far)?
You may look at DartGotoSuperTest, though I do not know how good is this test.

scheglov:
PTAL
I've added tests.

Although it is not quite complete - we cannot test selecting from the list if there are multiple super- classes/interfaces. OTOH, `DartGotoSuperTest ` also does not test it.

Dart/src/com/jetbrains/lang/dart/ide/actions/DartServerGotoSuperHandler.java
```diff
@@ -0,0 +1,110 @@
...
+        supers.add(itemComponent);
+      }
+    }
+    // prepare the title
+    final String title;
+    if (inComponentType == DartComponentType.CLASS) {
+      title = "Superclasses of " + inComponent.getName();
+    }
+    else {
+      title = "Overridden members of superclasses";
```
alexander-doroshko: I'll change titles to be the same or similar to what we have in Java

alexander-doroshko:
Server bug? Subclass C is suggested for B
image
Even more strange: if add a blank line, then C is not suggested any more and we silently jump to A as expected.
image

alexander-doroshko:
This is the simplest case where Go To Super jumps to subclass.
image

alexander-doroshko:
The same as for line markers: would be great to make it work for operators (I guess server already provides this info).

alexander-doroshko:
General note. As we noticed in a different thread, in Java both line markers and Go To Super jumps to direct parent(s) only, not to grandparents and the rest of the hierarchy.
As for the 'down' line markers and Go To Implementation action - I see that in Java it suggests not only direct children but the whole set of children.
I asked a couple of Dart aware colleagues if they think that in Dart we should make it differently and they said that they do not see reasons to make it different than in Java.
With current DAS-based implementation line markers jump to direct supers only, but Go To Super suggests the whole set of supers. Do you think this is justified?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -837,6 +838,34 @@ public void onError(RequestError requestError) {
     }
   }
 
+  public List<TypeHierarchyItem> search_getTypeHierarchy(@NotNull final String _filePath, final int offset, final boolean superOnly) {
```
alexander-doroshko: `superOnly` flag not supported by server yet?

scheglov: Yes, I'm working on the PR that will include the new DAS client, which supports this flag.
Support of it in DAS was committed yesterday.

This will fix bugs with jumping to subclasses.

scheglov:
1. Unfortunately we need to change DPI's AST to support operators. DAS provides this information.

2. I think it is interesting to see all the superclasses and all the definitions of a class member in superclasses. This helps in understanding a class. I don't think it there are usually *too* many of them. That said, I don't see much reasons to go against the IDEA style here, whether it is good or bad. So, I will limit amount of information we provide to the user.

PR #259, opened 2015-09-03
Inline refactoring handler.

scheglov:


Dart/testSrc/com/jetbrains/dart/analysisServer/DartInlineLocalRefactoringTest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @NotNull
+  private ServerInlineLocalRefactoring createInlineLocalRefactoring(String filePath, String atString) {
+    ((CodeInsightTestFixtureImpl)myFixture).canChangeDocumentDuringHighlighting(true);
+    final PsiFile psiFile = myFixture.configureByFile(filePath);
+    myFixture.doHighlighting(); // make sure server is warmed up
+    // find the Element to rename
+    final Document document = myFixture.getDocument(psiFile);
+    final int offset = document.getText().indexOf(atString);
```
alexander-doroshko: I'll change it to `getEditor().getCaretModel().getOffset()` and add `<caret>` markers to test data.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
```diff
@@ -0,0 +1,226 @@
...
+    final VirtualFile virtualFile = psiFile.getVirtualFile();
+    final String filePath = virtualFile.getPath();
+    // prepare navigation regions
+    final int offset = editor.getCaretModel().getOffset();
+    final List<PluginNavigationRegion> navigationRegions = DartAnalysisServerService.getInstance().getNavigation(virtualFile);
+    // find the navigation region
+    for (PluginNavigationRegion region : navigationRegions) {
+      if (region.getOffset() <= offset && offset <= region.getOffset() + region.getLength()) {
+        final List<PluginNavigationTarget> targets = region.getTargets();
+        final String kind = targets.get(0).getKind();
```
alexander-doroshko: Can `targets` be empty?

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      refactoring = new ServerInlineMethodRefactoring(context.filePath, context.offset, 0);
+    }
+    // validate initial status
+    {
+      final RefactoringStatus initialConditions = refactoring.checkInitialConditions();
+      if (showMessageIfError(editor, initialConditions)) {
+        return;
+      }
+    }
```
alexander-doroshko: Is this (and one more below) extra braces `{block}` kind of coding style or just a typo? I was taught to add empty lines to make code more readable instead.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/ServerRefactoring.java
```diff
@@ -73,13 +73,16 @@ public void run(@NotNull ProgressIndicator indicator) {
...
     RefactoringStatus result = new RefactoringStatus();
     result.merge(optionsStatus);
     result.merge(finalStatus);
     return result;
   }
 
-  @NotNull
+  @Nullable
   public RefactoringStatus checkInitialConditions() {
     ProgressManager.getInstance().run(new Task.Modal(null, "Checking initial conditions", true) {
```
alexander-doroshko: Cosmetics: messages like this give no clue to the user about what IDE is doing and what will fail if user clicks Stop. IMHO refactoring name would be better, for example "Inline Variable". May be even better message could be composed.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
```diff
@@ -0,0 +1,226 @@
...
+    final InlineRefactoringContext context = findContext(editor);
+    if (context == null) {
+      return;
+    }
+    // create refactoring
+    final ServerRefactoring refactoring;
+    if (ElementKind.LOCAL_VARIABLE.equals(context.kind)) {
+      refactoring = new ServerInlineLocalRefactoring(context.filePath, context.offset, 0);
+    } else {
+      refactoring = new ServerInlineMethodRefactoring(context.filePath, context.offset, 0);
```
alexander-doroshko: A mess with system dependent/independent slashes again. Somewhere we convert paths inside `DartAnalysisServerService` methods, somewhere - outside. I had to tweak to make it work on Windows.

Suggestion: work with system independent paths everywhere outside `DartAnalysisServerService` methods. Convert paths to system dependent only locally in the methods that call server. Convert paths back to system independent as soon as result is received. Make sure no fields contain system dependent slashes and no parameters expect to receive system dependent ones and no method returns system dependent path.

alexander-doroshko:
Inlining method/function/var that has no usages - code silently disappears. I'd expect to be notified that there are no usages and either veto refactoring or make user understand what will happen.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
```diff
@@ -0,0 +1,226 @@
...
+  InlineRefactoringContext(VirtualFile virtualFile, String filePath, int offset, String kind) {
+    this.virtualFile = virtualFile;
+    this.filePath = filePath;
+    this.kind = kind;
+    this.offset = offset;
+  }
+}
+
+class InlineMethodDialog extends InlineOptionsDialog {
+  public static final String REFACTORING_NAME = RefactoringBundle.message("inline.variable.title");
```
alexander-doroshko: `"inline.method.title"`

alexander-doroshko:
Clicking Preview runs refactoring the same as Refactor button
image

alexander-doroshko:
In general it's awesome!

Related: is it safe to inline methods that participate in hierarchy?

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
scheglov: It was undefined.
I'm changing the DAS spec to explicitly document that it cannot be empty.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
scheglov: It's a coding style.

I prefer keeping separate pieces of functionality (but not so much separate to justify extracting them as separate methods) in their own block and leak outside only variables with final results (if any), without exposing the rest of the method to all the temporary variables.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/ServerRefactoring.java
scheglov: OK
I'm going to move the description into the dialog and put the refactoring name into the title.

image

image

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartInlineHandler.java
scheglov: I agree, this is a good idea.

I think we already convert every path to system independent in notification handlers.

I looked through the server requests and updated them to convert paths to system dependent.

scheglov:
**Errors**
  Unfortunately it's unavoidable - there might be some error messages from DAS which we want to display. We need somehow decorate them to make them reasonable to users. DAS sends three kinds of errors:

1. Fatal errors - the refactoring cannot be applied at this position at all - unresolved unit, to identifier at the specified offset, etc

2. Errors - may be a problem, but may be also something what user wants to do anyway, such as shadow usage of a method, remove unused method, etc.

3. Warnings - just warnings, probably nothing too important, like a not recommended variable name.

If possible, DAS includes location of these problems, so that IDEs could show them for the users review.

image

  Ideally I'd like to have something like this in IDEA.

**Inlining methods in hierarchy**
Nope, it's probably not safe and we want to show it as an error to give user the choice - continue or cancel.

PR #260, opened 2015-09-03
Don't apply potential edits during Rename refactoring.

scheglov:
Applying potential edits is unsafe.
We need a way to preview these changes and let users to check them on/off individually.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
+++ b/Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
@@ -16,6 +16,7 @@
 package com.jetbrains.lang.dart.assists;
 
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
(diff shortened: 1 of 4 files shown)


Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
```diff
@@ -48,13 +49,25 @@
 
 public class AssistUtils {
   public static void applyFileEdit(@NotNull final SourceFileEdit fileEdit) {
+    final Set<String> excludedIds = Sets.newHashSet();
```
alexander-doroshko: here and below: IMO `Collections.emptySet()` would better reflect that this map is not going to be updated inside and we really do not want to exclude anything.

alexander-doroshko:
Please add a test that potential usage is not renamed (as a separate PR as I'm pushing this).

PR #261, opened 2015-09-04
New DAS Client. Use 'superOnly'. Show only direct supers.

scheglov:


alexander-doroshko:
Awesome! Pushed.
This feature together with 'up' line markers looks production ready.
A couple more tests please: that we do not jump to sub; that we silently jump to super even when grandSuper exists (both for methods and classes). Unfortunately no idea how to test multitarget popup.

PR #264, opened 2015-09-08
Expand 'Goto super' tests - include grandsuper and sub classes.

scheglov:
Please note, that the test DartServerGotoSuperHandlerTest.testSuperClassMethod now fails.
The reason is that it is enable by default, but in plugin.xml DartServerGotoSuperHandler is not.

And there is a bug in non-server version in case of grand-super.


alexander-doroshko:
Pushed.
I've enabled server driven Go To Super and 'up' line markers by default.

PR #265, opened 2015-09-08
Add folding for multi-line strings

stevemessick:
@alexander-doroshko Fixes https://youtrack.jetbrains.com/issue/WEB-16703

I could not find a way to make the test framework define folding regions for strings that use triple single quotes. I tested those by hand and they work.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
+++ b/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
@@ -45,6 +45,7 @@ public class DartFoldingBuilder extends CustomFoldingBuilder implements DumbAwar
     foldClassBodies(descriptors, dartFile);                                            // 5. Class body
     foldFunctionBodies(descriptors, root);                                             // 6. Function body
     foldTypeArguments(descriptors, psiElements);                                       // 7. Type arguments
+    foldMultilineStrings(descriptors, root);                                           // 8. Multi-line strings
   }
 
   protected String getLanguagePlaceholderText(@NotNull final ASTNode node, @NotNull final TextRange range) {
@@ -63,6 +64,8 @@ public class DartFoldingBuilder extends CustomFoldingBuilder implements DumbAwar
     }
     if (psiElement instanceof DartFunctionBody) return "{...}";                      // 6.   Function body
     if (psiElement instanceof DartTypeArguments) return SMILEY;                      // 7.   Type arguments
+    if (psiElement instanceof DartStringLiteralExpression)
+      return multilineStringPlaceholder(node);                                       // 8.   Multi-line strings
 
     return "...";
   }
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
```diff
@@ -258,4 +261,32 @@ private static void foldTypeArguments(@NotNull final List<FoldingDescriptor> des
       }
     }
   }
+
+  private static void foldMultilineStrings(@NotNull final List<FoldingDescriptor> descriptors,
+                                           @NotNull final PsiElement root) {
+    for (PsiElement dartString : PsiTreeUtil.findChildrenOfType(root, DartStringLiteralExpression.class)) {
+      PsiElement child = dartString.getFirstChild();
+      if (child == null) continue;
+      IElementType type = ((ASTNode) child).getElementType();
```
alexander-doroshko: `child.getNode()` is more safe than casting

Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
```diff
@@ -258,4 +261,32 @@ private static void foldTypeArguments(@NotNull final List<FoldingDescriptor> des
...
+      if (type == DartTokenTypes.RAW_TRIPLE_QUOTED_STRING || (type == DartTokenTypes.OPEN_QUOTE && child.getText().length() == 3)) {
```
alexander-doroshko: `child.getTextLength()` may be faster than `child.getText().length()`

Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
```diff
@@ -258,4 +261,32 @@ private static void foldTypeArguments(@NotNull final List<FoldingDescriptor> des
       }
     }
   }
+
+  private static void foldMultilineStrings(@NotNull final List<FoldingDescriptor> descriptors,
+                                           @NotNull final PsiElement root) {
+    for (PsiElement dartString : PsiTreeUtil.findChildrenOfType(root, DartStringLiteralExpression.class)) {
```
alexander-doroshko: `PsiTreeUtil.findChildrenOfType` iterates the whole AST in the file. We also need full AST traversing to fold comments and type arguments. There's an optimization that collects both comments and type arguments in one shot. You can collect DartStringLiteralExpression in that loop too for better performance.

alexander-doroshko:
General consideration: may be we shouldn't fold the string if it is actually single line as we do not fold standard single line strings? Or at least do not fold strings shorter than 4 symbols as replacement text `...` doesn't make it shorter.

alexander-doroshko:
I'll push this pull request after AST iteration optimization.

stevemessick:
@alexander-doroshko wrote "may be we shouldn't fold the string if it is actually single line"

I believe single-line regions are filtered out by the framework. I was going to add code to do it but the folding affordance did not show up in the UI for multi-line strings that start and end on the same line. No point duplicating the work.

Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
stevemessick: I see what you mean, now that you point it out. I'm still adjusting to the way IDEA avoids using the visitor pattern. Can't quite get used to all the different alternatives.

PR #267, opened 2015-09-09
Fold multi-line strings

stevemessick:
@alexander-doroshko This addresses review comments from the previous one.


alexander-doroshko:
Thank you, pushed!

If you find it easier you can add a commit to existing pull request instead of closing it and opening a separate one.

> I believe single-line regions are filtered out by the framework.

IntelliJ IDEA doesn't show gutter markers for single-line folding regions (just because one line doesn't have enough space for opening and closing icon), but in fact these regions are there and you are still able to fold/unfold them using shortcuts like Cmd +, Cmd -, Cmd Shift +, Cmd Shift -.

PR #268, opened 2015-09-10
Fold multi-line strings

stevemessick:
@alexander-doroshko Inspired by your optimization question for string splitting, I found and fixed a bug that occurred in complex interpolations. The new test shows the problem. Without this code it inserts single quotes around the additional whitespace, which is syntactically wrong. This case is not handled by dartfmt but I checked with Bob to make sure using continuation indent does not violate any unwritten rules.


stevemessick:
Well, the title is wrong. This isn't folding, it is splitting. And it isn't about multi-line strings, it is for any string that spans multiple lines.

A couple interesting lines from the spec:

> The grammar ensures that a single line string cannot span more than one line of
source code, unless it includes an interpolated expression that spans multiple lines.
>...
> The reader will note that the expression inside the interpolation could itself
include strings, which could again be interpolated recursively.

PR #272, opened 2015-09-25
Fix for disable/enable OK button, specify ModalityState.

scheglov:


alexander-doroshko:
I pushed more suitable modality state and also added expiration check: see [commit](https://github.com/JetBrains/intellij-plugins/commit/9082e4e3660a86dcbe1698afff92b40c66b0434c#diff-35900d987289a83af101d18a9f7c3501)

PR #273, opened 2015-09-28
Goto implementations markers using DAS.

scheglov:
Also includes the new DAS client for the new IMPLEMENTED notification.


alexander-doroshko:
Tried with today's Dart SDK nightly. `List<ImplementedClass>` is empty if there's more than one space between `class` keyword and class name:
```
class  A{}
class B extends A{}
```

alexander-doroshko:
Something really strange with spaces. Playing only with spaces in this file:
```
class A {}
class B {
  x()=>1;
}
class D extends B{
  x()=>1;
}
class C = A with B implements D;
```
I was getting completely different notifications (digit after `=` shows corresponding list size):
```
foo.dart: classes=3, members=1
foo.dart: classes=0, members=0
foo.dart: classes=3, members=0
foo.dart: classes=1, members=1
```

alexander-doroshko:
Clicking on line marker for Object.toString() leads to 
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at com.jetbrains.lang.dart.ide.marker.DartServerImplementationsMarkerProvider$4.navigate(DartServerImplementationsMarkerProvider.java:169)
	at com.intellij.codeInsight.daemon.NavigateAction.actionPerformed(NavigateAction.java:56)
```

One ms is not enough for `search_getTypeHierarchy()` in this case and in any case check list size before get().

alexander-doroshko:
I suggest to skip showing line marker for Object class as everyone knows that all classes are derived from it. Line markers for its methods are ok.

Dart/src/com/jetbrains/lang/dart/ide/marker/DartServerImplementationsMarkerProvider.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+    // not found
+    return null;
+  }
+
+  @NotNull
+  private static LineMarkerInfo createMarkerClass(@NotNull final DartComponentName name) {
+    final VirtualFile file = name.getContainingFile().getVirtualFile();
+    final int nameOffset = name.getTextOffset();
+    return new LineMarkerInfo<PsiElement>(name, name.getTextRange(), AllIcons.Gutter.ImplementedMethod, Pass.UPDATE_OVERRIDEN_MARKERS,
```
alexander-doroshko: `Pass.UPDATE_OVERRIDEN_MARKERS` is used for markers created in `collectSlowLineMarkers()`. For markers from `getLineMarkerInfo()` use `Pass.UPDATE_OVERRIDEN_MARKERS`. Otherwise you'll see gutter width blinking and duplicate markers.
This is hard to guess, Platform API is bad here.

alexander-doroshko:
What do you think would be the best icon / tooltip for class as we do not know whether it is extended, implemented, mixed in or any combination of these? May be (O) and "Has subclasses"? Or "Has descendants"? Currently it is (I) and "Has implementations".
The same question about members.

alexander-doroshko:
Something to think about (I do not know solution yet). When clicking on line marker (both up and down), and also when calling DAS-based Go To Super (and the same about Go To Implementation when we switch it to DAS) we call `search_getTypeHierarchy()` and calculate `PsiElement` for each returned element. This calculation requires all affected files parsing. As a result we may get performance issues when, for example, searching for methods overriding `Object.toString()`. I'm not sure if it is easy to do, but would be great to find a way of showing the list of navigation targets without parsing them.

alexander-doroshko:
Please add tests.

Dart/src/com/jetbrains/lang/dart/ide/marker/DartServerImplementationsMarkerProvider.java
scheglov: Hm... you said `Pass.UPDATE_OVERRIDEN_MARKERS` twice.
I'm not sure what to use instead :-)

alexander-doroshko: Sorry, `Pass.UPDATE_ALL`, like in `DartServerOverrideMarkerProvider` :)

scheglov:
1. There is a bug in incremental resolver in Analyzer, which causes problems in the `implemented` notification.

2. Maybe don't show any markers for `Object`? Computing them in DAS is not cheap either. We end up requesting every class in the universe and asking whether it implements `toString` or `hashCode`, etc.

3. I don't have strong opinion about icons and text.

4. The easy thing with requesting `PsiElement`s is that we can use the same presentation everywhere. OTOH, we could probably create good enough presentation using DAS `Element`. Now sure about documentation comments.

alexander-doroshko:
> There is a bug in incremental resolver in Analyzer, which causes problems in the implemented notification.

Is any other functionality affected by this bug? Until fixed we have to stay with own implementation.

> Maybe don't show any markers for Object? Computing them in DAS is not cheap either. We end up requesting every class in the universe and asking whether it implements toString or hashCode, etc.

I guess this can be optimized by iterating only those classes that have methods with corresponding names. Server could create such index during initial parsing. Until optimized, not showing markers is ok for me.

scheglov:
Indexing and search are affected.
Fixed in https://github.com/dart-lang/sdk/commit/1d28f99524f5861762007db419094cbd5d78f993

scheglov:
PTAL
I made fixes in DAS, checks for empty items and added tests.

alexander-doroshko:
notifications from Dart SDK 1.12.1 stop working after applying this PR :(

BTW since we are updating analysis-server.jar so often, are you ok to keep sources instead of jar in thirdPartySrc, next to vmServiceDrivers?

scheglov:
Yes, it is actually easier for me to update sources than to build jars.

jwren:
Agreed!  That would make it a lot easier to updated and then to debug, etc.

scheglov:
Hm...
Works for me with this PR and SDK 1.12.1.
Navigation, semantic highlighting, even implemented elements - all works.
[The change](https://github.com/dart-lang/eclipse3/commit/5c25da238e68b758c131293fdf1e5042b9a2f3cb) in the client was pretty innocent.

image

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1013,6 +1028,7 @@ private void analysis_setSubscriptions() {
 
       final Map<String, List<String>> subscriptions = new THashMap<String, List<String>>();
       subscriptions.put(AnalysisService.HIGHLIGHTS, myVisibleFiles);
+      subscriptions.put(AnalysisService.IMPLEMENTED, myVisibleFiles);
```
alexander-doroshko: Got it. It is not jar that breaks work with SDK 1.12. It is this line. After receiving unknown subscription key DAS from 1.12 stops sending any file-level notifications. At least on Windows. On your screenshot that might be cached results received before changing SDK, or just different behavior on Mac.
I'll surround this line with if.

scheglov: OK, I see it too. I was not able to reproduce because DAS based implemented markers were commented out in `plugin.xml`.

The following statement from the DAS spec causes the problem: `It is an error if any of the keys in the map are not valid services. If there is an error, then the existing subscriptions will remain unchanged.`

@bwilkerson 
This seems not very client friendly and backward (or forward?) compatible.
Maybe we could subscribe clients for notifications DAS understands, and send back a list of services which DAS did not understand.

alexander-doroshko:
merged manually

PR #275, opened 2015-10-02
TableView based Problems view.

scheglov:
What do you think?
Is it a viable approach?
Can we start iterating from this?


alexander-doroshko:
Merged manually.
I like table-based problems view and started to work on improvements. Merging table update requests already done. Action panel added to the tool window. More actions/sorters/filters/etc will come.

jwren:
LGTM.  I have been playing with it.  I just created a small pull request on it.  Just to confirm, you are owning this Alex?

alexander-doroshko:
> you are owning this Alex?

I'm working on further improvements, is that what you are asking? Help is always welcome.

jwren:
Awesome.  Of course.  It just also doesn't make sense to have us doing the same work.

PR #277, opened 2015-10-06
Add vm_service_lib-2.0-dev.3.jar and weberknecht-0.1.5.jar as libraries into Dart-community.iml

jwren:


alexander-doroshko:
Already there :)

PR #278, opened 2015-10-06
Schedule sending 'analysis.updateContent' after any document change.

scheglov:


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -225,9 +230,43 @@ public void run(@NotNull ProgressIndicator indicator) {
     @Override
     public void beforeDocumentChange(DocumentEvent e) {
       myServerData.onDocumentChanged(e);
+      synchronized (myLock) {
+        final Document document = e.getDocument();
+        myUpdatedDocuments.add(document);
```
alexander-doroshko: Not everyone in the world edits only Dart files.
Even if Dart  - file may be outside of project root.
Even if in project, it may be not of interest if we do not keep any regions for it.
There are also other ifs.
Checking file presence in DartServerData maps seems to solve all of them.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -225,9 +230,43 @@ public void run(@NotNull ProgressIndicator indicator) {
...
+        scheduleUpdateFilesContent();
+      }
     }
   };
 
+  private void scheduleUpdateFilesContent() {
```
alexander-doroshko: We seem to already have a mechanism to update file highlighting if it is visible in the editor, after any change, be it external or just typing. Annotators will be started by highlighting daemon after some delay. I'll try to plug into DartServerErrorsAnnotator.collectInformation() instead. Does it sound good?

alexander-doroshko:
I'm playing with a bit different fix, please do not update this PR.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Hm...
You're right, we need more checks.
Checkin in `DartServerData` sounds good.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Short term, yes.
Long term, I'd like to stop `getErrors()` and use just `analysis.errors` notifications.
This implementation is more general purpose IMHO - sending updates as we get them.

alexander-doroshko: updateFilesContent() is also called by DartColorAnnotator that will survive switching to notifications, I'll plug there.

scheglov: Maybe a little theoretical problem, but...
Will it work if some background process will update the document of an invisible editor?
So, we would not need to highlight it, but the change would cause creating new errors, changes in index, etc.

scheglov: Ah... And in this cause using maps of `DartServerData` will not work - we would not have subscriptions for invisible files :-(

scheglov: Thinking a little more about it - in general DAS is not limited to analysis just Dart and HTML.

Right now - yes.

I don't know yet all the details, but we might have SASS files, defining names referenced in Angular HTML templates. So, ideally we need to notify DAS about any file in any analysis root managed by DAS.

alexander-doroshko: DartServerData contains data for all editors, visible and not visible. It clears file data on editor tab close.
Subscriptions are made for visible files only.
If we get doc change for not visible editor - yes, we will not get new data, neither before onDocumentUpdate(), nor after, so cached ranges will be shifted and won't break (but may be incorrect until file opened and new data arrives).

Looks like the problem that we are trying to solve is actual for visible files only. Will `if(visible)` be a correct `if` in beforeDocumentChange()?

scheglov: Yes, checking in `DartServerData` will solve the problem this PR was going to address.

But in general it would be nice to notify DAS about any (unsaved) change in any file in a managed root.

alexander-doroshko: That's the question to updateFilesContent() that iterates all unsaved docs and decides which ones to send.
The problem we are solving now is not to break ranges in files that we do already highlight using DAS. So this looks like a long term solution: once document is updated and it is visible at the moment make sure that on next IDE highlighting daemon pass it will be sent to the server (in order to get ranges, even if file is saved).

scheglov: OK, we need to solve the problem with breaking regions.
So, it's fine with me.

We will solve other problems later :-)

alexander-doroshko:
@scheglov My fix will appear [here](https://github.com/JetBrains/intellij-plugins/commits/master/Dart) in 10 minutes. Can you please review it and my other today's fixes? I want to cherry-pick them to the release branch before today's nightly build starts, but I need somebody's review for that.

PR #280, opened 2015-10-14
Do not try to split before string

stevemessick:
@alexander-doroshko


Dart/testSrc/com/jetbrains/lang/dart/typing/DartTypingTest.java
```diff
@@ -607,4 +607,11 @@ public void testEnterAfterEqualsProperty() {
                  "o.x =\n" +
                  "    <caret>");
   }
+
+  public void testEnterBeforeStringParam() {
```
alexander-doroshko: This test passes without any other code changes. I guess it's because `return` in test data is not inside a function.

alexander-doroshko: @stevemessick ok, as this is an important fix, I fixed the test myself and pushed. Will cherry-pick to the release branch.

PR #281, opened 2015-10-17
Made the Flexmojos config generation work with maven versions greater… than 3.1.0

chrisdutz:
By turning on the classpath scan, the plexus container now initializes correctly.

### DIFF
--- a/flex/tools/flexmojos-flex-configs-generator/generator-server/src/com/intellij/flex/maven/GeneratorServer.java
+++ b/flex/tools/flexmojos-flex-configs-generator/generator-server/src/com/intellij/flex/maven/GeneratorServer.java
@@ -15,9 +15,7 @@ import org.apache.maven.settings.building.DefaultSettingsBuildingRequest;
 import org.apache.maven.settings.building.SettingsBuilder;
 import org.apache.maven.settings.building.SettingsBuildingException;
 import org.apache.maven.settings.building.SettingsBuildingRequest;
-import org.codehaus.plexus.DefaultContainerConfiguration;
-import org.codehaus.plexus.DefaultPlexusContainer;
-import org.codehaus.plexus.PlexusContainerException;
+import org.codehaus.plexus.*;
 import org.codehaus.plexus.classworlds.ClassWorld;
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
 import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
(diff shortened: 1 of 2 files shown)


develar:
Thanks for investigation and fix. It seems, I missed corresponding change in Maven CLI — https://git-wip-us.apache.org/repos/asf?p=maven.git;a=commit;h=6e901c57979a8c3ab1901d1921702182140e610b

Could you please explain why PlexusConstants.SCANNING_ON is used instead of PlexusConstants.SCANNING_INDEX (as in the MavenCli.java)?

alexander-doroshko:
@chrisdutz Have you submitted a contributor license agreement? If not, please download the PDF from http://www.jetbrains.org/display/IJOS/Contributor+Agreement and then print it out, sign (mention you github account name there), scan and email to contribute@jetbrains.com

chrisdutz:
Well I did have similar problems when updating Flexmojos to Maven 3.1.1 and in this case I debugged the Container Initialization Code and I could simply see that he didn't find any components because he didn't even start looking. So I had a look at which switch I need to flip and came across that constant .. tried I and was happy that it worked ... so let's call it a well educated guess ... and that's also the reason why I can't explain it to you ;-)

I hadn't signed the ICLA yet ... will do that right away.

alexander-doroshko:
@develar @chrisdutz So finally, do you think `PlexusConstants.SCANNING_INDEX` would be better?

chrisdutz:
I couldn't find documentation on any of the two, so I can't say ... but Index definitely sounds better ... eventually it's faster. If it works, I'm fine with that :-)

alexander-doroshko:
@chrisdutz Can you please check if it works?

chrisdutz:
Ok ... ICLA is signed and sent back to you guys.

chrisdutz:
But it seems that checking the change ist not that simple ... seems that in the JS plugin things have changed that make it necessary to update the Flex plugin ... currently it doesn't compile any more :-(

chrisdutz:
Ok ... I just double checked and got things back to working ... so it seems it also works with SCANNING_INDEX, but don't ask me if the one is better than the other.

alexander-doroshko:
Thank you for testing! IntelliJ IDEA 15 / 15.0.1 release branch is closed, so the fix will go to 15.0.2 update release.

chrisdutz:
Cool ... in my setup I made it build a flexmojos-flex-configs-generator-server-311.jar and to use that if the version is above or equal to 3.1.1 ... so I haven't tested the patch working for 3.1.0 (That's the strange maven version that probably none uses as it does seem to have a lot of problems) ... just wanted to point that out ... but I think it would be safe to not introduce a third generator-server jar.

alexander-doroshko:
@chrisdutz What maven repository I have to use to get net.flexmojos.oss.flexmojos-maven-plugin:7.1.0-SNAPSHOT? The latest on [maven central](http://search.maven.org/#search%7Cga%7C1%7Cflexmojos-maven-plugin) is 7.0.1.

BTW I followed instructions from [this page](https://cwiki.apache.org/confluence/display/FLEX/Quick+Start+Guide%3A+Building+Apache+Flex+applications+using+Maven). Is this the latest doc? Instructions about mavenizer seem to be outdated there, but I managed to guess a few tricks to proceed. I ended up with successful 'mvn compile' with flexmojos-maven-plugin 7.0.1, but failed to find 7.1.0.

alexander-doroshko:
Ok, while browsing the cloned https://git-wip-us.apache.org/repos/asf/flex-utilities.git repository I found the URL: http://repository.apache.org/snapshots/. I'm very curious how do other users find information about flexmojos.

PR #282, opened 2015-10-20
Add occurrence ranges to update name during the Introduce Variable refactoring.

scheglov:
This was reported couple times by internal users.
I also find it quite annoying :-)


alexander-doroshko:
Thanks for the fix! This is the last non-DAS refactoring. We should try to keep it inline when will be switching to DAS.

PR #286, opened 2015-10-21
New DAS client. Pass raw VM/DAS arguments.

scheglov:
We need to allow our internal clients to pass VM arguments, for example to start DAS with Observatory.
It seems that instead of having a bunch of options it's easier to give users just one really big hammer.

Ideally it would be nice to have the Registry value `dart.server.vm.arguments` just as we have `dart.server.additional.arguments` now.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -5,7 +5,6 @@ import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.dart.server.*;
 import com.google.dart.server.generated.AnalysisServer;
 import com.google.dart.server.internal.remote.DebugPrintStream;
-import com.google.dart.server.internal.remote.FileReadMode;
 import com.google.dart.server.internal.remote.RemoteAnalysisServerImpl;
 import com.google.dart.server.internal.remote.StdioServerSocket;
 import com.google.dart.server.utilities.logging.Logging;
@@ -48,8 +47,10 @@ import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiFileSystemItem;
 import com.intellij.psi.PsiManager;
 import com.intellij.psi.search.FilenameIndex;
-import com.intellij.util.*;
+import com.intellij.util.Alarm;
 import com.intellij.util.Consumer;
+import com.intellij.util.PathUtil;
+import com.intellij.util.Processor;
 import com.intellij.util.concurrency.Semaphore;
 import com.intellij.util.containers.ContainerUtil;
 import com.intellij.util.net.NetUtils;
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1171,18 +1172,28 @@ public void println(String str) {
...
 
-      String argsRaw;
+      String vmArgsRaw;
       try {
-        argsRaw = Registry.stringValue("dart.server.additional.arguments");
+        vmArgsRaw = System.getProperty("dart.server.vm.arguments", "");
       }
       catch (MissingResourceException e) {
-        argsRaw = "";
+        vmArgsRaw = "";
```
alexander-doroshko: I think `System.getProperty` doesn't throw MissingResourceException

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1171,18 +1172,28 @@ public void println(String str) {
...
-      argsRaw += " --useAnalysisHighlight2";
+
+      String serverArgsRaw;
+      try {
+        serverArgsRaw = Registry.stringValue("dart.server.additional.arguments");
+      }
+      catch (MissingResourceException e) {
+        serverArgsRaw = "";
+      }
+      serverArgsRaw += "--port=" + port;
```
alexander-doroshko: Leading space seems to be needed

alexander-doroshko:
I've added `dart.server.observe` registry option few weeks ago, but none of us added its usage :)
If it covers 99% of use cases we could use it instead of the newly added "dart.server.vm.arguments" system property.

What is more convenient: registry option or system property?

Super-stable release branch is already created. Is it ok if it will appear in WebStorm update release only? If they use IntelliJ IDEA it may appear much earlier in the plugin.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Thank you, fixed.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: This is kind of for a new Registry value, once we have it :-)

scheglov:
Maybe both? I'm not sure.
There are too many options - even just "--observe" has two - port and binding address.
Plus options (if a user will need them - how deeply to do sampling, what to do with code blocks, etc).
It's not practical to support all the options in a special way.
But it might be easier for user to do a simple thing if it is in UI, at least in some way.
OTOH, if you're already in Registry, it might be as easy to type --observe as it is to turn it ON.

scheglov:
Update release sounds good.
I don't want to break something :-)

alexander-doroshko:
AFAIK vm args are more often called vm **options**. Are you ok with "dart.server.vm.options" registry property?

scheglov:
It's all messy in VM and DAS.

For DAS we actually also specify options:
image

And Dart VM has `flags`, plus you also give your Dart program `options`:
```
scheglov@scheglov-macbookpro3:~/Source/Dart/sdk.git/sdk (master)$ dart --help
Usage: dart [<vm-flags>] <dart-script-file> [<dart-options>]
```

That said, the name `dart.server.vm.options` sounds good.

alexander-doroshko:
Mesy indeed.
In Dart run configuration UI we use VM Optoins and Program Arguments (I used these terms to be similar to Java run configuration).
Ok, added `dart.server.vm.options`, removed `dart.server.observe` that we didn't use anyway.

PR #288, opened 2015-10-22
Indentarg

stevemessick:
@alexander-doroshko I was surprised that PsiErrorElement didn't have something like isError() defined for it. But it looks like using instanceof is the way other projects detect it.


Dart/src/com/jetbrains/lang/dart/ide/formatter/DartIndentProcessor.java
```diff
@@ -133,6 +143,9 @@ public Indent getChildIndent(final ASTNode node) {
       if (EXPRESSIONS.contains(elementType) && elementType != FUNCTION_EXPRESSION) {
         return Indent.getContinuationIndent();
       }
+      if (elementType == NAMED_ARGUMENT) {
+        return Indent.getContinuationIndent();
```
alexander-doroshko: is there a test for it?

PR #291, opened 2015-10-30
Launcher for package:test

stevemessick:
@alexander-doroshko This replaces the previous prototype of this work.

At this point, it is sort of usable. The output parser populates the results view and the tests can be browsed from it. It isn't complete yet.

I added a test for the parser that converts output from the test runner to IJ UI test events.


alexander-doroshko:
Looking at it at the moment...

PR #292, opened 2015-11-02
DAS can highlight any referenced Dart file.

scheglov:
If an external file is referenced from a project, then it is also analyzed and DAS can provide its highlighting, navigation, etc.


alexander-doroshko:
Is it a real use case or just a project misconfiguration?

scheglov:
It's a real use case.
Internally at Google developers open only one or two folders with a small subset of files - only files that they want to edit.
But they use many external libraries, which are not physically in these folders.
And it's still nice to open a library you are using and read its code with semantic highlighting.

I guess it might be also the case when we stop using packages/ folders.

scheglov:
Actually, scratch it.
You're are right, with `dart.projects.without.pubspec` option DPI uses `analysis.analyedFiles` notification and semantic highlighting works just fine. I need to remember to turn this option on during my tests.

PR #293, opened 2015-11-03
Use DAS for type hierarchies.

scheglov:


alexander-doroshko:
Looks like server issue: for some classes DAS says that the class itself is its own child, even twice:
image

alexander-doroshko:
Can you please review [follow up commit](https://github.com/JetBrains/intellij-plugins/commit/5693e9740598b84cbf68557d24b28a585578ce20): I added tests, check for empty items list, no request to server for Object, correct class name offset. If all is fine we can include it in update release.

PR #294, opened 2015-11-04
Reveal 1.6.2 tvOS support

tonyarnold:
In Reveal 1.6.2, we ship a properly linked set of tvOS libraries that are distinct from the iOS libraries. I've updated the AppCode Reveal plugin to use the correct library when running against tvOS targets. 

I've only done preliminary testing on these changes (and as before, my Java-fü is not strong), so you may wish to review these changes a little more closely before deciding whether to merge. 

Thanks guys!


tonyarnold:
Superseded by #391.

PR #298, opened 2015-11-09
Integration with the DAS execution domain to map the file paths to uris and vice-versa.

jwren:


alexander-doroshko:
Breakpoints in SDK files do not work for me because `execution_mapUri(1, "C:\dart\dart-1.14\lib\core\print.dart", null)` returns `file:///C:/dart/dart-1.14-fake/lib/core/print.dart` instead of `dart:core/print.dart`. Server bug?

alexander-doroshko:
I pushed a follow up commit to handle dart:xxx URIs without server, I hope it works for internal users as well.

PR #303, opened 2015-11-23
Ignore TODOs from the Dart Analysis Server in the new Dart Analysis panel, like the previous panel

jwren:
…


alexander-doroshko:
Pushed, thanks. Next time please reformat added code. You can call reformat action manually or enable corresponding check box in the Commit dialog.

jwren:
Sorry!, Yes of course.

alexander-doroshko:
I noticed that I've accidentally removed similar check in DartServerErrorsAnnotator few months ago and currently all TODOs are highlighted as warnings. I think we have thousands of users and nobody reported... :). Pushing the fix.

jwren:
Maybe they have no TODOs   :)   Thanks for pushing a fix.

PR #304, opened 2015-11-23
Redirecting and super constructor invocation resolution.

scheglov:
Requires support form DAS landed as https://github.com/dart-lang/sdk/commit/670db507079be11890937b79b39b3635e11f7ca3 today.


alexander-doroshko:
Great!
I also managed to make `testSuperConstructorInvocation()` work against Dart SDK 1.13 where `super.foo` is treated as a single range. But it seems to be impossible to make `this` references work in SDK 1.13: no ranges for it at all.

alexander-doroshko:
Would be great to add a test to `DartServerFindUsagesTest` to make sure that `this` and `super` references are found as usages.

PR #306, opened 2015-12-01
Support for filtering problems in projects that use BUILD files.

scheglov:
@jwren Could you please check if it seems reasonable to you?

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/util/DartBuildFileUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/util/DartBuildFileUtil.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2000-2015 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


jwren:
I didn't patch it to try, but it LGTM.

Dart/src/com/jetbrains/lang/dart/util/DartBuildFileUtil.java
```diff
@@ -0,0 +1,58 @@
...
+  @Nullable
+  public static VirtualFile findPackageRootBuildFile(@NotNull final Project project, @NotNull final VirtualFile contextFile) {
+    final ProjectFileIndex fileIndex = ProjectRootManager.getInstance(project).getFileIndex();
+    VirtualFile parent = contextFile.isDirectory() ? contextFile : contextFile.getParent();
+
+    while (parent != null && fileIndex.isInContent(parent)) {
+      final VirtualFile file = parent.findChild(BUILD_FILE_NAME);
+      if (file != null && !file.isDirectory()) {
+        final VirtualFile parent2 = parent.getParent();
+        if (parent2 != null && parent2.findChild(LIB_DIR_NAME) == null) {
```
alexander-doroshko: Do you need to check BUILD file presence in parent2 also?

scheglov: We could, and it would express our expectations explicitly.
But it is not absolutely necessary.
If you or @jwren fell strong about this, I could add this check though.

PR #307, opened 2015-12-02
DAS based Extract Local Variable refactoring.

scheglov:
~~Plus new client.~~
Minus new client :-)


scheglov:
The PR was updated.
1. Tests added.
2. Added check for Dart SDK `>= 1.14` to activate in-place name editing, or the dialog otherwise.

PTAL

alexander-doroshko:
Awesome!

For SDK < 1.14 I returned using IDE's own `DartIntroduceVariableHandler`. `DartServerExtractLocalVariableHandler` doesn't seem to work for older SDKs and even if fixed, it's too risky to hang server (issue you fixed recently).

PR #311, opened 2015-12-08
Parse operators as method declarations.

scheglov:
This makes operators display in the Structure view.
Also, because it is a `DartComponent` now it can be used as a resolution / navigation target in the future.

In `Dart Analysis Server` we already parse operators as methods.


alexander-doroshko:
Cool!
I could easily make 'Override operator' action working (`Ctrl+O`):
image

alexander-doroshko:
Also we've got line markers for operators for free! I've added corresponding tests.
image

scheglov:
Nice :-)

jwren:
Very nice, good job you two.

alexander-doroshko:
Fixed Go To Super / Go To Implementations for operators as well + tests. It was easy now!

PR #313, opened 2015-12-09
Index all top-level component symbols.

scheglov:
This allows using Navigate | Symbol... for top-level functions, top-level variables and typedefs.
I discussed this locally in the office and we thought that classes also should be in the list of symbols.
It seems that in Java classes are included too.


alexander-doroshko:
Sorry, forgot to tell you before: when you touch indexes or significantly change parser make sure to increase `DartIndexUtil#INDEX_VERSION`. Otherwise index is not updated for files that you do not edit. For example go to `print` symbol won't start working.
Pushed with minor follow-up. Now we can even navigate to operator by name! :-)

PR #314, opened 2015-12-10
Generate documentation using Analysis Server, when possible.

scheglov:
This may be a solution for [WEB-15936](https://youtrack.jetbrains.com/issue/WEB-15936).

When DAS.getHover() is used, we can show the exact type information about the given element  both static and propagated. Note that even the static type is not always the same as the declared type because of the type promotion process or "strong mode".

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -89,6 +89,7 @@ public class DartAnalysisServerService {
   private static final long EDIT_ORGANIZE_DIRECTIVES_TIMEOUT = TimeUnit.SECONDS.toMillis(3);
   private static final long EDIT_SORT_MEMBERS_TIMEOUT = TimeUnit.SECONDS.toMillis(3);
   private static final long GET_ERRORS_TIMEOUT = TimeUnit.SECONDS.toMillis(5);
+  private static final long GET_HOVER_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
   private static final long GET_NAVIGATION_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
   private static final long GET_ERRORS_LONGER_TIMEOUT = TimeUnit.SECONDS.toMillis(60);
   private static final long GET_ASSISTS_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
@@ -655,6 +656,35 @@ public class DartAnalysisServerService {
         }
       }
     });
+  }
+
+  @NotNull
+  public List<HoverInformation> analysis_getHover(@NotNull final String _filePath, final int offset) {
+    final String filePath = FileUtil.toSystemDependentName(_filePath);
+    final List<HoverInformation> result = Lists.newArrayList();
+
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return HoverInformation.EMPTY_LIST;
+    }
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    server.analysis_getHover(filePath, offset, new GetHoverConsumer() {
+      @Override
+      public void computedHovers(HoverInformation[] hovers) {
+        Collections.addAll(result, hovers);
+        latch.countDown();
+      }
+
+      @Override
+      public void onError(RequestError error) {
+        logError("analysis_getHover()", filePath, error);
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, GET_HOVER_TIMEOUT);
+    return result;
   }
 
   @Nullable
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -32,27 +32,93 @@ public class DartDocUtil {
       return null;
     }
     final DartComponent namedComponent = (DartComponent)(element instanceof DartComponent ? element : element.getParent());
-    final StringBuilder builder = new StringBuilder();
 
-    appendLibraryName(builder, element);
-
-    builder.append("<code>");
-
-    for (DartMetadata metadata : namedComponent.getMetadataList()) {
-      builder.append(StringUtil.escapeXml(metadata.getText())).append("<br/>");
+    final String signatureHtml;
+    {
+      final StringBuilder builder = new StringBuilder();
+      appendSignature(namedComponent, builder);
+      signatureHtml = builder.toString();
     }
 
-    appendSignature(namedComponent, builder);
+    final String containingLibraryName;
+    final PsiFile file = element.getContainingFile();
+    if (file != null) {
+      containingLibraryName = DartResolveUtil.getLibraryName(file);
+    } else {
+      containingLibraryName = null;
+    }
 
-    builder.append("</code>");
+    final String containingClassDescription;
+    final DartClass dartClass = PsiTreeUtil.getParentOfType(namedComponent, DartClass.class);
+    if (dartClass != null) {
+      final StringBuilder builder = new StringBuilder();
+      builder.append(dartClass.getName());
+      appendTypeParams(builder, dartClass.getTypeParameters());
+      containingClassDescription = builder.toString();
+    } else {
+      containingClassDescription = null;
+    }
 
     final String docText = getDocumentationText(namedComponent);
+    return generateDoc(signatureHtml, true, docText, containingLibraryName, containingClassDescription, null, null);
+  }
+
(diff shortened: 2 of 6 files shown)


alexander-doroshko:
Which one do you like more?
image
For me the right one looks better. I see that Foo is a class and no senseless signature and type.

Dart/testSrc/com/jetbrains/lang/dart/documentation/DartDocUtilTest.java
```diff
@@ -22,58 +22,58 @@ private void doTest(String expectedDoc, String fileContents) {
   }
 
   public void testAbstractClassSig() throws Exception {
-    doTest("<code><small><b>test.dart</b></small></code><br/><br/><code>abstract class <b>Foo</b><br/>extends Bar</code>",
+    doTest("<code><b>Containing library:</b> <code><small>test.dart</small></code><br/><br/><code>abstract class <b>Foo</b> extends Bar</code>",
```
alexander-doroshko: Unbalanced <code> tags in expected data is suspicious

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -78,6 +79,8 @@
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceArray;
```
alexander-doroshko: unused import

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -657,6 +661,35 @@ public void run() {
     });
   }
 
+  @NotNull
+  public List<HoverInformation> analysis_getHover(@NotNull final String _filePath, final int offset) {
+    final String filePath = FileUtil.toSystemDependentName(_filePath);
+    final AtomicReference<List<HoverInformation>> result = new AtomicReference<List<HoverInformation>>();
```
alexander-doroshko: You do not use 'atomic' features of this reference. That's correct, we do not need them, we are not afraid of a race here. So simple `Ref` that we use in all other methods here is enough.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -657,6 +661,35 @@ public void run() {
     });
   }
 
+  @NotNull
```
alexander-doroshko: It returns null if timed out

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -657,6 +661,35 @@ public void run() {
...
+
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return HoverInformation.EMPTY_LIST;
+    }
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    server.analysis_getHover(filePath, offset, new GetHoverConsumer() {
+      @Override
+      public void computedHovers(HoverInformation[] hovers) {
```
alexander-doroshko: In what cases this array is longer than 1? Multi resolve like one part file for several main files?

Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
```diff
@@ -268,7 +269,7 @@ private static void appendClassSignature(final StringBuilder builder, final Dart
     final List<DartType> mixins = dartClass.getMixinsList();
     final DartType superClass = dartClass.getSuperClass();
     if (superClass != null) {
-      builder.append("<br/>extends ").append(StringUtil.escapeXml(superClass.getText()));
+      builder.append(" extends ").append(StringUtil.escapeXml(superClass.getText()));
```
alexander-doroshko: Don't you like it on separate lines like done for Java? Implements list may be long.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Correct.
I don't know for sure if DAS actually ever returns more than one, but it could.

Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocumentationProvider.java
```diff
@@ -94,18 +98,89 @@ else if (componentType == DartComponentType.CLASS) {
...
+    if (elementDescription != null) {
+      builder.append("<b>Signature:</b> ");
+      builder.append(StringUtil.escapeXml(elementDescription));
+      builder.append("<br>");
+    }
+    if (containingLibraryName != null || containingClassDescription != null) {
+      builder.append("<br>");
+      if (containingLibraryName != null) {
+        builder.append("<b>Containing library:</b> ");
+        builder.append(StringUtil.escapeXml(containingLibraryName));
```
alexander-doroshko: lib name may be an empty string

Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocumentationProvider.java
```diff
@@ -94,18 +98,89 @@ else if (componentType == DartComponentType.CLASS) {
     return resultUrl.toString();
   }
 
-  @Override
-  public String generateDoc(PsiElement element, PsiElement originalElement) {
-    return DartDocUtil.generateDoc(element);
+  @Nullable
+  private static String generateDocServer(PsiElement element) {
```
alexander-doroshko: Idea: unify doc preview regardless of the source of information: server or IDE. Extract a method that will build resulting HTML based on prepared input data.

alexander-doroshko:
Need a test for server based docs. At least one for each kind of declaration.
Review completed.

alexander-doroshko:
WEB-15936 is about hover, it is a bit more tricky to fix than docs.

> Note that even the static type is not always the same as the declared type because of the type promotion process or "strong mode".

I'm afraid I'm out of context. Can you provide an example?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Hint: to avoid unused imports and other typos I always review the code before commit. I press Ctrl+D (Cmd+D) in Local Changes tab right before commit or right in the commit dialog and browse all changes by simply pressing F7. Editing and reformatting code works in the diff view.

scheglov:
This is an example of type promotion.

See `19.1.1 Type Promotion` and `17.5 If` in the Dart specification.

```
class A {}
class B extends A {
  bbb() {}
}

main(A p) {
  if (p is B) {
    p.bbb();
  }
}
```

  In this example the declared type of `p` is `A`.
  But the the static type is `B`, because we "promoted" the type from the `if` statement.
image

scheglov:
We can also apply `type propagation` to a variable - additional changes to the type, which are not mandated by the specification. The following code has a required warning, but it has all three types - declared, static and propagated.

```
class A {}
class B extends A {
  bbb() {}
}
class C extends B {
  ccc() {}
}

main(A p) {
  if (p is B) {
    p.bbb();
    if (p is! C) {
      return;
    }
    p.ccc();
  }
}
```

image

alexander-doroshko:
Oh, too many types: declared, static, promoted, propagated, inferred. Is promoted == static? Is propagated == inferred?

scheglov:
Yes, too many types... of types :-)
It depends on the contexts and even options.

Yes, when we do `type promotion`, we update the `static` type.

In the strong mode static types of some parameters and return types of methods are replaced. This is called `inferred` types internally, we AFAIK we expose them as `static` types.

Types propagation is for local variables and parameters. No, `propagated != inferred`.

scheglov:
The PR is updated.

1. I agree that signature for classes is not as useful as it could be. This [was fixed](https://github.com/dart-lang/sdk/commit/168556a3e44364ac02555b4b3aec3d5821fb1545) in DAS.

2. Removed unused imports.

3. Removed usage of `Ref` at all, just `List` now, always not `null`, but maybe empty.

4. Yes, `extends` and `implements` can be both long. But not always. And they also can be long even if we put them on separate lines. So, I'd like to keep in on a single line and let the wrapper to do its work as needed.

5. Yes, library names can be empty. I converted all the checks to `!StringUtil.isEmpty()`.

6. Implemented a common documentation generator using given parameters.

7. Ported with some changes tests.

PR #315, opened 2015-12-11
Enable running package:test suites

stevemessick:
@alexander-doroshko This enables the new feature that runs tests written with package:test.

There are a few things I still want to do. First is to see about enabling navigation from test results to source code. Other user-visible changes include simplifying the run configuration editor and trimming repetitive name elements from the test result tree. You asked that the run-parameters be rewritten. I haven't done that but hopefully the current code won't clobber serialized run configurations. There's also an unknown amount of work needed to enable debugging test code in the IJ debugger.


alexander-doroshko:
Pushed, thanks!
Additionally I disabled Debug button for this run configuration.
BTW since there's not more old unittest support, do you think it'd be better to name the new one just 'Dart Test' or 'Dart Tests'? Screenshot for context:
image

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunningState.java
```diff
@@ -197,10 +209,11 @@ private void setupParameters(@NotNull final Project project,
...
     }
-    else {
-      myObservatoryPort = PubServerManager.findOneMoreAvailablePort(myDebuggingPort);
-      commandLine.addParameter("--enable-vm-service:" + myObservatoryPort);
-    }
+    // TODO Enable this for debugging package:test code, but make sure it is for the test runner VM.
+    //else {
+    //  myObservatoryPort = PubServerManager.findOneMoreAvailablePort(myDebuggingPort);
+    //  commandLine.addParameter("--enable-vm-service:" + myObservatoryPort);
+    //}
```
alexander-doroshko: This disables observatory for standard Dart command line app as well. I've added `instanceof DartTestRunnerParameters` check instead.

PR #316, opened 2015-12-15
Implement navigation to test source for tests defined in the test file

stevemessick:
@alexander-doroshko You can double-click a test result, or use the context menu, to show the test source code in the editor. This only works for tests define in the file loaded by the test runner. According to package:test maintainers that's how tests should be written -- there is no need for something like all_tests.dart anymore. We will have to talk about the desired UX for IntelliJ.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
@@ -6,11 +6,14 @@ import com.intellij.execution.testframework.sm.ServiceMessageBuilder;
 import com.intellij.execution.testframework.sm.runner.OutputToGeneralTestEventsConverter;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.text.StringUtil;
+import com.jetbrains.lang.dart.ide.runner.util.DartTestLocationProvider;
 import jetbrains.buildServer.messages.serviceMessages.ServiceMessageVisitor;
 import org.jetbrains.annotations.NotNull;
 
 import java.text.ParseException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
```diff
@@ -151,6 +166,11 @@ private boolean handleTestStart(JsonObject obj) throws ParseException {
     return result;
   }
 
+  private static String encodeLocation(String url, Item item) {
```
alexander-doroshko: Method is unused. Can it be removed?

alexander-doroshko:
`DartTestLocationProviderTest` has to be updated.

alexander-doroshko:
Will it make sense to handle that theoretically can be thrown by `fromJson()` methods used in different places?

alexander-doroshko:
Review completed. Need to update broken test before push.

alexander-doroshko:
Awesome side effect of the fix is that now it is possible to run a single test by a right-click on the node in test result tree!

Related question: is it possible to implement 'Jump to source' for test group nodes as well?

Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
stevemessick: Yes. I forgot to delete it after inlining it.

stevemessick:
*Will it make sense to handle that theoretically can be thrown by fromJson() methods used in different places?*

You mean doing something other than what processEventText() does? I don't really think it is necessary. That one catches all event-parsing exceptions. Decoding the URL in DartTestLocationProvider should not throw any exceptions since it is decoding JSON produced by the event converter.

*Awesome side effect of the fix is that now it is possible to run a single test by a right-click on the node in test result tree!*

How do you get that? I don't see it.

*Related question: is it possible to implement 'Jump to source' for test group nodes as well?*

Good question. I don't know how that is implemented.

alexander-doroshko:
Don't you have it in context menu?
image

Mechanism is the following: 
- `DartTestLocationProvider` returns `PsiElement` that corresponds to the node in test result tree. Returned `PsiElement` is a `DartCallExpression` that corresponds to `test()` or `group()` method call.
- `DartTestRunConfigurationProducer` suggests run configuration based on context `PsiElement`. For  `test()` or `group()` calls it suggests to sun a single test/group correspondingly. It is the same as if you right click on a `test()` or `group()` call in editor.

stevemessick:
My context menu is different. The Run item says it will run the whole test
suite again, not an individual test, even though I have a single test
selected. I'm not sure why it is different from yours.

As to groups (since I'm not sure which you were asking about) I don't get
much of a context menu for them. It does not have the run or navigate
items. Again, I'm not sure why.

Both of these things need to be fixed IMO. But I think they can wait while
I figure out how to get test navigation working, since that may require
changes to the test runner.

PR #318, opened 2015-12-28
Minor changes to the test runner

stevemessick:
@alexander-doroshko This adds the group nav feature we discussed. I also changed the description of the unit test config. Feel free to ignore that.


alexander-doroshko:
Thanks!
Does it fix [WEB-19591](https://youtrack.jetbrains.com/issue/WEB-19591) also?
'Dart Test' name is in public plugin already. I noticed that most of other run configurations use title capitalization, and none has plural noun, so I did the same. Is it ok to leave with it?

stevemessick:
I think that bug is a different problem.

Please use the wording that best suites you.

PR #319, opened 2015-12-28
Minor update to test runner

stevemessick:
@alexander-doroshko Same as before, mostly, with the addition of the fix for https://youtrack.jetbrains.com/issue/WEB-19591. Also adjusted config title text.


alexander-doroshko:
@stevemessick Navigation to group still doesn't work for me
image

PR #321, opened 2016-01-07
Fix problem with skipped group

stevemessick:
@alexander-doroshko Minor bug fix. Not sure if you prefer handling exceptions in this case or precomputing string length and explicitly checking it, so I did what I thought would be best.


alexander-doroshko:
@stevemessick May be the best would be 
```
if (myName.startsWith(getParent().getName() + " ") {...}
```
?

PR #322, opened 2016-01-08
fix split editor with JavaFx preview interfering with language fragme…

vsch:
Needed fix to allow language fragment editing with split editor showing both markdown and JavaFX preview.


valich:
@vsch Thank you very much, this works indeed. However, this particular fix is not very good since it's more of a component “bug” rather than some needed behaviour.
I think the panel without `peer` should not return `size()` on `getMinimumSize()` so it'll be better to make a suitable class which could possibly be reused later.

I've come with that: https://github.com/JetBrains/intellij-plugins/commit/ff38dc1057215799c02d72331ef58f6426c7ec15.

PR #323, opened 2016-01-08
Enable WolfTheProbemSolver for the Dart Plugin, highlighting of the files view when there are errors and warnings

jwren:


jwren:
This change is dependent on the following PR: https://github.com/JetBrains/intellij-plugins/pull/323

PR #324, opened 2016-01-08
Dart Plugin fix, show method separators when the preference is set, WEB-19748

jwren:


Dart/src/com/jetbrains/lang/dart/ide/marker/DartMethodLineMarkerProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import com.intellij.psi.PsiWhiteSpace;
+import com.intellij.util.FunctionUtil;
+import com.jetbrains.lang.dart.psi.*;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.List;
+
+public class DartMethodLineMarkerProvider extends LineMarkerProviderDescriptor {
```
alexander-doroshko: It is enough to implement `LineMarkerProvider`. Extending `LineMarkerProviderDescriptor` makes one more check box to appear in Preferences | Editor | General | Gutter Icons, this is not what we need for method separators as there's already an option for it.

alexander-doroshko:
Some thoughts:
image

1. Do we want separators between global functions as well?
2. As I see in Java separator before the first method in class is drawn only if there is anything else before this method in this class (e.g. field declaration). Sounds reasonable do do the same (see `bar` on screenshot).
3. Is there a need for separators between inner functions (like `foo` and `foo1` on screenshot)? Java doesn't draw them inside anonymous classes.

Some tests would be great.
Review completed.

Dart/src/com/jetbrains/lang/dart/ide/marker/DartMethodLineMarkerProvider.java
jwren: Done.

jwren:
PR updated.

1. Yes, done.
2. Yes, we can implement the same logic, but it looks like a bug to me, I could be convinced, but think this looks better.  LMK.
3. This definitely looks like a bug in the Java marker provider to me.  For a large inner class, why shouldn't the methods be marked if this preference is flipped?  Let me know if you want us to be consistent with Java...

Tests: I'd like them as well.  The tests for gutters, don't help us here, and I could not find the support in CodeInsightTestFixture to find and assert that the line marker is drawn.  I could not find tests against the JavaLineMarkerProvider either.  If you point me to the support (or example of such a test), I will followup with a PR.

alexander-doroshko:
1. Thanks! But separator before the first function is definitely needless :)
image

2. I see that it is done intentionally in Java, doesn't look like a bug. This is a 'separator', not a preceding decorator. I think you agree that there's no sense to draw it after the last method in class. So I think in class without any code before the first method is it not needed as well.
image

3. Again I see that it is done intentionally for Java. I didn't talk to the authors, but it looks correct to me as well. 
-- I feel that separators are separating structural entities. We do not show inner functions in Structure View (and I guess nobody wants it), probably that's why I do not expect them to be separated visually.
-- Other thought is that separating inner functions introduce kind of 'nested' separators,  i.e. separators of different levels, and to me they do not look like things that can be nested / multi-level.

As for tests I found `ExpectedHighlightingData#extractExpectedLineMarkerSet` method and finally came to this code. Please update test data to cover all corner cases.

```
public class DartMethodSeparatorsTest extends DaemonAnalyzerTestCase {
  @Override
  protected boolean doTestLineMarkers() {
    return true;
  }

  public void testMethodSeparators() throws Exception {
    DaemonCodeAnalyzerSettings.getInstance().SHOW_METHOD_SEPARATORS = true;
    try {
      configureByText(DartFileType.INSTANCE, "class A{\n" +
                                             "  <lineMarker>bar(){}</lineMarker>\n" +
                                             "}");
      doDoTest(false, false);
    }
    finally {
      DaemonCodeAnalyzerSettings.getInstance().SHOW_METHOD_SEPARATORS = false;
    }
  }
}
```
I'm pushing this, so separate PR is welcome.

jwren:
https://github.com/JetBrains/intellij-plugins/pull/326

PR #325, opened 2016-01-11
Dart Live Templates: sout and soutm

jwren:


Dart/resources/liveTemplates/dart_miscellaneous.xml
```diff
@@ -45,5 +45,21 @@
...
   </template>
+
+  <template name="sout" value="print('$END$');" description="print to standard out"
+            toReformat="true" toShortenFQNames="true">
+    <context>
+      <option name="DART" value="true"/>
+    </context>
+  </template>
+
+  <template name="soutm" value="print('$CLASS_NAME$.$METHOD_NAME$');&#10;$END$" description="prints the current class and method name to standard out"
```
alexander-doroshko: Line break at the end is unexpected to me. As I see no other similar templates have it. Was it intentional?

alexander-doroshko:
"soutm" doesn't work correctly for global functions and functions that are inside other methods. Not a major issue, pushing as is for now.

jwren:
Yes, soutm should be fixed for the global (not in a class) case.  I see how Java does it... for now at least we can add some tests: https://github.com/JetBrains/intellij-plugins/pull/327

PR #326, opened 2016-01-12
Address comments and add tests, in Dart method line separators feature

jwren:
... introduced in PR #324, https://github.com/JetBrains/intellij-plugins/pull/324


alexander-doroshko:
Cool, thanks!
I've also added a case with comments before method to test data.

PR #327, opened 2016-01-13
Some tests for the Dart Live Templates: sout and soutm

jwren:


alexander-doroshko:
I've also added `<caret>` marker to `*.after` test data files.

PR #328, opened 2016-01-13
Dart Live template changes

jwren:
- Add and modify the Dart template contexts to limit where additional Dart expressions can be added in the source
- Addition of the DartClassNameMethodNameMacro to fix soutm behavior when not contained in a class.
- Removal of the dart_surround.xml templates
- Moved dart_iterations.xml into dart_miscellaneous.xml


Dart/src/com/jetbrains/lang/dart/ide/template/DartTemplateContextType.java
```diff
@@ -1,27 +1,62 @@
...
+
   @Nullable
   @Override
   public SyntaxHighlighter createHighlighter() {
     return new DartSyntaxHighlighter();
   }
+
+  public static class Generic extends DartTemplateContextType {
+    public Generic() {
+      super("DART_CODE", DartBundle.message("dart.language.id"), EverywhereContextType.class);
```
alexander-doroshko: User defined live templates configured to be applicable in Dart will stop working if we change 'Dart' serialization to 'DART_CODE'. Are there downsides of keeping 'DART'?

See also `dart_surround.xml` file: templates defined there do not work with this pull request.
BTW `dart_surround.xml` itself looks suspicious as it introduces a separate templates group. Not sure if adding more than one group per language is justified. Also I never used templates from this group, may be this is a dead horse. We can look for similar things in other languages for reference.

jwren: I didn't think about user defined templates, I will change it back, I was just trying to be consistent with the Java implementation.

dart_surround.xml- agreed, it will be refactored into the existing group, just trying to keep the PRs small enough to be able to keep momentum.

PR #329, opened 2016-01-15
Dart generator work, constructor generators

jwren:
- Update the constructor generator for Dart;
- Addition of a new generator: named constructor.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateNamedConstructorFix.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateNamedConstructorFix.java
@@ -1,7 +1,23 @@
+/*
+ * Copyright 2000-2016 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.jetbrains.lang.dart.ide.generation;
 
 import com.intellij.codeInsight.template.Template;
 import com.intellij.codeInsight.template.TemplateManager;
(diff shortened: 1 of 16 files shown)


Dart/src/com/jetbrains/lang/dart/ide/generation/CreateConstructorFix.java
```diff
@@ -44,10 +44,9 @@ protected Template buildFunctionsText(TemplateManager templateManager, Set<DartC
         template.addTextSegment(",");
       }
     }
-    template.addTextSegment(")");
-    template.addTextSegment("{\n");
+    template.addTextSegment(");");
```
alexander-doroshko: I guess you are right, just wanted to make sure, is constructor without body more popular use case?

Dart/src/com/jetbrains/lang/dart/ide/generation/CreateNamedConstructorFix.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  protected Template buildFunctionsText(TemplateManager templateManager, Set<DartComponent> elementsToProcess) {
+    final Template template = templateManager.createTemplate(getClass().getName(), DART_TEMPLATE_GROUP);
+    template.setToReformat(true);
+
+    //noinspection ConstantConditions
+    template.addTextSegment(myDartClass.getName());
+    template.addTextSegment(".");
+    template.addVariable(new MacroCallNode(new CompleteSmartMacro()), true);
```
alexander-doroshko: CompleteSmartMacro does nothing but showing 'No suggestions' because we do not have smart completion in Dart. Standard completion won't help either as it doesn't suggest anything in this case. I suggest to add some default name, e.g. 'name' or 'constructor' and simply stop on it.
```
template.addVariable(new TextExpression("constructor"), true);
```
It becomes preselected, so user can start typing desired name.
image

Dart/src/com/jetbrains/lang/dart/ide/generation/CreateConstructorFix.java
jwren: With no body- yes.  Since the generator doesn't put any statements into the body, this is the natural choice.

Also, it is less keystrokes for users to add the braces, than to remove them.

Dart/src/com/jetbrains/lang/dart/ide/generation/CreateNamedConstructorFix.java
jwren: Will do.

jwren:
PR updated

PR #330, opened 2016-01-15
Remove throws Throwable from step template

Dayde:
Hi,
I've not been using Cucumber and its IntelliJ plugin for too long but every time I generate steps from my feature I find myself deleting on each method "throws Throwable". Is there a reason I should let it there ? Otherwise would you accept a real PR (not like this one ^^) to remove it or to enable the user to edit the templates (like junit test templates).

### DIFF
--- a/cucumber-java/src/org/jetbrains/plugins/cucumber/java/steps/JavaSnippet.java
+++ b/cucumber-java/src/org/jetbrains/plugins/cucumber/java/steps/JavaSnippet.java
@@ -25,7 +25,7 @@ class JavaSnippet implements Snippet {
   }
 
   public String template() {
-    return "@{0}(\"{1}\")\npublic void {2}({3}) throws Throwable \'{\'\n    // {4}\n{5}    throw new PendingException();\n\'}\'\n";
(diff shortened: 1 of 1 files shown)


BedfordWest:
I'm curious if there's been any thought or decision on this. I recently started removing this from our step definitions, as well, and actually find this behavior potentially dangerous. Using `throws Throwable` is bad practice, and potentially encourages or at least enables testers to pass off responsibility for exceptions they should actually be handling.

avokin:
I see contradiction with Cucumber Java library. If you run feature with undefined step definition Cucumber Java would inform that step is undefined and propose template for the missing step. And this template contains "throws Throwable". I prefer Cucumber plugin to correspond to Cucumber Java library by default. 

Probably it's better to have possibility to edit step definition template in Settings.

Dayde:
I think I was not clear in my comment, but I would like to delete the "throws Throwable" part of the template not the template itself which is very nice to have for generating the test class basecode.

avokin:
I suppose the best solution is to provide possibility to edit template for step definition.

https://youtrack.jetbrains.com/issue/IDEA-188946. I'll implement it in IDEA 2018.2

PR #337, opened 2016-01-26
New generator for the Dart plugin: toString()

jwren:


alexander-doroshko:
If `toString()` is already overridden in the current class this action should ask for the permission and overwrite it, or simply be disabled and invisible.

Dart/src/com/jetbrains/lang/dart/ide/generation/DartGenerateToStringHandler.java
```diff
@@ -0,0 +1,65 @@
...
+      public boolean value(DartComponent component) {
+        return DartComponentType.typeOf(component) == DartComponentType.FIELD;
+      }
+    }));
+  }
+
+  @Override
+  public boolean isValidFor(@NotNull Editor editor, @NotNull PsiFile file) {
+    return super.isValidFor(editor, file);
+  }
```
alexander-doroshko: Why method overridden with no additional logic?
If you decide to disable action if `toString()` is already declared in the current class you can implement this check here.

Dart/src/com/jetbrains/lang/dart/ide/generation/DartGenerateToStringHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  protected BaseCreateMethodsFix createFix(DartClass dartClass) {
+    return new CreateToStringFix(dartClass);
+  }
+
+  @Override
+  protected void collectCandidates(DartClass aClass, List<DartComponent> candidates) {
+    super.collectCandidates(aClass, candidates);
+  }
```
alexander-doroshko: `super.collectCandidates()` is doing a lot of heavy calculations that you simply filter out later in the another `collectCandidates()` below. As we do not need anything from superclasses I think we can completely override this method here without calling for `super.collectCandidates(aClass, candidates)`. Or if you do not like current API of the `BaseDartGenerateHandler` class feel free to change it.

alexander-doroshko:
OK in dialog should not be blocked if no members are selected to be used in `toString()`.

alexander-doroshko:
Review completed. Do you want me to push it as is and address review comments in a separate PR?

jwren:
Thanks for the review.  Don't commit it, I will iterate, and rebase for you to take a look tomorrow.  Thanks Alex

jwren:
Currently working on addressing the comments.  isValidFor is not called, the "enable/disable" is taken care of via the Action, not Handler...

Dart/src/com/jetbrains/lang/dart/ide/generation/DartGenerateToStringHandler.java
jwren: If we remove the contents of the first collectCandidates, the next collectCandidates is never called.

jwren:
PR fixed, and squashed into a single commit.

The action is only enabled if toString() is not already in the class, but not in a superclass.

The user can select no items and have the generator work appropriately (also, if there is no members to list in the dialog, it does show at all.)

Dart/src/com/jetbrains/lang/dart/ide/generation/BaseDartGenerateAction.java
```diff
@@ -39,9 +39,26 @@ public void update(final AnActionEvent e) {
...
-    e.getPresentation().setVisible(inClass);
+    e.getPresentation().setEnabled(enable);
+    e.getPresentation().setVisible(enable);
   }
+
+  protected boolean doEnable(@Nullable final DartClass dartClass) {
+    return dartClass != null;
+  }
+
+  protected static boolean doesClassContainMember(@NotNull final DartClass dartClass, @NotNull final String memberName) {
```
alexander-doroshko: This method name seems to be the opposite to what it returns.

Dart/src/com/jetbrains/lang/dart/ide/generation/BaseDartGenerateAction.java
```diff
@@ -39,9 +39,26 @@ public void update(final AnActionEvent e) {
...
+    if (memberName.isEmpty()) {
+      return false;
+    }
+    final DartComponent dartComponent = dartClass.findMemberByName(memberName);
+    if (dartComponent == null) {
+      return true;
+    }
+    final DartClass containingClass = PsiTreeUtil.getParentOfType(dartComponent, DartClass.class);
+    return containingClass != dartClass;
```
alexander-doroshko: As we do not need to iterate supeclasses here we can use `DartResolveUtil#getNamedSubComponents` that looks into current class only.
Or even more optimal: iterate `PsiTreeUtil.getChildrenOfType(DartResolveUtil.getBody(dartClass), DartMethodDeclaration.class)` checking for `toString()` among them.

Dart/src/com/jetbrains/lang/dart/ide/generation/DartGenerateToStringHandler.java
alexander-doroshko: > If we remove the contents of the first collectCandidates, the next collectCandidates is never called.

Yes, with current API of the `BaseDartGenerateHandler` class to optimize calculations we can completely override `collectCandidates(DartClass aClass, List<DartComponent> candidates)` which, for our case should only look for fields in the current class.
Or rewrite base class to be more convenient for all subclasses.

PR #338, opened 2016-01-29
Enable test nav for multiple-file runs

stevemessick:
@alexander-doroshko This allows double-click nav from test name to test source in the test results view. It also works for groups. The only real problem is for legacy projects that have a single file used to drive all tests, as was needed in package:unittest. The suites are not generated for those runs. Using the new standard "run all tests in a directory" form, the navigation works as expected, and that should make a lot of users happy.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
@@ -28,14 +28,16 @@ import java.util.regex.Pattern;
 public class DartTestEventsConverter extends OutputToGeneralTestEventsConverter {
 
   private static final String TYPE_START = "start";
-  private static final String TYPE_TEST_START = "testStart";
+  private static final String TYPE_SUITE = "suite";
   private static final String TYPE_ERROR = "error";
   private static final String TYPE_GROUP = "group";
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
Works great, pushing.

While testing I noticed that our events converter is not tolerant to unknown events, and therefore not future-compatible. I'm getting an error:
```
Caused by: com.google.gson.JsonSyntaxException: Unexpected type: allSuites (check for package:test update)
	at com.jetbrains.lang.dart.ide.runner.test.DartTestEventsConverter.process(DartTestEventsConverter.java:143)
	at com.jetbrains.lang.dart.ide.runner.test.DartTestEventsConverter.processEventText(DartTestEventsConverter.java:113)
	at com.jetbrains.lang.dart.ide.runner.test.DartTestEventsConverter.processServiceMessages(DartTestEventsConverter.java:97)
	at com.intellij.execution.testframework.sm.runner.OutputToGeneralTestEventsConverter.processConsistentText(OutputToGeneralTestEventsConverter.java:95)
...
```

[AllSuitesEvent](https://codereview.chromium.org/1641353002/) was introduced recently. It allows us to make accurate progress bar and also to count finished tests in each group and send `ServiceMessageBuilder.testSuiteFinished()` to the framework to stop showing 'in progress' group icon.

PR #342, opened 2016-02-01
In DART_STATEMENT template types, exclude comments

jwren:


Dart/src/com/jetbrains/lang/dart/ide/template/DartTemplateContextType.java
```diff
@@ -56,7 +57,8 @@ public Statement() {
 
     @Override
     protected boolean isInContext(@NotNull PsiElement element) {
-      return PsiTreeUtil.getParentOfType(element, DartBlock.class) != null;
+      return PsiTreeUtil.getNonStrictParentOfType(element, DartBlock.class) != null &&
+             PsiTreeUtil.getNonStrictParentOfType(element, PsiComment.class) == null;
```
alexander-doroshko: I pushed a bit more optimal:
```
return PsiTreeUtil.getNonStrictParentOfType(element, DartBlock.class, PsiComment.class) instanceof DartBlock;
```

PR #343, opened 2016-02-01
New generator for the Dart plugin: equals and hashCode

jwren:


Dart/resources/META-INF/plugin.xml
```diff
@@ -255,6 +255,10 @@
             text="toString()">
       <add-to-group anchor="after" relative-to-action="Generate.GetSetAccessor.Dart" group-id="GenerateGroup"/>
     </action>
+    <action id="Generate.EqualsAndHashcode.Dart" class="com.jetbrains.lang.dart.ide.generation.DartGenerateEqualsAndHashcodeAction"
+            text="Equals and HashCode">
```
alexander-doroshko: may be name it `==() and hashCode`?

Dart/testData/generate/EqualsAndHashCode2.txt
```diff
@@ -0,0 +1,15 @@
...
+    if (identical(this, other)) {
+      return true;
+    }
+    return other is EqualsAndHashCode2;
+  }
+
+  @override
+  int get hashCode {
+    return 0;
+  }<caret>
```
alexander-doroshko: `==()` and `hashCode` for classes without fields look useless. May be hide the action in classes without fields and do not allow OK in dialog without selection?

alexander-doroshko:
Corner case for all our code generating actions: they generate red code if `dart:core` is already imported with import prefix and/or if import directive contains `show`/`hide` that affects core classes/functions that we reference in generated code.

jwren:
https://github.com/JetBrains/intellij-plugins/pull/344

As for the corner case that all of our generators have: I'm not concerned.  If someone has that use case, they'll quickly discover that they need to modify the generators output.  For the use of the `show`, I suppose we could generate the correct code (low priority).  For the use of `hide`, I don't think users would want us to remove the hide....

PR #344, opened 2016-02-02
Minor changes to the ==() and hashCode Dart generator

jwren:


alexander-doroshko:
Action is still suggested in a class without fields. If you agree that it makes no sense we can either hide the action or on action activation show message in editor like we show for Generate Getter in a class without fields.

jwren:
I also added logic to make sure that the fields that exist are non-static. https://github.com/JetBrains/intellij-plugins/pull/345

PR #345, opened 2016-02-03
Dart '==() and hashCode' generator change

jwren:
Disable the Dart ==() and hashCode generator if there is not a non-static field in the containing class


alexander-doroshko:
Good catch with statics!
We seem to have the same problem when generating constructors, named constructors and toString().

jwren:
Still some work to cleanup the overrides/implements (in process of testing, I found this, https://github.com/dart-lang/sdk/issues/25686), but this is the next incremental PR: https://github.com/JetBrains/intellij-plugins/pull/346

PR #348, opened 2016-02-08
Fix for Dart Generator issue WEB-20204

jwren:
Fix for Dart Generator issue WEB-20204, generation of static getters and setters when the field is static

### DIFF
--- a/Dart/testData/generate/GetterSetter4.txt
+++ b/Dart/testData/generate/GetterSetter4.txt
@@ -1,0 +1,14 @@
+class GetterSetter4 {
+  static var _foo;
+  static int _bar;
+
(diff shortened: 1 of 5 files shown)


Dart/testData/generate/GetterSetter4.txt
```diff
@@ -0,0 +1,14 @@
...
+  static get foo => _foo;
+
+  static set foo(value) => _foo = value;
+
+  static int get bar => _bar;
+
+  static set bar(int value) => _bar = value;
+
+
+}
```
alexander-doroshko: I fixed extra new lines that are added after generated code

PR #349, opened 2016-02-08
WEB-16813- Dart generators inserted new members after method comments

jwren:
WEB-16813- Dart generators inserted new members after method comments


alexander-doroshko:
@jwren https://twitter.com/_dynaxis/status/697002966008475648 :)

PR #351, opened 2016-02-11
Start of work to get the Dart pkg/test support to work with bazel

jwren:


Dart/src/com/jetbrains/lang/dart/ide/runner/test/ui/DartTestConfigurationEditorForm.java
```diff
@@ -99,6 +102,7 @@ protected void textChanged(final DocumentEvent e) {
 
     myFileField.getTextField().getDocument().addDocumentListener(documentListener);
     myDirField.getTextField().getDocument().addDocumentListener(dirListener);
+    myTargetNameField.getDocument().addDocumentListener(dirListener);
```
alexander-doroshko: Is there a need in listener for 'Target Name' field? Nothing in UI seem to depend on its content.

jwren: Removed.

jwren:
PR updated

PR #352, opened 2016-02-12
Dart plugin generators- Continued work on fixing WEB-19915

jwren:
- Dart override and implements generators now include private methods from the same library,
- Fields are now included in the implements generator.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/generation/BaseDartGenerateHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/generation/BaseDartGenerateHandler.java
@@ -42,7 +42,7 @@ public abstract class BaseDartGenerateHandler implements LanguageCodeInsightActi
     invoke(project, editor, file, editor.getCaretModel().getOffset());
   }
 
-  public void invoke(Project project, Editor editor, PsiFile file, int offset) {
+  public void invoke(@NotNull final Project project, @NotNull final Editor editor, @NotNull final PsiFile file, final int offset) {
     if (!FileModificationService.getInstance().prepareFileForWrite(file)) return;
     final DartClass dartClass = PsiTreeUtil.getParentOfType(file.findElementAt(offset), DartClassDefinition.class);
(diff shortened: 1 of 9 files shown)


jwren:
https://github.com/JetBrains/intellij-plugins/pull/352

alexander-doroshko:
Thanks!
I've found some cases that could be discussed or improved. If it's not 5 minutes question feel free to file a YouTrack issue.

This is the generated code for our 'Implement methods' action (B extends A).
```
class C implements B {
  @override
  final String publicFieldInB = null;
  @override
  int _privateFieldInB;

  @override
  publicMethodInB() {
  }

  @override
  _privateMethodInB() {
  }
  @override
  var publicFieldInA;

  @override
  publicMethodInA() {
  }
}
```

1. Is there a need for `@override` annotation for fields?
2. Members order could be better. At least all fields before all methods. Additionally we may group private/public and alpha-sort. Need to check Dart Style conventions.
3. It's interesting to compare code generated by the Analysis Server 
image 

with our generator. One of the things that differ is that for final fields DAS generated getter, while our implementation generates final fields. To me our implementation looks better, but @scheglov may think different.
image

vs

image

jwren:
1) I went and read the `@override` dartdoc, and I believe that it is expected.

2) Yes, issue created here: https://youtrack.jetbrains.com/issue/WEB-20372

3) The Analysis Server behavior is a bit different for implements as you point out, this is because the semantic meaning of what the interface looks like, not the syntax.  If a class you are implementing has a getter and a setter, the implementation is just a field, etc.  I don't think we want this behavior with the generators (although the order should be better).

scheglov:
1. Yes, I think DAS also should be changed to generate `@override` when adds fields.

3. Semantically a final field and a getter are equivalent, so there is no much difference. Getters are IMHO marginally better because the spec describes resolution in terms of getters and setters. But I can be convinced otherwise.  We could also add a Quick Assist(s) to convert between a single expression getter and a final fields.

PR #354, opened 2016-02-18
Updates for Dart coverage

jingbian:
Better coverage message for directories.

Wrap all the loading of coverage data inside `runProcessWithProgressSynchronously`.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/coverage/DartCoverageAnnotator.java
+++ b/Dart/src/com/jetbrains/lang/dart/coverage/DartCoverageAnnotator.java
@@ -15,10 +15,15 @@
  */
 package com.jetbrains.lang.dart.coverage;
 
+import com.intellij.coverage.CoverageSuitesBundle;
 import com.intellij.coverage.SimpleCoverageAnnotator;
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
Thanks, pushed.
BTW I found much more efficient way to count files without extra file system traversing. Coverage engine already iterates all files, so to count untouched files it is enough to add the following to `DartCoverageAnnotator` (pushed myself):
```
  @Override
  protected FileCoverageInfo fillInfoForUncoveredFile(@NotNull final File file) {
    return new FileCoverageInfo();
  }
```

jingbian:
I know that part. But that will create problems.
If a file is recognized in the coverage data but has no lines possible to cover. Then it won't be counted as it won't go through `fillInfoForUncoveredFile`.

I chose the current because I think the other one will create confusion.

What's your thought?

jingbian:
I did try that. But then those files became 100% covered. I thought I should not tweak the `getLinesCoverageInformationString` part at that time.
But since you changed the `getLinesCoverageInformationString`, it will work.

Thanks for optimization!

PR #355, opened 2016-02-29
Some minor URI translator refactoring in preparation for flutter support

stevemessick:
@alexander-doroshko We need slightly different URI translations for flutter. Also, the workaround for a problem with DAS URIs causes problems with flutter. I just added a check to skip it. Also some protections were changed to enable subclassing.


Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -111,6 +111,8 @@ public void stackFrameChanged() {
       LOG.assertTrue(myExecutionResult == null && myDASExecutionContextId == null, myDASExecutionContextId + myExecutionResult);
     }
     else {
+      // This causes a runtime exception if test debugger starts before DAS is started. That can happen today if
+      // a project is opened with a YAML file in the editor and a pre-defined test is started via the Debug button.
```
alexander-doroshko: I've fixed it.

PR #356, opened 2016-03-01
Made check for "addEventListener" method precise.

fwienber:
`ActionScriptTypeChecker` complains if a method called `addEventListener` uses a callback that does not have a single parameter of type `flash.events.Event`.
However, the check used to simply compare the method name, not whether it actually refers to `flash.events.IEventDispatcher#addEventListener()`. Thus, if a method in another framework is named `addEventLister`, you get false negatives.

This fix checks whether the object on which `addEventListener()` is invoked is of a type that implements Flash's `IEventDispatcher` interface.

I would have loved to add unit tests for this fix, but like others mentioned on the Apache Flex mailing list, even without changing anything, Flex unit tests are not green.


alexander-doroshko:
Thank you for your contribution!
Have you submitted a contributor license agreement? If not, please [download the PDF] (http://www.jetbrains.org/display/IJOS/Contributor+Agreement) and then print it out, sign (mention you github account name there), scan and email to contribute@jetbrains.com

PR #357, opened 2016-03-01
Also allow catalog.xml files in JARs, not only in SWCs.

fwienber:
Jangaroo uses the same catalog.xml format as Flash/Flex SWC, but packages
its artifacts as JARs.
It should not do harm to look for catalog.xml entries in the root directory of
JAR archives that are in the dependency list of a Flex module, too.


fwienber:
This change is a one-liner and I really don't know how to test this.
If looking for catalog.xml in JARs, too, has any undesired effect, it will most likely not be discovered by a unit test.
Should the PR go on master instead of 143, too?

alexander-doroshko:
Initially all contributions may be accepted to master branch only. After that we may decide and cherry-pick it to other branch ourselves.

fwienber:
Okay, new PR upcoming.

PR #359, opened 2016-03-02
Initial definition of plugin for flutter

stevemessick:
@alexander-doroshko This code is still buggy. Half the time when I launch a flutter app I get errors, but it works some times, so it looks like a timing problem in start up. It has to run flutter twice. The first time launches the app. The second starts the log monitor, and keeps the debug session alive.

I don't really know what the Flutter.iml file is for, so it probably has a bunch of stuff it shouldn't have. Let me know and I'll clean it up, or you can delete things if you want.


alexander-doroshko:
*.iml files contain module configuration: paths, dependencies, etc. Visual representation of it is in Project Structure | Modules.

PR #361, opened 2016-03-03
Allow "native" constructor in MXML.

fwienber:
For Ext JS compatibility, Jangaroo needs to specify a constructor signature in MXML classes. Standard MXML always generates an implicit no-arg constructor,  thus IDEA marks any constructor in MXML script code as red.

This change allows to specify a constructor signature in MXML using the `native` keyword (the constructor implementation is generated from the declarative MXML part).
The `native` keyword does not appear in standard MXML, so no Flex developer would try to use it anyway, and thus it does not affect normal Flex development.

Test `FlexHighlightingTest#testConstructorNotAllowedInMxml()` is still green.
Added new test `FlexHighlightingTest#testNativeConstructorAllowedInMxml()` to assure that `native` constructors can now indeed be used in MXML.

### DIFF
--- a/flex/flex-tests/testSrc/com/intellij/lang/javascript/FlexHighlightingTest.java
+++ b/flex/flex-tests/testSrc/com/intellij/lang/javascript/FlexHighlightingTest.java
@@ -734,6 +734,11 @@ public class FlexHighlightingTest extends ActionScriptDaemonAnalyzerTestCase {
     doTestFor(true, getTestName(false) + ".mxml", getTestName(false) + "_2.as");
   }
 
+  @JSTestOptions({JSTestOption.WithFlexSdk})
+  public void testNativeConstructorAllowedInMxml() throws Exception {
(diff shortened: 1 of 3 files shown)


flex/flex-tests/testSrc/com/intellij/lang/javascript/FlexHighlightingTest.java
```diff
@@ -734,6 +734,11 @@ public void testConstructorNotAllowedInMxml() throws Exception {
     doTestFor(true, getTestName(false) + ".mxml", getTestName(false) + "_2.as");
   }
 
+  @JSTestOptions({JSTestOption.WithFlexSdk})
+  public void testNativeConstructorAllowedInMxml() throws Exception {
+    doTestFor(true, getTestName(false) + ".mxml", getTestName(false) + "_2.as");
```
alexander-doroshko: Second file is not needed for this test. I'll remove the file itself and reference to it myself.

PR #363, opened 2016-03-09
Emit <base /> tag when rendering markdown

sdeken:
The Markdown JavaFx previewer does not render images when the images are relative to the directory containing the .md file.  This patch should cause the renderer to emit a properly formed <base /> tag, allowing relative images to be referenced properly.

I'm unfortunately not able to build the plugins to test this.

### DIFF
--- a/markdown/src/org/intellij/plugins/markdown/ui/preview/MarkdownPreviewFileEditor.java
+++ b/markdown/src/org/intellij/plugins/markdown/ui/preview/MarkdownPreviewFileEditor.java
@@ -30,6 +30,8 @@ import org.jetbrains.annotations.Nullable;
 import javax.swing.*;
 import java.awt.*;
 import java.beans.PropertyChangeListener;
+import java.net.URI;
+import java.net.URISyntaxException;
 
 public class MarkdownPreviewFileEditor extends UserDataHolderBase implements FileEditor {
   private final static long PARSING_CALL_TIMEOUT_MS = 50L;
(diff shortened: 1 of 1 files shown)


valich:
Hi @sdeken, this is a needed feature, however I think it should be implemented in another way at least because when referencing of the links from text editor is implemented, the same "base" should be appended here.

Moreover, github itself converts `[foo](blob/master/...)` to `<a href="/<user>/<repo>/blob/master/blob/master/...">foo</a>` so it prepends the needed parts before in links themselves.

I think one needs either to replace links in plugin after html generation or do this right in the process of html generation (https://github.com/valich/intellij-markdown; https://github.com/valich/intellij-markdown/tree/master/src/org/intellij/markdown/html) which could be reused elsewhere.

Please tell me if you need any assistance on plugin building.

sdeken:
Hi @valich, thanks for responding - I wasn't entirely sure I submitted to the right place, or whether I needed to file a bug elsewhere to notify people of the pull request.

The beauty of using the base tag is that it applies to both the `img` and the `a` tag, so no further changes would be needed in order to render the page correctly.  It doesn't get much less intrusive than adding a single line of code!

Nevertheless, I updated it so that instead of emitting a `base` tag, the URLs are modified during the generation phase.  I submitted a matching pull request over on the other repository:

https://github.com/valich/intellij-markdown/pull/10

Hopefully yours is the official repository for intellij-markdown?

I was able to build and test the intellij-markdown project, but I'm still having no luck building the intellij plugin.  I'm not super worried about it for now, though.

valich:
I've updated the plugin to utilise changes in the parser code, the issue is fixed.
Will be available with the nearest plugin update.

PR #365, opened 2016-03-11
Add Rerun Failed Tests button to test results viewer

stevemessick:
@alexander-doroshko Testing is manual. I suggest checking a test suite that has no failures to ensure the button is visible but not active. Then a suite that has one failure, run it, fix the problem, rerun to get same view as first test. Then a suite with multiple failures. For the final one I re-ran it a couple times, fixed a problem, re-ran it and observed the fixed test disappeared from the view. Then re-introduce the problem, re-run failed tests, and see nothing happened (since the modified test was not in the list of failed tests after being fixed).

I needed to do a little refactoring in the framework, but it was mostly conceptual rather than structural.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestRerunnerAction.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestRerunnerAction.java
@@ -1,0 +1,53 @@
+package com.jetbrains.lang.dart.ide.runner.test;
+
+import com.intellij.execution.Executor;
+import com.intellij.execution.configurations.RunConfigurationBase;
+import com.intellij.execution.configurations.RunProfileState;
+import com.intellij.execution.runners.ExecutionEnvironment;
+import com.intellij.execution.testframework.TestConsoleProperties;
+import com.intellij.execution.testframework.actions.AbstractRerunFailedTestsAction;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.ui.ComponentContainer;
+import com.jetbrains.lang.dart.ide.runner.base.DartRunConfiguration;
+import com.jetbrains.lang.dart.ide.runner.server.DartCommandLineRunnerParameters;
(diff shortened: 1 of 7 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestRerunnerAction.java
```diff
@@ -0,0 +1,53 @@
...
+import com.intellij.execution.testframework.actions.AbstractRerunFailedTestsAction;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.ui.ComponentContainer;
+import com.jetbrains.lang.dart.ide.runner.base.DartRunConfiguration;
+import com.jetbrains.lang.dart.ide.runner.server.DartCommandLineRunnerParameters;
+import org.jetbrains.annotations.NotNull;
+
+class DartTestRerunnerAction extends AbstractRerunFailedTestsAction {
+
+  public DartTestRerunnerAction(@NotNull ComponentContainer componentContainer, @NotNull TestConsoleProperties consoleProperties) {
```
alexander-doroshko: Can we remove unused `consoleProperties` parameter?

Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestRunningState.java
```diff
@@ -32,6 +37,8 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import static com.intellij.openapi.command.impl.CommandLog.LOG;
```
alexander-doroshko: I guess this import was suggested by too smart IntelliJ's code completion :). In fact loggers from other classes rarely should be reused. I'll change it to current class own logger.

alexander-doroshko:
Works great, pushing!

One thought: I think it will be better to include full test name (including containing group names) in the regexp to avoid false matches with same named tests in different groups.

PR #367, opened 2016-03-15
Remove unused parameter

stevemessick:
@alexander-doroshko As requested!


alexander-doroshko:
I removed import statement that became unused as well :)

PR #376, opened 2016-04-07
Terminate tests when Rerun is clicked

stevemessick:
@alexander-doroshko This is in response to:
https://youtrack.jetbrains.com/issue/WEB-21067

I believe junit stops tests that are running when the Rerun button is clicked, so that's what this code does.

Manual testing is required, unfortunately.


alexander-doroshko:
@stevemessick It seems to be a valid use case to have several independent test runs in parallel. 'Rerun Failed Tests' in one of them shouldn't stop other processes. And I think it shouldn't even stop the current process. I see that in other testing frameworks (including JUnit) this action reruns tests that are already failed at the moment and doesn't stop the current one, so I did the same for Dart.

PR #378, opened 2016-04-27
Implement Call Hierarchy for Dart

stevemessick:
@alexander-doroshko This implements the Call Hierarchy menu item on the Navigate menu for Dart code.

The "Multi" tests cause harmless exceptions to be printed to the console but they pass. If that's an issue I'll either delete them or change them to eliminate the recursive method calls. But I like having a test of recursive calls.


Dart/src/com/jetbrains/lang/dart/util/DartResolveUtil.java
```diff
@@ -1030,8 +1030,13 @@ public static DartClass suggestType(PsiElement element) {
 
   @Nullable
   public static DartComponent findReferenceAndComponentTarget(@Nullable PsiElement element) {
-    DartReference reference = PsiTreeUtil.getParentOfType(element, DartReference.class);
-    PsiElement target = reference == null ? null : reference.resolve();
+    PsiElement target;
+    if (element instanceof DartReferenceExpression) {
+      target = ((DartReferenceExpression)element).resolve();
```
alexander-doroshko: `DartReferenceExpression` extends `DartReference`, I think instead of a separate if-branch it's better to change `getParentOfType` to `getNonStrictParentOfType`

Dart/resources/META-INF/plugin.xml
```diff
@@ -289,5 +290,22 @@
       <separator/>
     </group>
 
+    <action id="DartCallHierarchy.BaseOnThisFunction" text="Base on This Function"
+            class="com.jetbrains.lang.dart.ide.hierarchy.call.DartCallHierarchyBrowser$BaseOnThisFunctionAction"
+            use-shortcut-of="MethodHierarchy"/>
```
alexander-doroshko: MethodHerarchy -> CallHierarchy

Dart/resources/META-INF/plugin.xml
```diff
@@ -289,5 +290,22 @@
       <separator/>
     </group>
 
+    <action id="DartCallHierarchy.BaseOnThisFunction" text="Base on This Function"
+            class="com.jetbrains.lang.dart.ide.hierarchy.call.DartCallHierarchyBrowser$BaseOnThisFunctionAction"
```
alexander-doroshko: Since `DartCallHierarchyBrowser$BaseOnThisFunctionAction` is empty I think removing it and pointing here to `com.intellij.ide.hierarchy.CallHierarchyBrowserBase$BaseOnThisMethodAction` should work

alexander-doroshko:
I'll take a closer look tomorrow

Dart/src/com/jetbrains/lang/dart/util/DartResolveUtil.java
stevemessick: I made the change but I'm not sure it is still necessary.

Dart/resources/META-INF/plugin.xml
stevemessick: OK. I also changed the menu text from "Function" to "Component" to hopefully be more indicative that it works for getter, setters, and constructors, too.

Dart/resources/META-INF/plugin.xml
stevemessick: Done

Dart/testData/analysisServer/callHierarchy/MultiCallees/MultiCallees_verification.xml
```diff
@@ -0,0 +1,93 @@
...
+        <node text="A.a() → B A">
+          <node text="A.a() → B A">
+            <node text="A.a() → B A">
+              <node text="A.a() → B A">
+                <node text="A.a() → B A">
+                  <node text="A.a() → B A">
+                    <node text="A.a() → B A">
+                      <Probably infinite part skipped>
+                      <Probably infinite part skipped>
+                      <Probably infinite part skipped>
```
alexander-doroshko: trick with `<Probably infinite part skipped>` doesn't work for me. I changed it to `<any/>` - it is used to stop expanding the tree in test at this point.

alexander-doroshko:
Some presentation problems:
image
- `null` for global functions
- too many `ArgumentError` on 2nd line
- too little visual difference between return type and containing class `bool num`

May be smth like this:
`[ClassName (if applicable)].[signature] (grey file name in parens)`?

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        return comp;
+      }
+      Collection<PsiElement> candidates = TargetElementUtil.getInstance().getTargetCandidates(dartReference);
+      if (!candidates.isEmpty()) return candidates.iterator().next();
+      return dartReference.resolve();
+    }
+    else {
+      DartComponent comp = PsiTreeUtil.getParentOfType(psiElement, DartComponent.class);
+      if (comp != null) {
+        return comp;
```
alexander-doroshko: 1. Not all component types are eligible for Call Hierarchy. You can check `DartComponentType`
2. Using `getParentOfType` shows Call Hierarchy for function even if caret is deep inside the implementation. I suggest to make it work only if caret is on function name identifier, as in Java.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyBrowser.java
```diff
@@ -0,0 +1,80 @@
...
+  public DartCallHierarchyBrowser(Project project, PsiElement method) {
+    super(project, method);
+  }
+
+  @Nullable
+  @Override
+  protected PsiElement getElementFromDescriptor(@NotNull HierarchyNodeDescriptor descriptor) {
+    if (descriptor instanceof DartHierarchyNodeDescriptor) {
+      DartHierarchyNodeDescriptor pyDescriptor = (DartHierarchyNodeDescriptor)descriptor;
+      return pyDescriptor.getPsiElement();
```
alexander-doroshko: Copy/paste from python? ;) Ok, no harm here

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/DartHierarchyUtil.java
```diff
@@ -55,4 +60,8 @@ public static DartClass findDartClass(@NotNull final Project project, @NotNull f
 
     return DartAnalysisServerService.getInstance().search_getTypeHierarchy(file, name.getTextRange().getStartOffset(), false);
   }
+
+  public static boolean isTypeExecutable(IElementType type) {
+    return FUNCTION_DEFINITION.contains(type) || type == FACTORY_CONSTRUCTOR_DECLARATION || type == NAMED_CONSTRUCTOR_DECLARATION;
```
alexander-doroshko: I guess checking DartComponentType would be better. When PSI contains all info we need and the task can be solved at PSI level we try not to go down to ASTNode, IElementType, etc. 
The same in `DartCallHierarchyProvider`, I'd prefer using `instanceof` instead of `getTokenType()`.
Not a major thing, just a common practice.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
```diff
@@ -0,0 +1,165 @@
...
+public class DartCalleeTreeStructure extends DartCallHierarchyTreeStructure {
+
+  public DartCalleeTreeStructure(Project project, PsiElement element, String currentScopeType) {
+    super(project, element, currentScopeType);
+  }
+
+  private static void getCallees(@NotNull PsiElement element, @NotNull List<PsiElement> results) {
+    DartComponentName name = (DartComponentName)element;
+    DartComponent decl = (DartComponent)name.getParent();
+    PsiFile file = PsiTreeUtil.getParentOfType(decl, PsiFile.class);
```
alexander-doroshko: Each PsiElement has method `getContainingFile()`

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
```diff
@@ -0,0 +1,165 @@
...
+  private static boolean isExecutable(DartNavigationTarget target) {
+    String kind = target.getKind();
+    return ElementKind.METHOD.equals(kind) ||
+           ElementKind.FUNCTION.equals(kind) ||
+           ElementKind.GETTER.equals(kind) ||
+           ElementKind.SETTER.equals(kind) ||
+           ElementKind.CONSTRUCTOR.equals(kind);
+  }
+
+  private static PsiElement getDeclaration(DartNavigationTarget target, PsiElement reference) {
```
alexander-doroshko: Take a look at `DartResolver#getElementForNavigationTarget`, it seems to be solving the same task

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
```diff
@@ -0,0 +1,165 @@
...
+
+  @NotNull
+  @Override
+  protected List<PsiElement> getChildren(@NotNull PsiElement element) {
+    final List<PsiElement> list = new ArrayList<>();
+    getCallees(element, list);
+    return list;
+  }
+
+  private static class ExitVisitor extends Error {
```
alexander-doroshko: That's the first direct inheritor of `java.lang.Error` in our codebase :). I guess it's because according to its javadoc `Error indicates serious problems that a reasonable application should not try to catch`. I think we'll finally end up without throwing whatever from visitors and reusing existing resolution logic instead.

Dart/testSrc/com/jetbrains/dart/analysisServer/DartCallHierarchyTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+  public void testFunctionCallers() throws Exception {
+    doCallHierarchyTest(null, "b", true, "B.dart", "A.dart");
+  }
+
+  public void testFunctionCallees() throws Exception {
+    doCallHierarchyTest(null, "a", false, "A.dart", "C.dart", "B.dart");
+  }
+
+  public void testFunctionRefCallers() throws Exception {
```
alexander-doroshko: This test passes against SDK 1.15 but fails against 1.16. Can you please take a look?

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
```diff
@@ -0,0 +1,165 @@
...
+  private static void resolveReferences(@NotNull DartComponent component,
+                                        @NotNull List<DartNavigationRegion> regions,
+                                        @NotNull List<PsiElement> results) {
+    component.acceptChildren(new DartRecursiveVisitor() {
+      @Override
+      public void visitReferenceExpression(@NotNull DartReferenceExpression reference) {
+        int offset = reference.getTextOffset();
+        List<DartNavigationTarget> targets = getRegionAt(offset, regions);
+        for (DartNavigationTarget target : targets) {
+          if (isExecutable(target)) {
```
alexander-doroshko: Just an idea: may be we could iterate `regions` and skip looking for the corresponding `DartReferenceExpression` at all in this method? Looks like all we need are resolution targets, but not references. Need to check that we won't get false targets this way.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
```diff
@@ -0,0 +1,82 @@
...
+      if (dartReference.getTokenType() == CALL_EXPRESSION) {
+        dartReference = getRightmostReference(dartReference.getFirstChild());
+      }
+      DartComponent comp = DartResolveUtil.findReferenceAndComponentTarget(dartReference);
+      if (comp != null) {
+        return comp;
+      }
+      Collection<PsiElement> candidates = TargetElementUtil.getInstance().getTargetCandidates(dartReference);
+      if (!candidates.isEmpty()) return candidates.iterator().next();
+      return dartReference.resolve();
```
alexander-doroshko: I didn't guess why the latest 3 lines are needed, tried to run tests without them and all still pass. I suggest to either add tests for this case or remove lines if they are not needed.

Dart/testData/analysisServer/callHierarchy/MultiCallees/MultiCallees_verification.xml
stevemessick: Thanks. I tried to find something that wouldn't cause warnings or errors to be printed but did not discover `<any/>`.

stevemessick:
Regarding presentation problems: I made two of the three changes you suggested. If we are going to stop displaying the class name along with the constructor name here then we should do that everywhere. That's a different topic, so for consistency I did not change constructor presentation. Also, I went with the file name, not file path or library name. Paths can get very long, and libraries can be hard to find. But file names are not unique so there's no perfect choice.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
stevemessick: I'm not sure what "even if caret is deep inside the implementation" means. The feature is designed to work when the cursor is on an identifier referenced from within a declaration.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyBrowser.java
stevemessick: Yeah, I'm glad to have the python sources as a reference.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
stevemessick: Done.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
stevemessick: I don't understand this comment.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
stevemessick: I suppose I could have re-used `org.apache.velocity.runtime.directive.StopCommand`, but adding that dependency seemed wrong. This is such a common pattern (outside the Jetbrains codebase ;-) that it seemed more intuitive to define a new class.

Dart/testSrc/com/jetbrains/dart/analysisServer/DartCallHierarchyTest.java
stevemessick: I have 1.17.0-dev.2.0. All tests pass.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
stevemessick: This looks like something left over from an earlier version. But there wouldn't have been any tests for something only run by the UI. I'll remove it.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
alexander-doroshko: > I don't understand this comment.

That comment was for the line where `DartCalleeTreeStructure.getDeclaration()` method is declared. The purpose of this method is to find `PsiElement` that corresponds to the `DartNavigationTarget` parameter. My comment is that there's already a method that performs this task, so it can be reused: `DartResolver.getElementForNavigationTarget()`.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
alexander-doroshko: You've removed 2 lines, I'll remove the 3rd one as well (`return dartReference.resolve();`), because resolution is already done 4 lines above in `findReferenceAndComponentTarget()`.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCallHierarchyProvider.java
alexander-doroshko: Ok, let me clarify with examples.
1. About eligible component types: `DartCallHierarchyProvider.getTarget()` should return `null` for this context:
```
class <caret>Foo {}
```

2. About caret deep inside some method implementation: I do not expect Hierarchy to work from this context:
```
  foo() {
    // tons of code
    if (true) {
      if (true) {
        if (true) {
          if (true) {
            // <caret>
          }
        }
      }
    }
  }
```

stevemessick: Thanks for the clarification. I fixed several problems.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/DartHierarchyUtil.java
stevemessick: Done

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
stevemessick: Done. Thanks, that simplified the code a lot. It appears to be ever-so-slightly slower, but no more than 1%.

Dart/src/com/jetbrains/lang/dart/ide/hierarchy/call/DartCalleeTreeStructure.java
stevemessick: I gave that a try. There may have been an easy way to filter false targets but it wasn't obvious to me. I decided to leave the original alone.

PR #381, opened 2016-06-02
Implement DartClass.isInheritor()

stevemessick:
@alexander-doroshko This is needed for the Method Hierarchy filtering.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/psi/DartClass.java
+++ b/Dart/src/com/jetbrains/lang/dart/psi/DartClass.java
@@ -57,4 +57,6 @@ public interface DartClass extends DartComponent {
 
   @Nullable
   DartComponent findNamedConstructor(String name);
+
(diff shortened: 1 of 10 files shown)


alexander-doroshko:
`isInheritor()` implementation waits for response from the server for up to 10 seconds. May be keep it somewhere in `DartHierarchyUtil` instead of `DartClass` just for the purpose of being less visible for future uses when we all forget about its potential performance problem?

Other question is where are you going to use this method. Having `isInheritor()` method is convenient. But in some cases it may appear to be more performant to work with the once obtained results of `DartHierarchyUtil#getTypeHierarchyItems` instead of calling `isInheritor()` (and thus `DartAnalysisServerService.search_getTypeHierarchy`) multiple times.

stevemessick:
OK, I'll move it to DartHierarchyUtil.

PR #382, opened 2016-06-03
Method Hierarchy

stevemessick:
@alexander-doroshko Here's the implementation of Method Hierarchy.

I took your suggestion and rewrote it so there was no need to do inheritance checking outside the recursive method that builds the tree, which has the type hierarchy info.

There is one difference w.r.t Java: for some reason that I don't understand Java displays no icon for abstract methods. This implementation does display a '+' icon. If you want consistency with Java let's file a bug. I'll fix it soon.


alexander-doroshko:
Thanks! 
One thing noticed: it tries to work for targets that can not participate in hierarchy like constructors and static methods.

PR #383, opened 2016-06-15
Callbug

stevemessick:
@alexander-doroshko This fixes the bug in call hierarchy where nothing was displayed for calls represented by unnamed functions. It defines getPresentation() for functions, which requires a small change to the grammar.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/psi/impl/DartFunctionExpressionBase.java
+++ b/Dart/src/com/jetbrains/lang/dart/psi/impl/DartFunctionExpressionBase.java
@@ -1,21 +1,33 @@
 package com.jetbrains.lang.dart.psi.impl;
 
+import com.intellij.icons.AllIcons;
 import com.intellij.lang.ASTNode;
 import com.intellij.navigation.ItemPresentation;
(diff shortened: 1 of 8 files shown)


Dart/src/com/jetbrains/lang/dart/psi/impl/DartFunctionExpressionBase.java
```diff
@@ -0,0 +1,62 @@
...
+    baseIcon.setIcon(icon, 0);
+    Icon visibility = PlatformIcons.PRIVATE_ICON;
+    baseIcon.setIcon(visibility, 1);
+    return baseIcon;
+  }
+
+  @Override
+  public ItemPresentation getPresentation() {
+    final PsiElement parent = PsiTreeUtil.getParentOfType(this, DartFunctionExpressionBase.class, DartMethodDeclaration.class,
+                                                          DartFunctionDeclarationWithBodyOrNative.class);
```
alexander-doroshko: There are also other eligible parents: factory constructor, named constructor, getter, setter, named function declared inside other function (DartFunctionDeclarationWithBody).

PR #386, opened 2016-06-22
Multiexec

stevemessick:
@alexander-doroshko Fix presentation for more types of executable wrappers.

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -103,6 +103,9 @@
     <codeInsight.implementMethod language="Dart"
                                  implementationClass="com.jetbrains.lang.dart.ide.generation.DartImplementMethodHandler"/>
 
+    <statementUpDownMover implementation="com.jetbrains.lang.dart.ide.moveCode.DartStatementMover" id="statement" order="before component" />
+    <statementUpDownMover implementation="com.jetbrains.lang.dart.ide.moveCode.DartComponentMover" id="component" order="before xml" />
+
     <fileBasedIndex implementation="com.jetbrains.lang.dart.ide.index.DartImportAndExportIndex"/>
     <fileBasedIndex implementation="com.jetbrains.lang.dart.ide.index.DartPartUriIndex"/>
     <!--<fileBasedIndex implementation="com.jetbrains.lang.dart.ide.index.DartSourceIndex"/>-->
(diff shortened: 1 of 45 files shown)


Dart/resources/META-INF/plugin.xml
```diff
@@ -103,6 +103,9 @@
     <codeInsight.implementMethod language="Dart"
                                  implementationClass="com.jetbrains.lang.dart.ide.generation.DartImplementMethodHandler"/>
 
+    <statementUpDownMover implementation="com.jetbrains.lang.dart.ide.moveCode.DartStatementMover" id="statement" order="before component" />
+    <statementUpDownMover implementation="com.jetbrains.lang.dart.ide.moveCode.DartComponentMover" id="component" order="before xml" />
```
alexander-doroshko: Why relation to XmlMover is important?

Dart/resources/META-INF/plugin.xml
```diff
@@ -103,6 +103,9 @@
     <codeInsight.implementMethod language="Dart"
                                  implementationClass="com.jetbrains.lang.dart.ide.generation.DartImplementMethodHandler"/>
 
+    <statementUpDownMover implementation="com.jetbrains.lang.dart.ide.moveCode.DartStatementMover" id="statement" order="before component" />
```
alexander-doroshko: Technology-specific ids are preferred, like DartStatementMover, otherwise someone else (e.g. in 3rd party plugin) can use the same id and plugin loading will fail.

Dart/src/com/jetbrains/lang/dart/ide/moveCode/DartStatementMover.java
```diff
@@ -0,0 +1,380 @@
...
+import com.jetbrains.lang.dart.util.DartRefactoringUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Move executable statements within a method or function. Statements cannot be moved outside a component.
+ * TODO: What about moving statements from a local function to the surrounding method?
+ */
+public class DartStatementMover extends LineMover {
+  private PsiElement statementToSurroundWithCodeBlock;
```
alexander-doroshko: IDE shows me 2 warnings at this line

alexander-doroshko:
Statement/component mover is cool but very complicated feature. Thank you for starting with it! Be ready that fixing and supporting it may take months or even years.
I'm afraid thorough review requires the same time resources that you spend on writing it, so I did only basic review and a bit of manual testing. First thing to fix: comments and doc comments located before declaration should move up/down together with the declaration itself. Code to play:
```
// awesome
// variables
var a, b;

/**
 * doc
 */
var x;

/// useful
/// class
class A{
}

/**
 * doc
 */
foo(){
}
```

PR #387, opened 2016-06-29
Work on moving comments

stevemessick:
@alexander-doroshko Incremental improvement to Move Statement.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/moveCode/DartComponentMover.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/moveCode/DartComponentMover.java
@@ -9,13 +9,12 @@ import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
+import com.intellij.psi.PsiWhiteSpace;
+import com.intellij.psi.impl.source.tree.LeafPsiElement;
 import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.jetbrains.lang.dart.DartTokenTypesSets;
-import com.jetbrains.lang.dart.psi.DartComponent;
-import com.jetbrains.lang.dart.psi.DartFile;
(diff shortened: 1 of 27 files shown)


alexander-doroshko:
Hm, I'm afraid I do not agree with test data. See my comment and sample code in it: https://github.com/JetBrains/intellij-plugins/pull/386#issuecomment-228153133
See how similar things work in Java: comment and declaration behind it are moved a a whole.

alexander-doroshko:
Trying to move `var y;` down leads to exception
```
class Foo {
  var y;
  /*
   * doc
   */
  var x;
}
```

stevemessick:
Fair enough. My concern for this work is to identify comment blocks. Attaching them to declarations comes later.

alexander-doroshko:
Ok, then I suggest to integrate this pull request later together with the upcoming update.

Consequent single-line comments (both // and ///) should be treated as a block as well.

General suggestion: it seems to be more accurate to look at PsiElement.getNote().getElementType() instead of comparing PsiElement.getText() with string.

stevemessick:
On Thu, Jun 30, 2016 at 8:07 AM, Alexander Doroshko <
notifications@github.com> wrote:

> Ok, then I suggest to integrate this pull request later together with the
> upcoming update.
>
Is this the highest-priority task? Moving comments as a single block
instead of individual lines is already a huge usability improvement. I
thought moving statements into/out of blocks would be more important than
attaching comments to declarations.

> Consequent single-line comments (both // and ///) should be treated as a
> block as well.
>
I agree, but I haven't wanted to think about that yet.

> General suggestion: it seems to be more accurate to look at
> PsiElement.getNote().getElementType() instead of comparing
> PsiElement.getText() with string.
>
I thought you said to not fetch the AST if everything needed was in the PSI
nodes? I'd prefer to use getElementType() but I had been avoiding it
because of some earlier comment.

alexander-doroshko:
I'd rate comment+declaration handling as a must have for component mover feature.

> I thought you said to not fetch the AST if everything needed was in the PSI
nodes?

Yes, I said that checking `instanceof` for `PsiElement` is better than checking its `ASTNode.getElementType()`. But checking AST element type is better than using `PsiElement.getText()` :-).

PR #388, opened 2016-07-08
Update to Move Statement

stevemessick:
@alexander-doroshko Incremental improvement to Move Statement. Keeps comments with declarations.


alexander-doroshko:
Move main() up throws exception
```
import 'dart:core';
//

void main() {
}
```

alexander-doroshko:
Try to move `var x` up several times :)
```
class Foo {
  // foo
  var x;
  var y;
}
```

PR #389, opened 2016-07-18
Move Statement update

stevemessick:
@alexander-doroshko Here are some changes to handle two cases you mentioned in a previous review, from July 11. Sorry I didn't get to this earlier -- I was working on something else last week.

There are still problems with comments at the edge of a class declaration that I need to work on. I'll take another look when I get back in August.


alexander-doroshko:
Thanks, pushed.
Hint: when you had committed something (not yet pushed) and then noticed that you need to add some tweaks like reformat or something - you can check 'Amend commit' check box in IntelliJ's Commit dialog - this way you'll update previous commit instead of creating another one.

PR #390, opened 2016-07-29
Added support for resolving partial file paths from names

dudabone:
This feature adds support for resolving partial file names from the declaration. For example, if a certain template has this partial declaration:

{{>path/to/partial}}

Then when using Ctrl + mouse click on the partial path, the plugin will try to resolve this path to a real template file. In addition, it supports autocompletion of partial names / paths.

I found this feature useful because I am using libraries such as Handlebars.java and grunt-contrib-handlebars that allow declaring such paths to partials, with automatic path prefix and file extension resolving. With this capability, users may find it easy to work with such libraries. I will be glad to hear your opinion whether you find it useful and if it should be merged to the main repository. I will also add tests of course.


anstarovoyt:
@dudabone hi. I cannot find any documentation where the partials are used as file paths. Could you please share more information?

dudabone:
Hi @anstarovoyt, it is not that clear from the documentation, but in handlebars.java, there's:
https://jknack.github.io/handlebars.java/loader.html
https://github.com/jknack/handlebars.java/blob/master/README.md
https://github.com/jknack/handlebars.java/blob/master/handlebars-springmvc/src/main/java/com/github/jknack/handlebars/springmvc/HandlebarsViewResolver.java

And in grunt-contrib-handlebars, you can actually register the precompiled partials using the file path as a key in the partials map. Check out the partials section here:

http://danburzo.ro/grunt/chapters/handlebars/

Please let me know if you need more info.

anstarovoyt:
meteor (for example) uses the tags "{{> name}}" different way (they include other templates). I suggest to create a separate plugin for this until we will provide more generic solution. Sorry for the long waiting for decision.

PR #392, opened 2016-08-12
remove the 'Observatory listening on' message from the console output

devoncarew:
This PR removes the `Observatory listening on http://...` message from the console output on Dart CLI runs. Printing it each time adds clutter to the runs, especially as there's a toolbar button to open the Observatory directly.

@alexander-doroshko

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -191,6 +191,7 @@
     <runConfigurationProducer implementation="com.jetbrains.lang.dart.ide.runner.test.DartTestRunConfigurationProducer"/>
 
     <programRunner implementation="com.jetbrains.lang.dart.ide.runner.DartRunner"/>
+    <consoleInputFilterProvider implementation="com.jetbrains.lang.dart.ide.runner.DartConsoleInputFilterProvider"/>
 
     <localInspection bundle="com.jetbrains.lang.dart.DartBundle" key="outdated.dependencies.inspection.name"
                      groupName="Dart" enabledByDefault="true" level="WARNING" language="Dart"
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
@devoncarew 
It might be a matter of preference. I realized that I didn't think of this line as about just a garbage. My experience shows that users often do not notice additional buttons on the Run/Debug toolbar. How strong is your intention to wipe it out?

devoncarew:
Totally a matter of style :)

I think we print out too much stuff when running apps. Some data is noise to the user (and some useful). For instance, the cli args to set the debugger port is not very useful to the user; things like sending args to the app being run would be. For instance, the output for a hello world app is:

```
/Users/devoncarew/projects/flutter/flutter/bin/cache/dart-sdk/bin/dart --ignore-unrecognized-flags --checked --enable-vm-service:56361 --trace_service_pause_events /Users/devoncarew/projects/flutter/flutter/packages/flutter_tools/bin/flutter_tools.dart doctor
Observatory listening on http://127.0.0.1:56361
Hello world!

Process finished with exit code 0
```

A more terse representation might be:

```
dart --checked bin/flutter_tools.dart doctor

Hello world!

exited with code 0
```

You convey all the same info but in an easier to digest format :)

> How strong is your intention to wipe it out?

I would prefer to, esp. given the existence of the observatory button, but do agree it's a style issue, and not blocking anything.

alexander-doroshko:
Take a look at `com.intellij.execution.ConsoleFolding`. It is possible to fold CLI args to a short string like this:
image

devoncarew:
+1, that would be great to do

devoncarew:
I added an issue to track the console folding work: https://youtrack.jetbrains.com/issue/WEB-22970

jwren:
+1

alexander-doroshko:
@jwren, WEB-22970 has already been fixed at the moment of your '+1' :)

@devoncarew, speaking about 'Observatory ...' line, may be fold it like this:
image

Or you still think that it should be removed completely?

devoncarew:
That's much cleaner! I think it's a bit of a duplication to both show the URI and have a button, but that single line does make it very clean.

Happy to close this is you'd prefer the folding solution.

devoncarew:
(closing)

PR #396, opened 2016-08-18
Additional filter in the Dart Analysis view, filter by Directory.

jwren:


Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsFilter.java
```diff
@@ -109,6 +108,14 @@ public boolean include(@NotNull final DartProblem problem) {
       return false;
     }
 
+    if (myFileFilterMode == FileFilterMode.Directory && myCurrentFile != null) {
```
alexander-doroshko: If `currentFile` is `null` shouldn't table be empty?

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsFilter.java
```diff
@@ -109,6 +108,14 @@ public boolean include(@NotNull final DartProblem problem) {
...
+      VirtualFile parent = myCurrentFile.getParent();
+      VirtualFile child = problem.getFile();
+      if (child != null && parent != null && !child.getPath().startsWith(parent.getPath())) {
```
alexander-doroshko: I'd add slash to avoid false positives:
`...startsWith(parent.getPath() + "/")`

BTW with this logic files in subdirectories are also included. Is it your intent?

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsTableModel.java
```diff
@@ -367,6 +367,9 @@ public String getStatusText() {
         case Package:
           b.append("Filtered by current package");
           break;
+        case Directory:
+          b.append("Filtered by directory");
```
alexander-doroshko: Shouldn't it be 'Filtered by **current** directory', in compliance with the radio button label?

BTW do you know preferences about using 'directory' vs 'folder' in the IDE UI?

jwren: Yes, **current** should be added.

BTW do you... no I do not.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsFilter.java
alexander-doroshko: I've rewrote it like this. Let me know in case of objections.
```
    if (myFileFilterMode == FileFilterMode.Directory) {
      if (myCurrentFile == null) return false;
      ...
    }
```

PR #397, opened 2016-08-19
Update Dart Analysis progress task on analysis.errors events

jwren:


alexander-doroshko:
@jwren Sorry, I had to revert this commit because of a Project leak

devoncarew:
@alexander-doroshko, just for my own IntelliJ edification, can you clarify what a project leak is? Thanks!

alexander-doroshko:
IntelliJ has components/services of different levels: module-level (one per module), project-level (one per project) and application-level. Creation and disposal is controlled by the Platform. E.g. when project is closed all project-level components are disposed and there must be no strong references to them from components that remain alive. Project-level components may have references to app-level ones, but not vice versa.
`DartApplicationServerService` is an application-level component, nothing happens to it when project is closed. But it keeps strong reference to Project via its field (originally named `myStatusBarTask` in this PR, later rnamed to `myServerProgressTask`), see memory snapshot:
image

Strong references to disposed Project object do not allow GC to remove it and thousands of other objects tied to this Project also remain leaked.

I could find this earlier, instantiating an app-level field [inside projects loop](https://github.com/JetBrains/intellij-plugins/pull/397/files#diff-26c8291690a1272db82164d3970eab40R235) must have attracted my attention.

PR #398, opened 2016-08-22
remove noop flags on launch

devoncarew:
- remove two flags passed into the VM when launching Dart CLI apps (`--ignore-unrecognized-flags` and `--trace_service_pause_events`)

We're displaying a lot of text when launching apps; this trims it down a bit.

@alexander-doroshko


alexander-doroshko:
@devoncarew Provided that we'll implement command line folding to just 'dart foo.dart args' do you still think these 2 options should be removed?
`--ignore-unrecognized-flags` looks useless indeed as we do not use backward-incompatible flags. As for `--trace_service_pause_events` - I do not know. It may help those who added `--pause-isolates-on-exit` VM option manually, but I guess they are 0.001%.

devoncarew:
Yup, I think they can be removed :)

Do not know how `trace_service_pause_events` and `pause-isolates-on-exit` are related, but people are still able to manually add the more esoteric VM flags if they desire.

alexander-doroshko:
> how `trace_service_pause_events` and `pause-isolates-on-exit` are related

With `pause-isolates-on-exit` the app never exits, even when it ends it execution, and `trace_service_pause_events` shows a hint that to exit the app user should open Observatory and resume (i.e. kill) finished isolate(s) there.

PR #399, opened 2016-08-22
regenerate the latest analysis server files

devoncarew:
- regenerate the latest analysis server files

@jwren @alexander-doroshko

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
+++ b/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
@@ -122,7 +122,7 @@ public class AnalysisError {
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
-    String code = jsonObject.get("code").getAsString();
+    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
     Boolean hasFix = jsonObject.get("hasFix") == null ? null : jsonObject.get("hasFix").getAsBoolean();
     return new AnalysisError(severity, type, location, message, correction, code, hasFix);
   }
@@ -219,7 +219,9 @@ public class AnalysisError {
     if (correction != null) {
       jsonObject.addProperty("correction", correction);
     }
-    jsonObject.addProperty("code", code);
+    if (code != null) {
+      jsonObject.addProperty("code", code);
+    }
     if (hasFix != null) {
       jsonObject.addProperty("hasFix", hasFix);
     }
--- a/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
+++ b/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
@@ -54,11 +54,6 @@ public class AnalysisError {
   private final String type;
 
   /**
-   * The code of the error that can be used for error suppresion.
-   */
-  private final String code;
-
-  /**
    * The location associated with the error.
    */
   private final Location location;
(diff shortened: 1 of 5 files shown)


Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
```diff
@@ -119,12 +119,12 @@ public boolean equals(Object obj) {
   public static AnalysisError fromJson(JsonObject jsonObject) {
     String severity = jsonObject.get("severity").getAsString();
     String type = jsonObject.get("type").getAsString();
-    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
+    String code = jsonObject.get("code").getAsString();
```
alexander-doroshko: @devoncarew This makes Dart plugin incompatible with SDKs older than 1.16 because of an NPE. I guess there's and error in spec that is used to generate this code and 'code' attribute should be marked as optional there. 

// cc @pq (related commit: https://github.com/dart-lang/sdk/commit/e7e205f00b24faf9ee5cdf0baf21b8a0a04f8c3a)

devoncarew: I agree it looks like `code` should be optional. @pq, @brianwilkerson, thoughts?

pq: Making it optional works for me since it addresses the immediate but it's a bit of a hack.  In the long run we probably need to have a better means to evolve server protocols.  Adding new data really shouldn't blow up clients IMO.

Anyway, hopping of the soapbox, if making this optional works, I say go for it! 👍

pq: Regarding the larger evolving API question, what did you do in Atom @devoncarew?  I'm guessing you check for the API version and have different handlers?

devoncarew: For atom, the library we use is lenient about missing fields. Optional fields are tagged w/ an `@optional` annotation, but that's really just an FYI to the API users.

bwilkerson: I suppose that in some sense we could mark all fields returned by the server as being optional, but I think it's useful to the implementors of clients to know which fields they can count on being present, so I'd rather not change it in the spec.

That leaves us with three options that I can think of:

1. Decide that it isn't really a problem that we can't use an SDK older than 1.16, given that we're about to ship 1.19.

2. Manually change the copy of the generated Java to tolerate missing data. I don't know how frequently the API interface classes are updated, so I don't know how often this would need to be done before we stopped caring about 1.16.

3. Update the generator to make the code always be backward compatible by always tolerating missing data.

devoncarew: It's just this particular field which is the issue. Given the current generated Java impl., adding a new, non-optional field to an existing class is a breaking change. I don't know if you speak to what's considered a breaking change in the protocol spec doc, but it would strike me as strange to support missing required fields.

I think marking the field as optional is a pretty low-cost solution. I'm happy to add some docs to the field - `only optional for backwards compatibility reasons`. What do you think?

devoncarew: Per an off-line discussion, I'll manually edit these files to make the library tolerant of missing `code` fields.

devoncarew:
@alexander-doroshko, ptal, updated

PR #401, opened 2016-08-22
change to a method which doesn't throw exceptions

devoncarew:
When debugging w/ break on exceptions on, we break on `process.exitValue()` frequently - it throws an exception when called if the process is still running.

This changes the isOpen() impl to process.isAlive(), which doesn't throw.

@alexander-doroshko


alexander-doroshko:
Looking in JDK sources I see that `java.lang.Process.isAlive()` is doing exactly the same :).

devoncarew:
Closing this PR out -

PR #404, opened 2016-08-23
Add default implementations of default methods

stevemessick:
@alexander-doroshko These methods used to be implemented by ProjectLifecycleListener.Adapter but that was deprecated. New implementations were not provided, so here they are.

I only noticed this while testing a plugin I built from sources, as per the artifact-building instructions you sent me.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
@@ -55,6 +55,14 @@ public class DartServerRootsHandler {
           DartAnalysisServerService.getInstance().updateVisibleFiles();
         }
       }
+
+      @Override
+      public void beforeProjectLoaded(@NotNull Project project) {
+      }
+
+      @Override
+      public void projectComponentsInitialized(@NotNull Project project) {
+      }
     });
   }
 


alexander-doroshko:
@stevemessick Sorry, I do not understand the problem that is solved by this PR. With JDK 1.8 you do not need to implement methods of interface that itself has default implementations.

stevemessick:
Without these methods I get runtime errors. I have not read the spec for
default methods. Do they function at both runtime and compile time? If so,
these methods should not be needed, but I still got errors when I built the
Dart plugin and installed it in the stable Webstorm.

alexander-doroshko:
With JDK 1.8 both at compile and at run time it should work of course. What JDK does you WebStorm use (see About banner)?
You may switch IDE boot JDK to 1.8 (Help | Find Action | Switch IDE Boot JDK), or (**much better**) just install today's [WebStorm 2016.3 EAP](https://blog.jetbrains.com/webstorm/2016/08/webstorm-2016-3-early-access-preview/).

stevemessick:
It uses JRE 1.8.

alexander-doroshko:
Ok, sorry, I understood the problem. Current Dart plugin sources in 'master' branch are not compatible with WebStorm 2016.2. You need to install today's EAP anyway.

stevemessick:
I see. I will delete this PR. In the near future we need to be able to
distribute flutter and dart plugins easily that do not require EAP. We'll
have to discuss that further, probably at our Thursday meeting next week.

alexander-doroshko:
> In the near future we need to be able to distribute flutter and dart plugins easily that do not require EAP. We'll have to discuss that further

Plugins for 2016.2 release must be developed in '162' branch, plugins for 2016.3 - currently in master branch, but in a couple of months special '163' branch will be created. Plugin specifies IDE versions it is applicable for (since/until versions), so it is possible to release plugin builds for different IDE versions independently.

PR #405, opened 2016-08-30
remove the older (pre-Dart 1.14) debugger client

devoncarew:
- remove the older (pre-Dart 1.14) debugger client
- fix https://youtrack.jetbrains.com/issue/WEB-22983

@alexander-doroshko for review

This doesn't address any other places where we handle older SDK versions. @jwren, @stevemessick, @pq, @danrubel, thoughts on removing pre-1.14 debugging support? 1.14 has been out since Jan 28.


danrubel:
Removing old functionality specific to pre-1.14 SDKs makes sense to me.

pq:
👍 I'm all for removing pre-1.14 support altogether.

PR #407, opened 2016-09-01
implement async step; update debugger library

devoncarew:
- implement async stepping (fix https://youtrack.jetbrains.com/issue/WEB-22910)
- update the debugger library to the latest; this properly supports optional fields in the protocol (and avoids an NPE), and exposes a few fields that had not yet been exposed

@alexander-doroshko, @jwren, @danrubel

This implementation essentially store the `atAsyncSuspension` field when an isolate pauses, and when the user steps, we either issue a `StepOption.OverAsyncSuspension` or a `StepOption.Over`.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
@@ -64,7 +64,7 @@ public class DartVmServiceListener implements VmServiceListener {
         ApplicationManager.getApplication().executeOnPooledThread(() -> {
           final ElementList<Breakpoint> breakpoints = event.getKind() == EventKind.PauseBreakpoint ? event.getPauseBreakpoints() : null;
           final InstanceRef exception = event.getKind() == EventKind.PauseException ? event.getException() : null;
-          onIsolatePaused(event.getIsolate(), breakpoints, exception, event.getTopFrame());
+          onIsolatePaused(event.getIsolate(), breakpoints, exception, event.getTopFrame(), event.getAtAsyncSuspension());
         });
         break;
       case PauseExit:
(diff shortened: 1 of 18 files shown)


danrubel:
LGTM

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
```diff
@@ -129,6 +136,8 @@ private void onIsolatePaused(@NotNull final IsolateRef isolateRef,
       final boolean suspend = myDebugProcess.getSession().breakpointReached(xBreakpoint, logExpression, suspendContext);
       if (!suspend) {
         myDebugProcess.getVmServiceWrapper().resumeIsolate(isolateRef.getId(), null);
+      } else {
+        myDebugProcess.getSession().positionReached(suspendContext);
```
alexander-doroshko: `breakpointReached()` is already called 3 lines above, why do you think you need to call `positionReached()`?

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
```diff
@@ -118,6 +119,12 @@ private void onIsolatePaused(@NotNull final IsolateRef isolateRef,
 
       final XLineBreakpoint<XBreakpointProperties> xBreakpoint = myBreakpointHandler.getXBreakpoint(vmBreakpoints.get(0));
 
+      if (xBreakpoint == null) {
+        myLatestSourcePosition = sourcePosition;
+        myDebugProcess.getSession().positionReached(suspendContext);
+        return;
+      }
```
alexander-doroshko: I guess this `if` is for stopping on breakpoints set in the Observatory, right?
If so, please make separate feature as a separate commit next time. Thanks!

alexander-doroshko:
Sorry for delay and thanks for the feature! Pushed.
I've only removed `positionReached()` call after `breakpointReached()`.
Next time please prefer separate commits for separate features. For example: update VM Service drivers; handle breakpoints set in the Observatory; async step.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
devoncarew: The VM implements the async step feature by setting temporary VM-only breakpoints. When we pause, we see it as stopping at a breakpoint, but the location doesn't match up to any of our known breakopints.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
devoncarew: Hmm, I suppose I was unaware that breakpointReached() implied positionReached()? In any case, thanks for the  review! I'll re-test once this lands.

devoncarew:
> Next time please prefer separate commits for separate features. For example: update VM Service drivers; handle breakpoints set in the Observatory; async step.

Will do re: separate features per PR; in this particular case, this was dealing w/ the internal implementation of async stepping.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
alexander-doroshko: Ok, thanks for letting me know! I'll add comment to the code.

PR #410, opened 2016-09-05
Some UI tweaks to the Dart problems view

devoncarew:
Some UI tweaks to the Dart problems view:

- in the filter dialog for the Dart problems view, rename `content root` to `module` (I think this maps closer to IntelliJ concepts the user would be familiar with)
- rename `package` to `Dart package`

screen shot 2016-09-05 at 12 19 55 pm

- if there are no issues in the problems view, hide the status label at the bottom
- change the summary text at the bottom to just show counts for the items displayed in the errors view, and if a filter is active, tell the user what we're filtering on

screen shot 2016-09-05 at 12 18 50 pm

@alexander-doroshko @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
@@ -25,7 +25,7 @@ public class DartProblem {
 
   @Nullable private VirtualFile myFile;
   @Nullable private VirtualFile myPackageRoot;
-  @Nullable private VirtualFile myModuleRoot;
+  @Nullable private VirtualFile myContentRoot;
 
   private String myPresentableLocationWithoutLineNumber;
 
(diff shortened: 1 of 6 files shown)


alexander-doroshko:
> rename `content root` to `module`

I'm afraid it is 
a) not true for IntelliJ IDEA (each module may have any number of content roots, including 0, 1 and >1)
b) not applicable for smaller IDEs because they do not have `module` concept at all.

[Content root](https://www.jetbrains.com/help/idea/content-root.html) is one of the basic concepts in any JetBrains IDE, and this term can be met in Project Structure dialog (IntelliJ IDEA) or in Preferences | Directories (other IDEs).

You should understand that for our IDEs there are no such things as 'project root' or 'module root'. There are:
- folder where IDE stores project files (.idea folder), and it is never of interest  for us
- IntelliJ IDEA project may contain zero or more modules. Smaller IDEs as a rule contain 1 module (that's why there's no `module` concept there)
- module settings are stored in *.iml file and location of this file is not connected to .idea folder, module roots, or whatever else, and never of interest for us
- each module may contain zero or more content roots, in general case location of these directories are absolutely independent and not related to anything else
- each content root may have (all or some, depending on the IDE) source folders, test source folders, resource folders, test resources, excluded folders. At any level inside content root.

So, 'project root' term makes any sense only when project contains one module and this module contains one content root (and usually this content root is the folder where .idea folder is stored). But we must not rely on any specific project structure and that's why we avoid 'project root' and 'module root' terms.

alexander-doroshko:
> rename `package` to `Dart package`

No objections if everyone agrees. May be do the same for filter status in the status text?

alexander-doroshko:
> if there are no issues in the problems view, hide the status label at the bottom

That's fine if there are no filters enabled. If there are issues, but all filtered out, I think it makes sense to say that filtering is enabled.

alexander-doroshko:
> change the summary text at the bottom

May be loop Luke as UX expert or our UX expert in JetBrains?
For the reference, here's what we compare:
image

image

devoncarew:
OK, I updated the PR to revert the content root => module filter text change. Sounds like if we wanted to filter at a module level, we'd need to accept files in any of the content roots for that module.

> That's fine if there are no filters enabled. If there are issues, but all filtered out, I think it makes sense to say that filtering is enabled.

Makes sense; I updated the PR to do that.

screen shot 2016-09-06 at 11 40 39 am

screen shot 2016-09-06 at 11 40 56 am

> May be loop Luke as UX expert or our UX expert in JetBrains?

Happy to! @lukechurch for thoughts :)

I think the updated status line summary is a bit more terse, and a bit easier for users to parse. Most of what they want to know will be in the body of the problems view itself.

alexander-doroshko:
image
:)

alexander-doroshko:
[Pushed](https://github.com/JetBrains/intellij-plugins/commit/883fcbf1f121b18acff8ce02fd2ac86b651d95c7) manually (to avoid rename/revert in commits) + added fix for `( and severity)`.

devoncarew:
> added fix for ( and severity).

oops :)

PR #411, opened 2016-09-11
implement run to cursor

devoncarew:
- implement the run to cursor functionality (using temporary breakpoints, removed on the next breakpoint pause event)
- fixed an issue found as part of this work, where we were not removing vm breakpoints from a collection when an IntelliJ breakpoint was removed (https://github.com/JetBrains/intellij-plugins/compare/master...devoncarew:run_to_cursor?expand=1#diff-aeaaa0c5a2f9ccc7970be0eac34e566cR49) (this fixes force run to cursor)
- (also, some minor renames in associated files)
- fix https://youtrack.jetbrains.com/issue/WEB-17091

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
@@ -345,8 +345,11 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
 
   @Override
   public void runToPosition(@NotNull XSourcePosition position, @Nullable XSuspendContext context) {
-    // todo implement
-    resume(context);
+    if (myLatestCurrentIsolateId != null && mySuspendedIsolateIds.contains(myLatestCurrentIsolateId)) {
+      // Set a temporary breakpoint and resume.
+      myVmServiceWrapper.addTemporaryBreakpoint(position, myLatestCurrentIsolateId);
+      myVmServiceWrapper.resumeIsolate(myLatestCurrentIsolateId, null);
+    }
   }
 
   public void isolateSuspended(@NotNull final IsolateRef isolateRef) {
@@ -410,7 +413,7 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
       if (myDASExecutionContextId != null && !uriByIde.startsWith(DartUrlResolver.DART_PREFIX)) {
         final String uriByServer = DartAnalysisServerService.getInstance().execution_mapUri(myDASExecutionContextId, file.getPath(), null);
         if (uriByServer != null) {
-          return mayBeAppendOneMoreUri(file, uriByServer);
+          return maybeAppendOneMoreUri(file, uriByServer);
         }
       }
     }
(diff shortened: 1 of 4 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
```diff
@@ -198,6 +198,12 @@ public void addBreakpoint(@NotNull final String isolateId,
                             @NotNull final XLineBreakpoint<XBreakpointProperties> xBreakpoint,
                             @NotNull final VmServiceConsumers.BreakpointConsumerWrapper consumer) {
     final XSourcePosition position = xBreakpoint.getSourcePosition();
+    addBreakpointFromPosition(isolateId, position, consumer);
+  }
+
+  public void addBreakpointFromPosition(@NotNull final String isolateId,
+                                        @NotNull XSourcePosition position,
+                                        @NotNull final VmServiceConsumers.BreakpointConsumerWrapper consumer) {
     if (position == null || position.getFile().getFileType() != DartFileType.INSTANCE) {
```
alexander-doroshko: Just to make sure that inspections work in your setup, do you see this yellow code?
image
Unless there are reasons not to do so try to avoid introducing new warnings. I'll change `@NotNull` to `@Nullable` in `XSourcePosition position` argument annotation to fix yellow code.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -345,8 +345,11 @@ public void startPausing() {
 
   @Override
   public void runToPosition(@NotNull XSourcePosition position, @Nullable XSuspendContext context) {
-    // todo implement
-    resume(context);
+    if (myLatestCurrentIsolateId != null && mySuspendedIsolateIds.contains(myLatestCurrentIsolateId)) {
+      // Set a temporary breakpoint and resume.
+      myVmServiceWrapper.addTemporaryBreakpoint(position, myLatestCurrentIsolateId);
+      myVmServiceWrapper.resumeIsolate(myLatestCurrentIsolateId, null);
```
alexander-doroshko: Do we want to set temporary breakpoint in all isolates and resume all isolates?

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
devoncarew: I don't remember seeing any, but will double check.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
devoncarew: I'm assuming that when the user does run-to-cursor, they're expecting the current isolate to resume, and break when it reaches the current cursor position. Other isolates may be paused debugging completely separate bits of code and it might be unexpected to resume them.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
devoncarew: But yeah, +1 to not introducing new warnings :)

devoncarew: (looks like my inspections are working correctly, I just didn't notice this one)

alexander-doroshko: No worries!
Hint: file analysis status is shown in the editor's top-right, hover shows detailed summary. F2 / Shift+F2 jumps to next/previous warning. Also there's 'Perform code analysis' check box in the Commit dialog, though I do not use it.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
alexander-doroshko: ok

PR #415, opened 2016-09-11
fix missing case inspection

devoncarew:
Fix an inspection warning about missing switch case statements.


devoncarew:
@alexander-doroshko

PR #416, opened 2016-09-12
Add support for Reveal 2

tonyarnold:
This PR introduces basic support for Reveal 2 and upcoming future releases. 

Some minor work has been started in this PR to support both automatic and manual selection when there are multiple copies of Reveal installed. At this point, there is no UI for this feature, so it is not exposed to the user. 

As always, my Java is pretty rusty so please feel free to give me feedback if there are things that need to be fixed.


tonyarnold:
It's worth noting that 28cb46381ab3aec21a645bc672a1764009749314 was necessary, as Reveal is looking for `UIDevice.name`, whereas AppCode's supporting library appears to be returning the result of `simctl`/CoreSimulator. 

In normal use, it shouldn't matter unless the user is running multiple versions of exactly the same app with the same bundle identifier. It would be nice to fix this problem on one or both ends in future, but this is OK for now, I think.

tonyarnold:
This update is not currently compatible with iOS 10, tvOS 3 and watchOS 3 as I could not find enough exposed API to identify this in the plugin's available libraries.

klikh:
Have you submitted a contributor license agreement? If not, please download the PDF from http://www.jetbrains.org/display/IJOS/Contributor+Agreement and then print it out, sign (mention you github account name there), scan and email to contribute@jetbrains.com

PR #419, opened 2016-09-13
cosmetic tweaks to the Dart error view

devoncarew:
Some cosmetic tweaks to the Dart error view:
- replace the death's head icon with a stop-and-restart icon
- reorganize the actions so that the kill server action isn't right next to the reanalyze action (to eliminate accidental presses / unindented restarts)
- group the other, more related actions together

screen shot 2016-09-13 at 12 29 56 pm

@jwren @alexander-doroshko


alexander-doroshko:
Scull icon was discussed with Luke and the rest of the team and it really emphasize that this is a killing action. Your suggestion looks like an innocent replay, even more innocent than red Reanalyze action.

devoncarew:
The skull icon has always seemed a bit alarming to me as an icon, even if it does accurately convey the fact that you're terminating some process or task.

I do agree that the terminate-and-restart icon I have above could be improved. Perhaps if the stop/square part of the icon were larger and red (with changes to the restart arrow to compensate).

@lukechurch will be in town in a week; perhaps we can discuss with him and revisit this PR then?

devoncarew:
We could alternatively decide that we no longer need the kill-analysis-server option - that the re-analyze sources button (and the ability to restart IntelliJ) serves our needs enough.

alexander-doroshko:
> discuss with him and revisit this PR then?

Sure

> We could alternatively decide that we no longer need the kill-analysis-server option

I remember I had cases when Reanalyze didn't help, but restart did. Unfortunately do not remember STR. May be the problem has gone. 
We may also consider medium decision: remove Kill icon from the tool window and keep it discoverable via Find Action only.

devoncarew:
> We may also consider medium decision: remove Kill icon from the tool window and keep it discoverable via Find Action only.

I didn't know this was possible; I like that idea. I'll update this PR (or start a new one w/o the icon changes).

alexander-doroshko:
It's a one-line change, I'll do it myself if everyone agrees.

@scheglov Do you think Analysis Server is stable enough to hide Restart action from the tool window (keep Restart discoverable via Find Action only; Reanalyze action remains in the tool window)?

jwren:
+1 for removing.  It will still be accessible in the actions menu.  If we are wrong and people complain, it is also a one-line change to put back in.

scheglov:
Honestly, I don't know - we don't have statistics about how often it is used, do we?
We can try to remove it.

devoncarew:
@alexander-doroshko - closing this PR. I have a follow-up one w/ just the minor tweak to the grouping. It does sound like people are ok with trying it w/o the restart action in the view.

alexander-doroshko:
Ok, removed Restart action from the tool window.

PR #421, opened 2016-09-16
have the content enabled checkbox selected

devoncarew:
- minor tweak to have the generate sample content checkbox selected by default

@alexander-doroshko @pq 

screen shot 2016-09-16 at 8 20 53 am

PR #422, opened 2016-09-17
enable the template select box the first time through

devoncarew:
Follow up to https://github.com/JetBrains/intellij-plugins/pull/421, enable the template select box the first time through. The first time through there would be no value for the previous dart template setting, which would cause the checkbox to not be selected.

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/projectWizard/DartGeneratorPeer.java
+++ b/Dart/src/com/jetbrains/lang/dart/projectWizard/DartGeneratorPeer.java
@@ -172,8 +172,9 @@ public class DartGeneratorPeer implements WebProjectGenerator.GeneratorPeer<Dart
     myCreateSampleProjectCheckBox.setEnabled(true);
     myTemplatesList.setEnabled(true);
 
-    final String selectedTemplateName = PropertiesComponent.getInstance().getValue(DART_PROJECT_TEMPLATE);
-    myCreateSampleProjectCheckBox.setSelected(selectedTemplateName != null);
+    final PropertiesComponent properties = PropertiesComponent.getInstance();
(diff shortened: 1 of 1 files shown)


Dart/src/com/jetbrains/lang/dart/projectWizard/DartGeneratorPeer.java
```diff
@@ -172,8 +172,9 @@ private void onTemplatesLoaded(final List<DartProjectTemplate> templates) {
     myCreateSampleProjectCheckBox.setEnabled(true);
     myTemplatesList.setEnabled(true);
 
-    final String selectedTemplateName = PropertiesComponent.getInstance().getValue(DART_PROJECT_TEMPLATE);
-    myCreateSampleProjectCheckBox.setSelected(selectedTemplateName != null);
+    final PropertiesComponent properties = PropertiesComponent.getInstance();
+    final String selectedTemplateName = properties.getValue(DART_PROJECT_TEMPLATE);
+    myCreateSampleProjectCheckBox.setSelected(selectedTemplateName != null || !properties.isValueSet(DART_PROJECT_TEMPLATE));
```
alexander-doroshko: It's equivalent to just `setSelected(true)`. This way we enable check box even if it was previously disabled by user.
Do you agree it should be disabled if user has manually disabled it at previous project creation? 
I have fix in hands, just wanted to ask your opinion before pushing.

devoncarew:
> Do you agree it should be disabled if user has manually disabled it at previous project creation? 

No :) My intent here - w/o being that familiar w/ the framework - is for the checkbox to be selected the first time through the dialog, or if it was selected the last time the dialog was open.

devoncarew:
Cool, sounds like we're in agreement :)

> I have fix in hands, just wanted to ask your opinion before pushing.

I'll close this PR out then -

PR #423, opened 2016-09-19
Remote debugging

devoncarew:
Fix several issues in the Dart remote debugging launch config:
- when remote debugging an app that does not use a pubspec.yaml file, resolve some urls relative to the source location given in the remote debug dialog
- when remote debugging, register for stdout and stderr events from the vm (WriteEvents) and display them to the output console
- `getUrisForFile()` now returns a few more types, including `package:` (if relevant), `file:`, remote prefixed paths, and straight file paths (`/path/to/file`); straight file paths are used for some Flutter VM embedders
- handle a race condition between setting breakpoints and resuming. We now resume an isolate once all breakpoints have been set on it, and we've received the `StartPaused` event (the logic is tracked in the `IsolateInfo` class and the `checkInitialResume()` method)
- register all isolates when we first connect, not just one (there can be more than one when connecting to an already running VM)
- handle the case where we connect to an already running VM and there are existing paused isolates (say, paused at breakpoints)
- fix https://youtrack.jetbrains.com/issue/WEB-23266, fix https://youtrack.jetbrains.com/issue/WEB-23278

@alexander-doroshko, this is not a huge PR, but does address several (related) issues. Would definitely appreciate feedback on it! I may or may not have some follow-up to this after some additional testing Monday.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
@@ -1,5 +1,6 @@
 package com.jetbrains.lang.dart.ide.runner.server.vmService;
 
+import com.google.common.base.Charsets;
 import com.intellij.execution.ExecutionResult;
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.execution.ui.ConsoleViewContentType;
@@ -383,6 +384,11 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
     if (isolateRef.getId().equals(myLatestCurrentIsolateId)) {
       resume(getSession().getSuspendContext()); // otherwise no way no resume them from UI
     }
+  }
+
+  public void handleWriteEvent(String base64Data) {
+    String message = new String(Base64.getDecoder().decode(base64Data), Charsets.UTF_8);
+    getSession().getConsoleView().print(message, ConsoleViewContentType.NORMAL_OUTPUT);
   }
 
   @Override
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
@@ -35,7 +35,6 @@ import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceSt
 import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceSuspendContext;
 import com.jetbrains.lang.dart.util.DartResolveUtil;
 import com.jetbrains.lang.dart.util.DartUrlResolver;
-import com.jetbrains.lang.dart.util.PubspecYamlUtil;
 import gnu.trove.THashMap;
 import gnu.trove.THashSet;
 import gnu.trove.TIntObjectHashMap;
@@ -147,7 +146,6 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
           return;
         }
 
-        getSession().getConsoleView().print("Error: " + message + "\n", ConsoleViewContentType.ERROR_OUTPUT);
         LOG.warn(message);
       }
 
@@ -260,7 +258,9 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
 
     for (LibraryRef library : libraries) {
       final String remoteUri = library.getUri();
-      if (!remoteUri.startsWith(DartUrlResolver.FILE_PREFIX)) continue;
+      //if (!remoteUri.startsWith(DartUrlResolver.FILE_PREFIX)) continue;
+      if (remoteUri.startsWith("dart:")) continue;
+      if (remoteUri.startsWith("package:")) continue;
 
       final PsiFile[] localFilesWithSameName = ApplicationManager.getApplication().runReadAction(new Computable<PsiFile[]>() {
         @Override
@@ -279,13 +279,13 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
 
         LOG.assertTrue(file.getPath().startsWith(localProjectRoot.getPath() + "/"), file.getPath() + "," + localProjectRoot.getPath());
         final String relPath = file.getPath().substring(localProjectRoot.getPath().length()); // starts with slash
-        if (!relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
+        if (relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
           howManyFilesMatch++;
           myRemoteProjectRootUri = remoteUri.substring(0, remoteUri.length() - relPath.length());
         }
       }
 
-      if (howManyFilesMatch == 1) {
+      if (howManyFilesMatch > 0) {
         break; // we did the best guess we could
       }
     }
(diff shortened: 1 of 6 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -279,13 +286,13 @@ public void guessRemoteProjectRoot(@NotNull final ElementList<LibraryRef> librar
 
         LOG.assertTrue(file.getPath().startsWith(localProjectRoot.getPath() + "/"), file.getPath() + "," + localProjectRoot.getPath());
         final String relPath = file.getPath().substring(localProjectRoot.getPath().length()); // starts with slash
-        if (!relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
+        if (relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
```
alexander-doroshko: I'm afraid I do not understand this change. For standard project this condition is always `false` because files from 'lib' folder do not have 'lib' in uri, they have 'package:' uri and can't help us to guess remote project root. 
My idea was to catch some file from, for example, 'bin' folder having URI=file:///remote/path/to/project/bin/foo/bar/file.dart, find local file in /local/path/to/project/bin/foo/bar/file.dart and make a conclusion that remote root is /remote/path/to/project.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -74,6 +74,7 @@
   private final boolean myRemoteDebug;
   private final boolean myEntryPointInLibFolder;
   private final int myTimeout;
+  private final VirtualFile myCurrentWorkingDirectory;
```
devoncarew: Cool, I updated this to `@Nullable` (and adjusted one use of myCurrentWorkingDirectory).

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -147,7 +150,6 @@ public void logError(final String message) {
           return;
         }
 
-        getSession().getConsoleView().print("Error: " + message + "\n", ConsoleViewContentType.ERROR_OUTPUT);
```
devoncarew: I can remove it below as well - this was displaying some API usage error information in the user's console. I thought it would be better to only log it is `LOG.warn()` / `LOG.error()`. Let me know which you'd prefer -

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -279,13 +286,13 @@ public void guessRemoteProjectRoot(@NotNull final ElementList<LibraryRef> librar
...
         final String relPath = file.getPath().substring(localProjectRoot.getPath().length()); // starts with slash
-        if (!relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
+        if (relPath.startsWith("/lib/") && remoteUri.endsWith(relPath)) {
           howManyFilesMatch++;
           myRemoteProjectRootUri = remoteUri.substring(0, remoteUri.length() - relPath.length());
         }
       }
 
-      if (howManyFilesMatch == 1) {
+      if (howManyFilesMatch > 0) {
```
devoncarew: Shouldn't we be looking for `bar/main.dart` in that case (i.e., something relative to the local project root)? I worry that we'll reject a match and come up with no value for `myRemoteProjectRootUri`.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
devoncarew: Flutter apps generally have an entry-point of `lib/main.dart`, so not all (any?) of their files in `lib/` will be loaded via self-references.

Let me revisit the logic here. I think currently it only works if some files are loaded from `lib/`.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
alexander-doroshko: It depends on what kind of garbage can go here and how severe there warnings are.
Main point is that none of the predictable messages should go here. Only indicators of some unexpected problems.
On one hand the more visible is the message the more chances that user will report it to us. OTOH we do not want to bother user too much especially if functionality doesn't break because of it.
Truly speaking, LOG.warn() has little chances to reach developers. LOG.error has ore chances because errors have indication in EAPs.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
alexander-doroshko: > I worry that we'll reject a match and come up with no value for myRemoteProjectRootUri

We always assign myRemoteProjectRootUri to something and here we decide whether to continue heuristics. If we are not sure our guess is the best one (howManyFilesMatch > 1) then we try to guess better with another file name.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
alexander-doroshko: then probably `if (remoteUri.endsWith(relPath))` is enough

devoncarew: I removed the check for `lib/` completely - I think this is necessary to work w/ apps that can have entrypoints in lib/.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
devoncarew: Makes sense. I saw this frequently when working on this patch, but don't know if its something that users will encounter as much.

I can change the warn() to an error(), or add the print statement back in.

(re-tested: it looks like I still see it print out quite a bit: 

```
Error Response: {"code":-32602,"message":"Invalid params","data":{"request":{"method":"removeBreakpoint","params":{"isolateId":"isolates/47369335","breakpointId":"breakpoints/7"}},"details":"removeBreakpoint: invalid 'breakpointId' parameter: breakpoints/7"}}
```
)

devoncarew: update:

- I restored the print
- changed the warn to an error
- and adjusted the filter (to ignore the warning) to better match the removeBreakpoint call, which is what was triggering this

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
devoncarew: (I changed this back to the original logic, and added a TODO: to look for additional ways to match if howManyFilesMatch > 1)

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -279,14 +287,16 @@ public void guessRemoteProjectRoot(@NotNull final ElementList<LibraryRef> librar
...
+        if (remoteUri.endsWith(relPath)) {
           howManyFilesMatch++;
           myRemoteProjectRootUri = remoteUri.substring(0, remoteUri.length() - relPath.length());
         }
       }
 
       if (howManyFilesMatch == 1) {
         break; // we did the best guess we could
+      } else if (howManyFilesMatch > 0) {
+        // TODO: Look for an additional way to determine `myRemoteProjectRootUri`.
```
alexander-doroshko: I've removed TODO, if we do not break on `howManyFilesMatch == 1` then we try to improve our guess on next iteration (we are inside of a loop here).

PR #424, opened 2016-09-20
Fix IDEA-98836

fwienber:
"Good code red when using non top-level package resource bundle"
https://youtrack.jetbrains.com/issue/IDEA-98836

To resolve package name from relative file position, reused existing
utility method JSResolveUtil.getExpectedPackageNameFromFile().

Sorry that there are no tests, but there haven't been any tests for
FlexPropertySupport before so I don't really know where to start.

### DIFF
--- a/flex/flex-tests/testData/flex_highlighting/ResourceReferencesWithPackages/ResourceReferencesWithPackages.mxml
+++ b/flex/flex-tests/testData/flex_highlighting/ResourceReferencesWithPackages/ResourceReferencesWithPackages.mxml
@@ -1,0 +1,15 @@
+<?xml version="1.0" ?>
+<mx:Application xmlns:mx="http://www.adobe.com/2006/mxml">
+  <mx:Metadata>
+    [ResourceBundle("resourcepackage.mybundle4")]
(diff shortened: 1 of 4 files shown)


alexander-doroshko:
Thank you for your contribution!
Take a look at `FlexHighlightingTest.testResourceReferences2()`, you may update this test or add a similar one.

fwienber:
Thanks for the code pointer! With all the examples of testing for green/red Flex code, it was easy to add a test (see new commit) that is red with the previous code and green with my fix.
Works for you?

alexander-doroshko:
Yep, applying the PR now, thanks!

One question though. In the test you invoke quick fix, but it does nothing in this case. Is it needed? If not I'll remove last 3 lines of the test.

fwienber:
Probably not, I was asking that myself, too. I copied these lines from a similar test and thought they might be needed to trigger error-checking, but if this is not the case, then they can of course be removed.

Btw., one thing that could be improved is that completion of resource bundle names does not work as I expected with my solution. Completion suggestions are correctly found for the fully-qualified name, but it would of course also be nice if you could type the non-qualified name and IDEA would complement the package (like when referencing normal classes).

PR #425, opened 2016-09-26
Support using constants for resource bundle names.

fwienber:
So far, IDEA only supports resource bundle references when given directly as string literals.
This changes additionally allows variables with a string literal initializer to be used.
This makes code more readable when (especially fully-qualified) resource bundle names are rather long.

Example:

    var foo:String = resourceManager.getString("my.very.long.package.MyBundle", "foo");
    var bar:String = resourceManager.getString("my.very.long.package.MyBundle", "bar");

is now also fully supported when introducing a constant:

    private static const MY_BUNDLE:String = "my.very.long.package.MyBundle";
    
    var foo:String = resourceManager.getString(MY_BUNDLE, "foo");
    var bar:String = resourceManager.getString(MY_BUNDLE, "bar");

Before, the second code would suggest / accept any key defined in any resource bundle,
not only the keys of `MyBundle`.

The adapted test (`ResourceReferences2.mxml`) proves that: The incorrect access to
key `prompt4` was not detected without the change to `FlexPropertyReferenceProvider`
(`FlexHighlightingTest#testResourceReferences2()` red), but after applying the changes
to `FlexPropertyReferenceProvider`, the test is green again.

### DIFF
--- a/flex/flex-tests/testData/flex_highlighting/ResourceReferences2.mxml
+++ b/flex/flex-tests/testData/flex_highlighting/ResourceReferences2.mxml
@@ -25,7 +25,12 @@
     resourceManager.getResourceBundle('ru', '<error>mybundle2</error>')
     resourceManager.findResourceBundleWithResource('<error>mybundle2</error>', '<error>prompt1</error>')
     resourceManager.findResourceBundleWithResource('mybundle3', '<error>prompt2</error>')
-  }
+
+    resourceManager.getString('mybundle', '<error>prompt4</error>')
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
@fwienber  Thanks, accepted!
I also added positive test (you added negative only) and renamed local var (local vars and args should not have 'my' prefix, this prefix is for instance fields only).

fwienber:
Thanks for accepting (and improving!) this change!
Any chance this will arrive in 2016.3?

alexander-doroshko:
Yes, 2016.3 release branch is not created yet, everything from master branch goes there

PR #431, opened 2016-10-17
Add FLUTTER_ROOT env info to Flutter module Dart Pub actions.

pq:
Checks if a module is a Flutter module and sets the `pub` environment variable appropriately.  This is required for `pub` to work in the context of Flutter.

Fixes: https://github.com/flutter/flutter-intellij/issues/271

/cc @alexander-doroshko @devoncarew

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
@@ -1,18 +1,3 @@
-/*
- * Copyright 2000-2016 JetBrains s.r.o.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 package com.jetbrains.lang.dart.flutter;
 
 import com.intellij.openapi.module.Module;
--- a/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
@@ -4,14 +4,11 @@ import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleType;
 import com.jetbrains.lang.dart.sdk.DartSdk;
 import org.jetbrains.annotations.NotNull;
-
-import java.io.File;
-import java.net.URI;
-import java.net.URISyntaxException;
+import org.jetbrains.annotations.Nullable;
 
 public class FlutterUtil {
 
-  private static final String FLUTTER_MODULE_TYPE_NAME = "Flutter";
+  private static final String FLUTTER_MODULE_TYPE_ID = "FLUTTER_MODULE_TYPE";
 
   /**
    * Test if the given module is a Flutter module.
@@ -21,7 +18,7 @@ public class FlutterUtil {
    */
   public static boolean isFlutterModule(@NotNull Module module) {
     final ModuleType moduleType = ModuleType.get(module);
-    return FLUTTER_MODULE_TYPE_NAME.equals(moduleType.getName());
+    return FLUTTER_MODULE_TYPE_ID.equals(moduleType.getId());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Test if the given module is a Flutter module.
+   *
+   * @param module the module to test
+   * @return true if the module is a Flutter module, false otherwise
+   */
+  public static boolean isFlutterModule(@NotNull Module module) {
+    final ModuleType moduleType = ModuleType.get(module);
+    return FLUTTER_MODULE_TYPE_NAME.equals(moduleType.getName());
```
alexander-doroshko: I think it's better to check module type id. Module type name can be safely changed (for example someone may decide to rename 'Flutter' to 'Flutter Module' as all other module types have 'Module' word), but module type id is less likely to be ever changed as it is a breaking change.

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Get the Flutter root relative to the given Dart SDK.
+   *
+   * @param dartSdk the Dart SDK
+   * @return the relative Flutter root, or null if this SDK is not relative to one
+   */
+  public static String getFlutterRoot(@NotNull DartSdk dartSdk) {
+    // Navigate up from `bin/cache/dart-sdk/lib/`.
+    final File flutterRoot = new File(dartSdk.getHomePath() + "/../../..");
+    if (flutterRoot.exists()) {
```
alexander-doroshko: This seems to work not as we expect. If `dartSdk.getHomePath()` is less than 3 levels deep to the file system root then with "/../../.." or with "/../../../../../../../../../../../../../../.." it returns the same: FS root.

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -0,0 +1,47 @@
...
+      try {
+        final URI uri = new URI(flutterRoot.getAbsolutePath()).normalize();
```
alexander-doroshko: This fails on Windows because absolute path contains back slashes

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -0,0 +1,47 @@
...
+        return uri.getPath();
+      }
+      catch (URISyntaxException e) {
+        // Ignored
+      }
+    }
+    return null;
```
alexander-doroshko: How about simplifying the whole method to these 2 lines
```
    final String flutterPath = PathUtil.getParentPath(PathUtil.getParentPath(PathUtil.getParentPath(dartSdk.getHomePath())));
    return flutterPath.isEmpty() ? null : flutterPath;
```
?

Or, even better, with Flutter-specific checks:
```
    final String dartPath = dartSdk.getHomePath();
    final String suffix = "/bin/cache/dart-sdk";
    return dartPath.endsWith(suffix) ? dartPath.substring(0, dartPath.length() - suffix.length()) : null;
```
?

Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
```diff
@@ -147,6 +148,14 @@ public void performPubAction(final @NotNull Module module, final @NotNull Virtua
 
     if (pubParameters != null) {
       final GeneralCommandLine command = new GeneralCommandLine().withWorkDirectory(pubspecYamlFile.getParent().getPath());
+
+      if (FlutterUtil.isFlutterModule(module)) {
+        final String flutterRoot = FlutterUtil.getFlutterRoot(sdk);
+        if (flutterRoot != null) {
+          command.getEnvironment().put("FLUTTER_ROOT", flutterRoot);
```
alexander-doroshko: If Flutter be ever supported on Windows it may expect native slashes in env vars. It may be safer to wrap path with `FileUtil.toSystemDependentName(flutterRoot)`.

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    return FLUTTER_MODULE_TYPE_NAME.equals(moduleType.getName());
+  }
+
+  /**
+   * Get the Flutter root relative to the given Dart SDK.
+   *
+   * @param dartSdk the Dart SDK
+   * @return the relative Flutter root, or null if this SDK is not relative to one
+   */
+  public static String getFlutterRoot(@NotNull DartSdk dartSdk) {
```
alexander-doroshko: Please annotate with `@Nullable`.

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
pq: Ah, good thinking.  Done.

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
pq: Oh.  You're right.  (Will fix as you suggest below.)

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
pq: (Will address below.)

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
pq: Great.  Advice taken!

Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
pq: Good thinking.  Done.

pq:
Thanks for the feedback!  Comments addressed.

PR #435, opened 2016-11-04
Update dart_style tests

stevemessick:
@alexander-doroshko Update dart_style tests to be consistent with current version.

As before, I updated the `testData` files from the dart_style repo, then added entries to ignore new failures. Lots of new failures related to trailing commas, as expected.

Please ignore the bogus commit messages.

/cc @jwren @devoncarew


alexander-doroshko:
Thanks! 
Merged manually (to avoid a big list of empty commits)

PR #437, opened 2016-11-11
Restart if needed

stevemessick:
@alexander-doroshko Brian mentioned that this might be the default behavior and, to me, it looks like a bug in the current implementation.

When `Reanalyze Sources` is clicked the server will be restarted if it is not currently running.

Once we implement the logic to automatically restart the server when it crashes this will be revised to work with that.

cc @bwilkerson

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -1386,6 +1386,15 @@ public class DartAnalysisServerService {
     }
   }
 
+  public boolean isServerProcessActive() {
+    synchronized (myLock) {
+      if (myServer == null || !myServer.isSocketOpen()) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   public boolean serverReadyForRequest(@NotNull final Project project) {
     final DartSdk sdk = DartSdk.getDartSdk(project);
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
How can server be not alive, but Reanalyze action visible?
I know only one way - if user has killed server process manually outside of the IDE. Are you covering this case or there are other scenarios?

P.S. Your PRs contain useless commits in addition to the main one. Unfortunately I do not have 'Squash and merge' button, as I'm merging it not to the github repo. I have to merge such PRs manually in order to skip useless commits.

stevemessick:
This is just good defensive programming. I don't know how many ways the server can crash, neither does the server team.

I spent several hours trying to rid my clone of the useless commits. This is the best I've been able to come up with. I'm happy to do what is needed, but the only thing I can think of is to clone the repo again. I was hoping they'd go away if you committed the previous PR, which only had two useless commits. If I have to rebuild then I'll do it sometime next week.

alexander-doroshko:
No worries, will merge manually.
I don't know, what's up on your end. May be you have to rewind to the state before these 2 commits appeared and then pull again from the base repo. Even if you go from scratch, Dart plugin rebuild is just 14 seconds for me, though cloning intellij-plugins is a bit longer.

alexander-doroshko:
Merged manually.
Good catch that the list of tracked projects may be lost! Added the same to `RestartDartAnalysisServerAction`.

PR #438, opened 2016-11-14
parse the PausePostRequest event

devoncarew:
Add support for the new `PausePostRequest` vm event type. This is sent by the VM after it has processed a hot reload call, when that reload call was invoked with the `pause` option, to pause execution of the isolate after the reload. This gives clients the ability to restore breakpoints before the program resumes.

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
@@ -78,6 +78,11 @@ public class DartVmServiceListener implements VmServiceListener {
           onIsolatePaused(event.getIsolate(), breakpoints, exception, event.getTopFrame(), event.getAtAsyncSuspension());
         });
         break;
+      case PausePostRequest:
+        // We get this event after an isolate reload call, when pause after reload has been requested.
+        // TODO: Set current breakpoints; resume the isolate.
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
```diff
@@ -78,6 +78,11 @@ public void received(@NotNull final String streamId, @NotNull final Event event)
           onIsolatePaused(event.getIsolate(), breakpoints, exception, event.getTopFrame(), event.getAtAsyncSuspension());
         });
         break;
+      case PausePostRequest:
+        // We get this event after an isolate reload call, when pause after reload has been requested.
+        // TODO: Set current breakpoints; resume the isolate.
+        myDebugProcess.resumeIsolate(event.getIsolate().getId(), null);
```
alexander-doroshko: I'll replace with `myDebugProcess.getVmServiceWrapper().resumeIsolate(isolateRef.getId(), null)` to avoid introducing a delegating method in `DartVmServiceDebugProcess` class.

alexander-doroshko:
👍

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
devoncarew: cool, makes sense

PR #439, opened 2016-11-14
Dart plugin- Disable super mixins when configuring the Dart Analysis Server

jwren:


alexander-doroshko:
What's wrong with super mixins?

devoncarew:
There's a discussion about how best to support them. They're currently available in the analyzer and the VM but not in dart2js, and clients like Flutter are using them. If we enable for all users, they will fail at compile time in dart2js. Users do have the recourse of enabling via a flag in `.analysis_options`.

I would hold off on landing this PR for a bit until it's more clear what to do -

PR #440, opened 2016-11-17
add a DartPluginCapabilities.java class

devoncarew:
- per email, add a DartPluginCapabilities.java class

I'd like to get feedback on the API (and land this), before I start using the API from the Flutter plugin. Thanks!

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartPluginCapabilities.java
+++ b/Dart/src/com/jetbrains/lang/dart/DartPluginCapabilities.java
@@ -1,21 +1,21 @@
 package com.jetbrains.lang.dart;
 
-import com.intellij.util.containers.HashMap;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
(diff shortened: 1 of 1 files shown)


Dart/src/com/jetbrains/lang/dart/DartPluginCapabilities.java
```diff
@@ -0,0 +1,21 @@
...
+import com.intellij.util.containers.HashMap;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Map;
+
+/**
+ * A queryable class about the capabilities of the Dart plugin.
+ */
+public class DartPluginCapabilities {
+  public static Map<String, Boolean> capabilities = new HashMap<>();
```
alexander-doroshko: may be make it `private`?

Also, may be a `Set` is enough? Set contains -> feature supported, doesn't contain -> not supported.

devoncarew: Makes sense! Will update

devoncarew:
updated

PR #441, opened 2016-11-19
Add a button to Dart problems view that submits feedback

stevemessick:
@alexander-doroshko This is not finished but is ready to commit for the next release. I'll be working to fix known problems in the next week.

The "yellow" button displays too often. I need to adjust the timing for it.

The feedback form needs much more info about what the analysis server is doing. I'm working with that team to determine what's going to be added.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/AnalysisServerStatusAction.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/AnalysisServerStatusAction.java
@@ -1,0 +1,95 @@
+package com.jetbrains.lang.dart.ide.errorTreeView;
+
+import com.intellij.icons.AllIcons;
+import com.intellij.ide.actions.SendFeedbackAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.Presentation;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.extensions.Extensions;
+import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.MessageDialogBuilder;
+import com.intellij.openapi.ui.Messages;
+import com.intellij.openapi.wm.ToolWindowManager;
+import com.jetbrains.lang.dart.DartBundle;
+import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+
+public class AnalysisServerStatusAction extends DumbAwareAction {
+  private static final Icon STATUS_GOOD = AllIcons.Process.State.GreenOK;
+  private static final Icon STATUS_UNKNOWN = AllIcons.Process.State.YellowStr;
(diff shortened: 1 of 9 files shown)


alexander-doroshko:
@stevemessick Are you sure there will be a different implementation of the `FeedbackBuilder` in the Flutter plugin?
Although Dart Analysis Server has some knowledge about Flutter, I thought all its issues are handled in the https://github.com/dart-lang/sdk/issues tracker.
Also it wouldn't be great if just the presence of the Flutter plugin changes behavior of the Dart plugin when working on non-Flutter projects.
I'm not against extension points. My point is to add them lazily, i.e. only when needed and not too long before corresponding implementation appears in the Flutter pulgin code base.

stevemessick:
I sent you email about this.

I'll bring in @pq to make sure the concern about changing behavior for Dart projects is addressed, too.

devoncarew:
@alexander-doroshko, good point about not designing too much up-front. In this case, we will want to be able to report internal issues to our internal bug tracker. It'll let us report a lot more data in those cases. /cc @jwren

alexander-doroshko:
Don't you mind if I rename `FeedbackBuilder` and `DartFeedbackBuilder` to `DartFeedbackBuilder` and `DartDefaultFeedbackBuilder` correspondingly? Having 200+ plugins and 600+ modules in a single project I love colleagues who use [Smurf Naming Convention](https://blog.codinghorror.com/new-programming-jargon/) within each plugin. It helps greatly when analyzing search results across the whole code base and in many other cases.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/AnalysisServerStatusAction.java
```diff
@@ -0,0 +1,94 @@
...
+import com.intellij.openapi.ui.MessageDialogBuilder;
+import com.intellij.openapi.ui.Messages;
+import com.intellij.openapi.wm.ToolWindowManager;
+import com.jetbrains.lang.dart.DartBundle;
+import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+
+public class AnalysisServerStatusAction extends SendFeedbackAction {
```
alexander-doroshko: Looks like this class doesn't use the fact that it extends `SendFeedbackAction`. All meaningful methods are overridden and have no `super()` calls. How about extending `DumbAwareAction` directly?

pq:
> I love colleagues who use Smurf Naming Convention 

I needed a good laugh today.  Thanks for delivering!

stevemessick:
I renamed two classes, changed one superclass, and changed one bundle message.

stevemessick:
@alexander-doroshko PTAL

alexander-doroshko:
@stevemessick I'm about to land it.
Just trying to simplify it a bit. I'm going to remove `Alarm` and the whole `startMonitor()` method together with `AnalysisServerStatusAction#myPresentation` field and tricks with `Disposable`. The Platform itself will call `AnalysisServerStatusAction#update()` frequently enough (and only when the button is visible) and in its `update()` implementation the action will manage its icon and text. This is a standard way of managing all action states.
I'll paste a commit url to review when ready.

PR #443, opened 2016-11-28
ui tweaks to the problems view

devoncarew:
Various tweaks to the problems view:

- change the tooltip for the status/feedback button to make it clear what plugin the feedback is for and to separate out a bit the feedback portion of the tooltip from the analysis server status portion
- allow implementations of `DartFeedbackBuilder` to require (or not) a dialog before feedback is sent. In the case of the default feedback provider, don't require a dialog - the user has clicked on a feedback button, and has to explicitly submit the report in the github UI.
- remove the `reanalyze` button, and only use the `restart` button. They're pretty similar and I don't think we need the distinction
- change the restart button icon. Two good options I found:

AllIcons.Actions.Restart:

screen shot 2016-11-28 at 10 44 24 am

AllIcons.Actions.ForceRefresh:

screen shot 2016-11-28 at 10 47 19 am

@stevemessick, @alexander-doroshko, comments on this much appreciated!

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -260,11 +260,11 @@ rename.need.reRun=Cannot rename.\nThere were changes in code after usages have b
 
 exportable.DartCodeFoldingSettings.presentable.name=Code Folding Settings
 
-analysis.server.status.good.text=Send feedback
+analysis.server.status.good.text=Dart Plugin: Submit Feedback
 analysis.server.status.good.desc=Analysis server is operating normally
-analysis.server.status.unknown.text=Send feedback about busy server
+analysis.server.status.unknown.text=Dart Plugin: Submit Feedback (server busy)
 analysis.server.status.unknown.desc=Analysis server is running but has been busy a long time
-analysis.server.status.bad.text=Submit an issue about the server instability
+analysis.server.status.bad.text=Dart Plugin: Submit Feedback (analysis process not active)
 analysis.server.status.bad.desc=Analysis server needs to be restarted
 
 dart.feedback.url.template=https://github.com/dart-lang/sdk/issues/new?body=Analyzer Feedback from IntelliJ\n\
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -260,11 +260,11 @@ rename.need.reRun=Cannot rename.\nThere were changes in code after usages have b
 
 exportable.DartCodeFoldingSettings.presentable.name=Code Folding Settings
 
-analysis.server.status.good.text=Dart Plugin: Submit Feedback
+analysis.server.status.good.text=Analyzer: Submit Feedback
 analysis.server.status.good.desc=Analysis server is operating normally
-analysis.server.status.unknown.text=Dart Plugin: Submit Feedback (server busy)
+analysis.server.status.unknown.text=Analyzer: Submit Feedback (server busy)
 analysis.server.status.unknown.desc=Analysis server is running but has been busy a long time
-analysis.server.status.bad.text=Dart Plugin: Submit Feedback (analysis process not active)
+analysis.server.status.bad.text=Analyzer: Submit Feedback (analysis process not active)
 analysis.server.status.bad.desc=Analysis server needs to be restarted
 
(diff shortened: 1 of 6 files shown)


Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/AnalysisServerStatusAction.java
```diff
@@ -28,12 +28,16 @@ public AnalysisServerStatusAction() {
   public void actionPerformed(AnActionEvent e) {
     DartFeedbackBuilder builder = DartFeedbackBuilder.getFeedbackBuilder();
 
-    if (MessageDialogBuilder.yesNo(builder.title(), builder.prompt())
-          .icon(Messages.getQuestionIcon())
-          .yesText(builder.label())
-          .show() == Messages.YES) {
-      builder.sendFeedback(e.getProject());
+    if (builder.isPromptRequired()) {
```
stevemessick: This seems like a premature optimization of the UX to me. I think many people do not pay attention to the status line and if they have not seen the message then clicking this button is going to have a very surprising effect.

I'm moving the MessageDialogBuilder construction into the DartFeedbackBuilder.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
```diff
@@ -175,11 +179,6 @@ private void updateStatusBar() {
   }
 
   private static void addReanalyzeActions(@NotNull final DefaultActionGroup group) {
-    final AnAction reanalyzeAction = ActionManager.getInstance().getAction("Dart.Reanalyze");
```
stevemessick: Removing this means that if multiple projects are open they all have to be reanalyzed when any one of them seems to be out of sync. Is that what you want?

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/AnalysisServerStatusAction.java
stevemessick: Forgot to mention that I agree with your concern here. I was thinking of adding a "Don't show this again" checkbox. But that means adding a preference and requires some code I hadn't gotten around to writing yet.

devoncarew: sgtm; I'll back this part out.

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
```diff
@@ -260,11 +260,11 @@ rename.need.reRun=Cannot rename.\nThere were changes in code after usages have b
 
 exportable.DartCodeFoldingSettings.presentable.name=Code Folding Settings
 
-analysis.server.status.good.text=Send feedback
+analysis.server.status.good.text=Dart Plugin: Submit Feedback
```
alexander-doroshko: Strictly speaking, the button is about Dart Analysis Server feedback, not about the Plugin. Maybe we can find more appropriate words?

devoncarew: `Analyzer: Submit Feedback`?

`Dart Analyzer: Submit Feedback`?

More descriptive is good; I think it would be good to keep it terse as well.

alexander-doroshko:
@devoncarew wrote:

> remove the reanalyze button, and only use the restart button. They're pretty similar and I don't think we need the distinction

I'd like to loop @bwilkerson and @scheglov here. Currently we have both 'Reanalyze' button (that calls the corresponding method of the Analysis Server) and 'Restart' that shuts down (or kills) the server process and starts from scratch. Devon is suggesting to leave 'Restart' button only. Are you ok with it?

scheglov:
Yes, I agree that distinction is not big enough.
Let's keep only restart.

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
alexander-doroshko: Both lgtm. This is inside of a Dart Analysis tool window, so probably it's ok to skip one more 'Dart' word. But I'll accept any.

bwilkerson:
The hope is that we won't need either very often, in which case having both is overkill. I'm not entirely convinced we're there today, but we can always add it back in if it hurts the UX to have removed it.

alexander-doroshko:
Merged manually. I didn't commit TODOs, will address them separately.

PR #444, opened 2016-11-29
Prompt user to report analysis server errors

stevemessick:
@alexander-doroshko When an error in the analysis server is reported to IntelliJ we now prompt the user to submit feedback to github.

Testing was difficult. I had to make a repository that causes the analyzer to throw an error. I got it to do the same error twice, quickly, and tuned the code to filter out too many disruptions. Then added a counter heuristic to keep from being totally obnoxious about asking the user to report errors.

Stack traces can be quite long. For those that do not fit within IE's (current) URL length limit we dump the trace to a file and ask the user to insert the content of the file into the issue. To make that easy, the file is reported as a URL that can be opened in a browser.

@devoncarew This does not have the "Do not ask again" checkbox that I plan to add to the query, but I haven't forgotten about it.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -267,8 +267,10 @@ analysis.server.status.unknown.desc=Analysis server is running but has been busy
 analysis.server.status.bad.text=Analyzer: Submit Feedback (analysis process not active)
 analysis.server.status.bad.desc=Analysis server needs to be restarted
 
-dart.feedback.url.template=https://github.com/dart-lang/sdk/issues/new?body=Analyzer Feedback from IntelliJ\n\
-  IDEA: {0} Dart: {1} {2}
+dart.feedback.url.template=https://github.com/dart-lang/sdk/issues/new?body=# Analyzer Feedback from IntelliJ\n\
+  IDEA: {0} Dart: {1}\n\
+  ## Configuration{2}\n\
+  ## Stack
 dart.analysis.server.error=Dart analysis server error
 dart.error.file.instructions=Please insert the content of this file here:\n\
   file://{0}
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -269,3 +269,6 @@ analysis.server.status.bad.desc=Analysis server needs to be restarted
 
 dart.feedback.url.template=https://github.com/dart-lang/sdk/issues/new?body=Analyzer Feedback from IntelliJ\n\
   IDEA: {0} Dart: {1} {2}
+dart.analysis.server.error=Dart analysis server error
+dart.error.file.instructions=Please insert the content of this file here:\n\
+  file://{0}
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
@@ -29,9 +29,10 @@ public class DefaultDartFeedbackBuilder implements DartFeedbackBuilder {
     boolean eap = appInfo.isEAP();
     String ijBuild = eap ? appInfo.getBuild().asStringWithoutProductCode() : appInfo.getBuild().asString();
     String sdkVsn = getSdkVersion(project);
-    String platDescr = getDescription();
+    String platDescr = getDescription().replaceAll(";", " ");
     String template = DartBundle.message("dart.feedback.url.template", ijBuild, sdkVsn, platDescr);
     if (myMessage != null) {
+      myMessage = "```dart\n" + myMessage + "```";
       String potentialTemplate = template + "\n\n" + myMessage;
       if (potentialTemplate.length() <= MAX_URL_LENGTH) {
         template = potentialTemplate;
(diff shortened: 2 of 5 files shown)


Dart/src/com/jetbrains/lang/dart/DartBundle.properties
```diff
@@ -268,4 +268,7 @@ analysis.server.status.bad.text=Analyzer: Submit Feedback (analysis process not
 analysis.server.status.bad.desc=Analysis server needs to be restarted
 
 dart.feedback.url.template=https://github.com/dart-lang/sdk/issues/new?body=Analyzer Feedback from IntelliJ\n\
-  IDEA: {0} Dart: {1} {2}
\ No newline at end of file
+  IDEA: {0} Dart: {1} {2}
+dart.analysis.server.error=Dart analysis server error
+dart.error.file.instructions=Please insert the content of this file here:\n\
```
devoncarew: Perhaps `Please insert the contents of '{0}'` ?

devoncarew:
Awesome to see! Just for FYI, here's the crash template used by VSCode: https://github.com/dart-lang/sdk/issues/27904

alexander-doroshko:
Sorry, extremely busy today, will review tomorrow.

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
stevemessick: Those temp file URLs are very long, spanning a full line. Putting it on a separate line makes it easy to select and paste into a new browser tab. It seems markdown (or github) only formats links to http URLs, too (these are file: ones), so this isn't even a link.

stevemessick:
OK @alexander-doroshko I will take my laptop to Hawaii so I can respond to your comments while everything is still fresh in my mind.

Thanks @devoncarew for that template link. I have not done much to the formatting of the info section of the template but that obviously could be improved. I noticed that even without formatting characters the VSCode template is way over 4k. Makes me wonder if we can ignore the URL limit of older IE ... If we can it will make it more likely to get the full stack traces.

devoncarew:
Yeah, looks like Edge supports long urls. But note that the template for VSCode is always generated into a file. The user (in that tool) is required to open a browser themselves and paste in the content from the file (VSCode does open the file in the IDE to make that part easier).

stevemessick:
Following up the VSCode hint, I added Markdown formatting to the issue. I also discovered that a semicolon could, somewhat arbitrarily, cause all following text to be deleted when the generated URL was opened in Chrome (did not try other browsers). So, no more semicolon.

devoncarew:
Cool! Re: the semi-colon issue, you may need to url encode text inserted into the url (I believe the `URLEncoder` class would work for this).

stevemessick:
I'm pretty sure the encoding is done by the browser API.

On Nov 30, 2016 11:12, "Devon Carew" <notifications@github.com> wrote:

Cool! Re: the semi-colon issue, you may need to url encode text inserted
into the url (I believe the URLEncoder class would work for this).

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/JetBrains/intellij-plugins/pull/444#issuecomment-263965616>,
or mute the thread

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
```diff
@@ -3,25 +3,51 @@
...
 import com.jetbrains.lang.dart.sdk.DartSdk;
 import org.jetbrains.annotations.Nullable;
 
+import java.io.File;
+import java.io.IOException;
+
 import static com.intellij.ide.actions.SendFeedbackAction.getDescription;
 
 public class DefaultDartFeedbackBuilder implements DartFeedbackBuilder {
+  private String myMessage;
```
alexander-doroshko: Is it possible to make the class stateless and pass a message as a parameter of sendFeedback() method instead?

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
```diff
@@ -3,25 +3,51 @@
...
   public void sendFeedback(@Nullable Project project) {
     final ApplicationInfoEx appInfo = ApplicationInfoEx.getInstanceEx();
     boolean eap = appInfo.isEAP();
     String ijBuild = eap ? appInfo.getBuild().asStringWithoutProductCode() : appInfo.getBuild().asString();
     String sdkVsn = getSdkVersion(project);
-    String platDescr = getDescription();
+    String platDescr = getDescription().replaceAll(";", " ");
     String template = DartBundle.message("dart.feedback.url.template", ijBuild, sdkVsn, platDescr);
+    if (myMessage != null) {
+      myMessage = "```dart\n" + myMessage + "```";
```
alexander-doroshko: Will it modify myMessage several times if clicked on 'Submit feedback' button several times?

stevemessick:
Thanks for the comments. I'll look into it.

alexander-doroshko:
Tried to use it today. 
- Is there some reasonable issue title we could prefill?
- Is 'Configuration' part useful? Maybe omit it?
image

stevemessick:
My concern with the title is that if we make it too easy we won't get any
info other than what we automatically include. That may not be a good basis
for deciding to not pre-fill the title.

Some of the configuration is needed. Screens probably are not. I used it
because it was convenient but we could tune it.

BTW can you open the URL in that image in your browser? I'm never sure
about Windows syntax.

alexander-doroshko:
> Some of the configuration is needed.

Yes, OS may be of use. IDE version is already printed in the previous section

> BTW can you open the URL in that image in your browser?

Yes, Chrome showed me the file contents.

alexander-doroshko:
Got multiple ` ```dart ` in issue template:
image

devoncarew:
FYI, I'm working on a PR with some tweaks to the bug report template - it may cover some of these issues.

PR #445, opened 2016-12-02
Add Do Not Ask button to reporting query

stevemessick:
@alexander-doroshko Add a Do Not Ask button to the dialog that requests an issue report.

This also makes the builder stateless, which should fix the duplicate style issue reported earlier today.

Eventually, the value saved from the Do Not Ask check box should be stored as a preference.


Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartFeedbackBuilder.java
```diff
@@ -38,40 +42,61 @@ default String title() {
...
-  }
-
-  /**
    * Perform the action required to send feedback.
    *
    * @param project the current project
+   * @param errorMessage additional information for the issue, such as a tack trace
    */
-  void sendFeedback(@Nullable Project project);
+  public abstract void sendFeedback(@Nullable Project project, String errorMessage);
```
alexander-doroshko: I've added `@Nullable` to the `errorMessage` param here and for the implementation

PR #446, opened 2016-12-03
ui tweaks to the pubspec.yaml notifications provider

devoncarew:
Some ui tweaks to the pubspec.yaml editor notifications provider:

- add an icon
- use the `text()` label (for a similar look to other notifications providers)
- as a JSeparator to separate package actions from global ones

screen shot 2016-12-03 at 1 52 12 pm

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/actions/DartEditorNotificationsProvider.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/actions/DartEditorNotificationsProvider.java
@@ -105,11 +105,14 @@ public class DartEditorNotificationsProvider extends EditorNotifications.Provide
   private static class PubActionsPanel extends EditorNotificationPanel {
     private PubActionsPanel() {
       super(EditorColors.GUTTER_BACKGROUND);
-      myLinksPanel.add(new JLabel("Pub actions:"));
+
+      icon(DartIcons.Dart_16);
+      text("Pub actions");
+
       createActionLabel(DartBundle.message("get.dependencies"), "Dart.pub.get");
       createActionLabel(DartBundle.message("upgrade.dependencies"), "Dart.pub.upgrade");
       createActionLabel("Build...", "Dart.pub.build");
-      myLinksPanel.add(new JLabel("        "));
+      myLinksPanel.add(new JSeparator(SwingConstants.VERTICAL));
       createActionLabel("Repair cache...", "Dart.pub.cache.repair");
     }
   }


Dart/src/com/jetbrains/lang/dart/ide/actions/DartEditorNotificationsProvider.java
```diff
@@ -105,11 +105,14 @@ public EditorNotificationPanel createNotificationPanel(@NotNull final VirtualFil
   private static class PubActionsPanel extends EditorNotificationPanel {
     private PubActionsPanel() {
       super(EditorColors.GUTTER_BACKGROUND);
-      myLinksPanel.add(new JLabel("Pub actions:"));
+
+      icon(DartIcons.Dart_16);
+      text("Pub actions");
+
```
alexander-doroshko: I used `myLinksPanel` instead of `text()` to address layout on wide monitors. And used no icon, because it looks strange when not aligned to the left. Which one do you like more:
image
image

Dart/src/com/jetbrains/lang/dart/ide/actions/DartEditorNotificationsProvider.java
```diff
@@ -105,11 +105,14 @@ public EditorNotificationPanel createNotificationPanel(@NotNull final VirtualFil
...
       createActionLabel(DartBundle.message("get.dependencies"), "Dart.pub.get");
       createActionLabel(DartBundle.message("upgrade.dependencies"), "Dart.pub.upgrade");
       createActionLabel("Build...", "Dart.pub.build");
-      myLinksPanel.add(new JLabel("        "));
+      myLinksPanel.add(new JSeparator(SwingConstants.VERTICAL));
```
alexander-doroshko: Cool, committed manually.

Dart/src/com/jetbrains/lang/dart/ide/actions/DartEditorNotificationsProvider.java
devoncarew: It does degrade a bit when the editor is very wide. I think I still prefer using an icon and `text()` - the icon is nice that have visually, and the text() call makes the notification have a similar look to other ones.

But I don't feel strongly about it, if you want to handle the wide monitor use case well. I agree that an icon wouldn't make sense if not using text().

alexander-doroshko: I asked 2 more colleagues and we decided to keep label and links close to each other.

PR #448, opened 2016-12-05
move the errors status text to the view toolbar

devoncarew:
- remove the dedicated JLabel below the problems view table
- use the view content's display name area to display the summary of the errors and warnings
- use the toolbar to hold the analysis server status and feedback icon

This change allows more of the area in the problems view to be used for displaying errors, and leverages some IntelliJ constructs (`setDisplayName()`) to display UI (helping us unify our look with other views).

screen shot 2016-12-05 at 8 18 46 am

screen shot 2016-12-05 at 8 18 54 am

screen shot 2016-12-05 at 8 19 15 am

@alexander-doroshko @stevemessick for thoughts

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
@@ -29,6 +29,7 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.openapi.wm.ToolWindowAnchor;
 import com.intellij.openapi.wm.ToolWindowManager;
+import com.intellij.openapi.wm.ex.ToolWindowEx;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentFactory;
 import com.intellij.util.Alarm;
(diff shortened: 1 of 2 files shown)


devoncarew:
(I was not able to remove the double border around the view; to see the double border, switch between the dart analysis view and the todo view)

jwren:
LGTM

alexander-doroshko:
I like it. Pushing.
For the action in the tool window title we've lost 2 things: action description in the status bar on hover and hovered icon state. Not sure about the former, but the latter is doable by `presentation.setHoveredIcon` in `AnalysisServerStatusAction#updateStatus`.

For removing 2nd border there's `UIUtil.removeScrollBorder`, but it removes visual separation of the actions bar from the table as well:
image

devoncarew:
> but the latter is doable by presentation.setHoveredIcon in AnalysisServerStatusAction#updateStatus

I'm happy to iterate on this a bit; I do like the more concise layout.

alexander-doroshko:
> For removing 2nd border there's UIUtil.removeScrollBorder, but it removes visual separation of the actions bar from the table as well

extending `SimpleToolWindowPanel` does the trick pretty good:
image
Though there's still kinda double border because of a table cells at the right side:
image

alexander-doroshko:
I know how to disable drawing all vertical lines in the table, this way wee loose both double border and vertical line between description and location:
image

stevemessick:
Just chiming in with another Looks Good! I didn't know it was permissable
to put status indicators in the tool bar but it makes sense. Sorry I
haven't been able to do code reviews. Connectivity is sporadic here.

devoncarew:
> I know how to disable drawing all vertical lines in the table,

This looks great! And does look similar to other views, like the `TODO` and `Messages` views.

alexander-doroshko:
> And does look similar to other views, like the TODO and Messages views.

I think we are the only tool window with a table, all others are trees. Yes, vertical line in the middle is not so important. Strange that JTable renders it for selected lines anyway:
image

jwren:
Thanks for all the images Alex.  I've been quiet, but following the progress here, it is looking great.

PR #450, opened 2016-12-08
remember user settings in the Dart analysis view

devoncarew:
- remember user settings in the Dart analysis view (currently just the auto scroll to source setting)

@alexander-doroshko


alexander-doroshko:
Just perfect!

PR #451, opened 2016-12-08
don't show lines in the analysis view

devoncarew:
- don't show lines in the analysis view
- space the rows out a bit (they seem compressed currently; this spaces them out closer to how the other views (tree based) look)

I didn't see a good constant to use for the spacing, so just chose a min spacing (20) that looked good.

before:

screen shot 2016-12-07 at 6 47 45 pm

after:

screen shot 2016-12-07 at 6 48 48 pm

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsViewPanel.java
@@ -115,6 +115,8 @@ public class DartProblemsViewPanel extends SimpleToolWindowPanel implements Data
                                           : "");
 
     table.setShowVerticalLines(false);
+    table.setShowHorizontalLines(false);
+    table.setMinRowHeight(20);
     return table;
   }
 


devoncarew:
@alexander-doroshko

alexander-doroshko:
How about stripes?
image

devoncarew:
Looks good; does this mirror some other views in IntelliJ, or is there no established pattern here?

alexander-doroshko:
There are 30 usages of `setStripes(true)`, but hard to say how many tables without it. I'm afraid there's no common style, we can do as we want.
So would you prefer both zebra and increased row height or zebra only?
image

devoncarew:
> prefer both zebra and increased row height

Both sound good; the table definitely looks a little compressed without it.

alexander-doroshko:
Thanks! Merged manually.

alexander-doroshko:
Hint: we shouldn't use numbers directly when rendering UI, there's `JBUI.scale()` utility to respect HiDPI monitors. I changed it to
```
    table.setRowHeight(table.getRowHeight() + JBUI.scale(4));
```

devoncarew:
Cool, thanks for the info re: `JBUI.scale`.

PR #453, opened 2016-12-19
use LOG.warn() in the feedback provider

devoncarew:
When handling exceptions from the analysis server, use LOG.warn() instead of LOG.error(). The error() method will show an exception toast from IntelliJ, and we're already offering to report the issue via the feedback provider.

@alexander-doroshko @stevemessick


stevemessick:
LGTM

PR #454, opened 2016-12-20
Add server log to feedback

stevemessick:
@alexander-doroshko Add most recent server log lines to the feedback file. Make the file always get written.

This may take a few iterations to determine just what log info the analyzer teams wants to see.

cc @devoncarew

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -105,6 +105,9 @@ public class DartAnalysisServerService {
   private static final long MIN_DISRUPTION_TIME = 10000L; // 10 seconds minimum between error report queries
   private static final int MAX_DISRUPTIONS_PER_SESSION = 5; // Do not annoy the user too many times
 
+  private static final int DEBUG_LOG_CAPACITY = 30;
+  private static final int MAX_DEBUG_LOG_LINE_LENGTH = 200; // Saw one line while testing that was > 50k
+
   // Do not wait for server response under lock. Do not take read/write action under lock.
   private final Object myLock = new Object();
   @Nullable private AnalysisServer myServer;
@@ -139,6 +142,14 @@ public class DartAnalysisServerService {
 
   public long maxMillisToWaitForServerResponse = 0L;
   @NotNull private final InteractiveErrorReporter myErrorReporter = new InteractiveErrorReporter();
+
+  @NotNull private final Object myDebugLogLock = new Object();
+  @NotNull private final LinkedHashMap<String, String> myDebugLog = new LinkedHashMap<String, String>(DEBUG_LOG_CAPACITY) {
+    @Override
+    protected boolean removeEldestEntry(Map.Entry<String, String> eldest) {
+      return this.size() >= DEBUG_LOG_CAPACITY;
+    }
+  };
 
   private final AnalysisServerListener myAnalysisServerListener = new AnalysisServerListenerAdapter() {
 
@@ -1367,7 +1378,10 @@ public class DartAnalysisServerService {
       final DebugPrintStream debugStream = new DebugPrintStream() {
         @Override
         public void println(String str) {
-          //System.out.println("debugStream: " + str);
+          str = str.substring(0, Math.min(str.length(), MAX_DEBUG_LOG_LINE_LENGTH));
+          synchronized (myDebugLogLock) {
+            myDebugLog.put(str, str);
+          }
         }
       };
 
@@ -1672,7 +1686,7 @@ public class DartAnalysisServerService {
    * - The same message is not reported twice in a row
    * - The user is not interrupted too often
    */
-  private static class InteractiveErrorReporter {
+  private class InteractiveErrorReporter {
 
     @NotNull private QueueProcessor<Runnable> myErrorReporter = QueueProcessor.createRunnableQueueProcessor();
     private long myPreviousTime;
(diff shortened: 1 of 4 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -140,6 +143,14 @@
   public long maxMillisToWaitForServerResponse = 0L;
   @NotNull private final InteractiveErrorReporter myErrorReporter = new InteractiveErrorReporter();
 
+  @NotNull private final Object myDebugLogLock = new Object();
```
alexander-doroshko: Maybe use the collection itself as a lock, just to save a field?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -140,6 +143,14 @@
...
+  @NotNull private final LinkedHashMap<String, String> myDebugLog = new LinkedHashMap<String, String>(DEBUG_LOG_CAPACITY) {
```
alexander-doroshko: Maybe use some standard solution? I've found `com.google.common.collect.EvictingQueue` that seems to serve the purpose.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
```diff
@@ -29,21 +29,19 @@ public void sendFeedback(@Nullable Project project, @Nullable String errorMessag
...
-          template = potentialTemplate.substring(0, MAX_URL_LENGTH);
-        }
-        catch (IOException e) {
-          // ignore it
+      try {
+        File file = FileUtil.createTempFile("report", ".txt");
+        FileUtil.writeToFile(file, errorMessage);
+        if (serverLog != null) {
+          // Assume serverLog is never long enough that opening and closing the file is cheaper than copying it.
+          FileUtil.writeToFile(file, "\n\n" + serverLog, true);
```
alexander-doroshko: I think we can calculate contents first and then write it to a file in one shot.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
stevemessick: That caused IntelliJ to signal the field was accessed in both sync and async contexts.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
stevemessick: Is that considered "best practice" at JetBrains? That first string can be quite long, which was why I didn't try combining them, but it looks easy enough to change.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: It was StackOverflow where I found EvictingQueue :). Queue/List works O(1) and its algorithm is very simple whereas LinkedHashMap adds some overhead. Not a big deal for our case, just `put(str, str)` looked a bit unnatural to me.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Need to check. If inspection is buggy I'll file an issue.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: As one more pro I guess `debugLogContent()` implementation would be much simpler with simple collection :)

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
alexander-doroshko: I do not have any metrics in hands to support one or another approach. Just intuitively thought that string concatenation is more efficient than appending a file on a disk. Not a big deal anyway, this code is not called in a loop.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
stevemessick: I double-checked this and found that it works as expected. Removing the lock field is fine. I can't explain why it was complaining yesterday. I can change this when I get back to work after the holidays.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
stevemessick: I must be missing something. The complexity is due to reversing the order of the entries. Changing to a queue doesn't help with that. I don't think it is worth the trouble to rewrite.

alexander-doroshko:
Ok, pushed as is so far.

alexander-doroshko:
Also caught an exception and pushed a fix
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 4000
	at java.lang.String.substring(String.java:1963)
	at com.jetbrains.lang.dart.ide.errorTreeView.DefaultDartFeedbackBuilder.sendFeedback(DefaultDartFeedbackBuilder.java:41)
```
BTW do we still need to care about URL length since it doesn't include stack any more?

PR #457, opened 2016-12-29
add support for the PostPause debugger event

devoncarew:
- add the capability '`supports.pausePostRequest`'; the Flutter plugin uses that capability to know when it's safe to send the `pause` parameter when hot reloading an app
- when we get the `PausePostRequest` event, remove all existing breakpoints for the isolate and re-send the current user breakpoints. script ids for reloaded scripts will be different; we want to make sure that all reloaded scripts have the appropriate breakpoints set and that we no longer hit breakpoints in older scripts
- introduce a new private class (`IsolateBreakpointInfo`) to help organize per-isolate breakpoint information

@alexander-doroshko


alexander-doroshko:
@devoncarew Is there a way for me to test it without the Flutter plugin?

devoncarew:
The only client of the VM's hot reload APIs are Flutter currently. The current Flutter plugin does query for the `supports.pausePostRequest` capability and set the `pause` flag on a reload. So you could use it -

The only other way would be to add some temporary reload capability to the CLI debugger. This would be something like an action that issues the `_reloadSources` call (https://github.com/flutter/flutter/blob/25e6cc6743d35c50d6264dbd83b4978417382a3b/packages/flutter_tools/lib/src/vmservice.dart#L795).

devoncarew:
> Is there a way for me to test it without the Flutter plugin?

(I have done testing on my end)

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
```diff
@@ -116,3 +104,70 @@ public void breakpointFailed(@NotNull final XLineBreakpoint<XBreakpointPropertie
...
+    }
+
+    for (String vmBreakpointId : allVmBreakpoints) {
+      myDebugProcess.getVmServiceWrapper().removeBreakpoint(myIsolateId, vmBreakpointId);
+    }
+
+    return allVmBreakpoints;
+  }
+
+  public void addTemporaryBreakpoint(String breakpointId) {
```
alexander-doroshko: Since we have a lot of entities that have 'breakpoint' word, we could clarify parameter to be `vmBreakpointId`.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
```diff
@@ -116,3 +104,70 @@ public void breakpointFailed(@NotNull final XLineBreakpoint<XBreakpointPropertie
     return myVmBreakpointIdToXBreakpointMap.get(vmBreakpoint.getId());
   }
 }
+
+class IsolateBreakpointInfo {
+  private final String myIsolateId;
+  private final DartVmServiceDebugProcess myDebugProcess;
+  List<String> myTemporaryBreakpoints = new ArrayList<>();
```
alexander-doroshko: Since we have a lot of different entities that have 'breakpoint' word, we can clarify field to be `myTemporaryVmBreakpointIds`.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
```diff
@@ -67,35 +54,36 @@ public void vmBreakpointAdded(@NotNull final XLineBreakpoint<XBreakpointProperti
...
-    isolateAndVmBreakpointIds.add(Pair.create(isolateId, vmBreakpoint.getId()));
+    IsolateBreakpointInfo info = getIsolateInfo(isolateId);
+    info.vmBreakpointAdded(xBreakpoint, vmBreakpoint);
 
     if (vmBreakpoint.getResolved()) {
       breakpointResolved(vmBreakpoint);
     }
   }
 
   public void temporaryBreakpointAdded(String isolateId, Breakpoint breakpoint) {
```
alexander-doroshko: parameter could be `vmBreakpoint`

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
alexander-doroshko: This method name makes me think that it changes the VM state (i.e. uses methods of `VmServiceWrapper`), similar to `removeTemporaryBreakpoints()`. In fact it behaves similarly to `vmBreakpointAdded()` and could be called like `temporaryVmBreakpointAdded()`. Agree?

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
```diff
@@ -116,3 +104,70 @@ public void breakpointFailed(@NotNull final XLineBreakpoint<XBreakpointPropertie
...
+  }
+
+  public void removeTemporaryBreakpoints() {
+    for (String breakpointId : myTemporaryBreakpoints) {
+      myDebugProcess.getVmServiceWrapper().removeBreakpoint(myIsolateId, breakpointId);
+    }
+    myTemporaryBreakpoints.clear();
+  }
+
+  public Set<String> removeAllVMBreakpoints() {
```
alexander-doroshko: Lowercase `m` in `Vm` for consistency

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
devoncarew: done

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
devoncarew: done

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
devoncarew: done

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
devoncarew: done

devoncarew: makes sense - updated

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
```diff
@@ -80,7 +80,8 @@ public void received(@NotNull final String streamId, @NotNull final Event event)
         break;
       case PausePostRequest:
         // We get this event after an isolate reload call, when pause after reload has been requested.
-        // TODO: Set current breakpoints; resume the isolate.
+        // This adds the "supports.pausePostRequest" capability.
+        myDebugProcess.getVmServiceWrapper().restoreBreakpointsForIsolate(event.getIsolate().getId());
```
devoncarew: I can add a `CountDownLatch` into restoreBreakpointsForIsolate. I don't think they'll be an issue however, as the requests into the isolate will be received in order (set bp 1, set bp 2, resume). Happy to do either -

devoncarew: Re: the shared logic - I would prefer to have the two different variants, as what they do is slightly different (set all and resume the isolate, vs remove and re-set all).

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceBreakpointHandler.java
```diff
@@ -46,15 +40,8 @@ public void registerBreakpoint(@NotNull final XLineBreakpoint<XBreakpointPropert
   public void unregisterBreakpoint(@NotNull final XLineBreakpoint<XBreakpointProperties> xBreakpoint, boolean temporary) {
     myXBreakpoints.remove(xBreakpoint);
 
-    final Collection<Pair<String, String>> isolateAndVmBreakpointIds = myXBreakpointToIsolateAndVmBreakpointIdsMap.remove(xBreakpoint);
```
devoncarew: Good catch, and I can confirm that that's what was happening. I added a line in `IsolateBreakpointInfo.unregisterBreakpoint()` to clear the map entry for the given xBreakpoint.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
devoncarew: Update here: I did switch to using a closure and an AtomicInteger.

devoncarew:
ptal, I think I addressed most of your comments.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
```diff
@@ -280,6 +280,46 @@ public void onError(RPCError error) {
...
+
+        @Override
+        public void onError(RPCError error) {
+          myBreakpointHandler.breakpointFailed(xBreakpoint);
+          checkDone();
+        }
+
+        private void checkDone() {
+          if (counter.decrementAndGet() == 0 && onFinished != null) {
+            onFinished.run();
```
alexander-doroshko: Looks like `onFinished.run()` is never called if `xBreakpoints` is empty. I've extracted common code including the fix for this, no actions needed.

devoncarew: Opps, thanks!

PR #460, opened 2017-01-04
rev to the latest version of the service protocol

devoncarew:
- rev to the latest version of the service protocol

@alexander-doroshko, this brings in the new stepping type `Rewind`; when passed into `resume()` this should let us implement drop frame.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceListener.java
@@ -59,6 +59,8 @@ public class DartVmServiceListener implements VmServiceListener {
       case IsolateExit:
         myDebugProcess.isolateExit(event.getIsolate());
         break;
+      case IsolateReload:
+        break;
       case IsolateRunnable:
         break;
       case IsolateStart:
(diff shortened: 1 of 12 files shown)


devoncarew:
And re: `Rewind` - I'm happy to work on that later. If you do get to it though, please note that not all clients support poping stack frames off, so checking the return code from that step operation is important.

devoncarew:
Looks like resume() is missing the new (optional) `frameIndex` param. We can hold off on landing this PR, or land and update the library later.

alexander-doroshko:
> And re: Rewind - I'm happy to work on that later

Sounds great! :)

> We can hold off on landing this PR, or land and update the library later.

Either way looks good, let me know about your preferences.

devoncarew:
@johnmccutchan will likely have an update to the protocol spec today - I can update this PR then.

johnmccutchan:
Note that even on supported platforms, Rewind can and will fail for certain `frameIndex` values. Some frames are optimized and we cannot rewind to them. @turnidge can explain in detail.

johnmccutchan:
https://codereview.chromium.org/2624493003/ <-- spec fix for frameIndex

devoncarew:
Updated, ptal.

PR #461, opened 2017-01-05
[osgi] Add Resolver support for .bndrun files

brampouwelse:
@trespasserw Could you have a look at this one aswel for me? This adds support for using the resolver from IntelliJ.


trespasserw:
Hi Bram,

I'm back, finally. While I work on other requests, could you please describe in a couple words, what does this feature do?

brampouwelse:
Hi @trespasserw, 

Sure I can, the resolver can be used to create a list of `-runbundles` based on requirements. So instead of listing all bundles you need in order to run your application you basically state I want to have a runtime that's capable of running `my.module` by adding 

```
osgi.identity;filter:='(osgi.identity=my.module)'
```

Based on requirements of `my.module` the resolver determines which bundles are necessary to run that module and generates the `-runbundles` list. 

Some more detailed information on the topic of resolving: 
 - http://enroute.osgi.org/appnotes/resolution-notes.html

trespasserw:
Hi Bram,

sorry, almost forgot about this request. I have merged the changes, then updated the action to do its work in background, so it would not lock the UI during the long resolve.

brampouwelse:
Cool, thanks! Will have a look on how to properly do these background actions for future changes :)

PR #463, opened 2017-01-10
make scheduleConnect public

devoncarew:
- make `scheduleConnect()` public (this will help us override its behavior in the Flutter subclass)

@alexander-doroshko


alexander-doroshko:
Will `protected` be enough?

devoncarew:
Yes, protected should work.

devoncarew:
Updated -

alexander-doroshko:
Ok, pushed manually.

PR #464, opened 2017-01-10
Enhanced language guessing for markdown code blocks to support R-markdown

holgerbrandl:
In its current implementation the "JetBrains Markdown" plugin fails to correctly inject R into r-codechunks.

A very popular approach in the R community is to write so-called [Rmd-files](http://rmarkdown.rstudio.com/) which are regular markdown files that are rendered into html/pdf/doc using tools like [knitr](https://yihui.name/knitr/). During the rendering process code chunks  are replaced with plots and results.

To guide formatting and to adjust chunk execution the initial line of the code chunks often includes some formatting and evaluation hints:
~~~~
```{r fig.height=23, eval=T}
1+1
plot(1:1)
```
~~~~

See http://rmarkdown.rstudio.com/authoring_rcodechunks.html for a reference.

The provided patch enhances the language guessing for markdown code blocks in the "Jetbrains Markdown" plugin to support the **r**markdown chunks as well.

### DIFF
--- a/markdown/src/org/intellij/plugins/markdown/injection/LanguageGuesser.java
+++ b/markdown/src/org/intellij/plugins/markdown/injection/LanguageGuesser.java
@@ -50,7 +50,10 @@ public enum LanguageGuesser {
 
   @Nullable
   public Language guessLanguage(@NotNull String languageName) {
-    final Language languageFromMap = langIdToLanguage.getValue().get(languageName.toLowerCase(Locale.US));
+    // Add support for rmarkdown chunks (see http://rmarkdown.rstudio.com/authoring_rcodechunks.html)
+    languageName = languageName.startsWith("{r ") ? "r" : languageName;
+
+    final Language languageFromMap = langIdToLanguage.getValue().get(languageName);
     if (languageFromMap != null) {
       return languageFromMap;
     }
--- a/markdown/src/org/intellij/plugins/markdown/injection/LanguageGuesser.java
+++ b/markdown/src/org/intellij/plugins/markdown/injection/LanguageGuesser.java
@@ -50,6 +50,8 @@ public enum LanguageGuesser {
 
   @Nullable
   public Language guessLanguage(@NotNull String languageName) {
+    languageName = languageName.toLowerCase(Locale.US);
+
     // Add support for rmarkdown chunks (see http://rmarkdown.rstudio.com/authoring_rcodechunks.html)
     languageName = languageName.startsWith("{r ") ? "r" : languageName;
 


valich:
Hi @holgerbrandl,

I am sorry I must reject your change once again :( I don't think it's a good idea to mention more custom cases in the markdown plugin so this logic actually should be implemented by the plugin providing such language(s).

The good news are:
1. At last [I've added](https://github.com/JetBrains/intellij-plugins/commit/00d669fab679c1f1f9031350fe4f4b9db10eb390) the normal way to work on the plugin externally so you may try to do this soon enough (there is a commit that relies on the newest API to you'll need to wait for the next IC EAP)
2. I've [just implemented](https://github.com/JetBrains/intellij-plugins/commit/e6d916f203851e186fd1b5ab74d2d5d8139b25c1) an extension which allows for such behaviour changes. You can see the new test in that commit for an example.

So what I propose is to implement such extension in your R plugin. There is an ability to set up optional dependency in `plugin.xml`, [see](https://github.com/JetBrains/intellij-plugins/tree/master/markdown/resource/META-INF) how optional dependency for IntelliLang is implemented in markdown plugin.

Whether you have any questions don't hesitate to ask and please contribute more if you'll have any ideas.

holgerbrandl:
>  I don't think it's a good idea to mention more custom cases in the markdown plugin

I totally agree, I just followed the javascript lead in code even if it already started to look ugly. Imho also the javascript plugin should/could provide a custom fenceLanguageProvider for sake of consistency now.

> 2. I've just implemented an extension which allows for such behavior changes

If I understand correctly, I simply would need to add 

```
<depends optional="true" config-file="plugin-intelliLang.xml">org.intellij.intelliLang</depends>

...
    < fenceLanguageProvider implementation="com.r4intellij.RCodeFenceLanguageProvider"/>

```
If so, it seems like a piece of cake.

Would intelli gracefullyj ignore the `fenceLanguageProvider` setting on installations which still run the old/current versions of your Markdown plugin and IJ? 

While looking at your changelist, I've noticed MarkdownHtmlPanelProvider. Can I modulate html-preview in a similar way by providing a language specific HtmlPanelProvider? (which would be fantastic)

valich:
1. > Imho also the javascript plugin should/could provide a custom fenceLanguageProvider for sake of consistency now.

   I totally agree however I realized that not so long ago; I I'll address that sometime.

2. Your dependency should rather look like `<depends optional="true" config-file="whatever-name.xml">org.intellij.plugins.markdown</depends>` and in that `whatever-name.xml` you should include your extension point. Not sure about plugin version dependency though. I'll check that tomorrow but in general since that functionality in markdown plugin will be available in 2017.1 releases anyway (as a bundled or installed plugin) your implementation could work only there; that means that you're safe to assume that extension will be available if you set `since-build` for your plugin as `2017.1.2272`.

3. > Can I modulate html-preview in a similar way by providing a language specific HtmlPanelProvider?

   You can provide own `HtmlPanelProvider` along with `MarkdownHtmlPanel` but this particular one is not so easy. In fact, we're talking not about rendering itself customization (which is HtmlPanels are about) but html which is generated. If I understand what you'd like to provide, you'll need to make your own markdown flavour for that. And this is quite easily possible to do as an extension to [intellij-markdown](https://github.com/valich/intellij-markdown) parsing which may be done in your R plugin, too. If you're keen to do that, I'll make some effort to write a short tutorial on that (and update Readme at last) and you'll see it should not be difficult task.

valich:
Sorry for incorrect info, it will not be available in `171.2272`s but rather `171.2303` and later.

PR #465, opened 2017-01-13
Fix WEB-24013

stevemessick:
@alexander-doroshko This fixes https://youtrack.jetbrains.com/issue/WEB-24013

It changes three tests to be more consistent with Java behavior. I believe you had requested that change earlier.

I added a note to one of the tests explaining that it is moving the code one line too far. I don't have time to fix that right now. I'll add an issue to the tracker.

Edit: https://youtrack.jetbrains.com/issue/WEB-24986

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/moveCode/DartStatementMover.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/moveCode/DartStatementMover.java
@@ -58,7 +58,7 @@ public class DartStatementMover extends LineMover {
     if (statements.length == 0) return false;
     range.firstElement = statements[0];
     range.lastElement = statements[statements.length - 1];
-    info.indentTarget = false;
+    info.indentTarget = true;
     if (!checkMovingInsideOutside(file, editor, info, down)) {
       info.toMove2 = null;
     }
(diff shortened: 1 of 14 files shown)


alexander-doroshko:
I tried the example from https://youtrack.jetbrains.com/issue/WEB-24986. 
I'd expect `int i=0;` to become the last statement inside `else{}` block after moving up. It works this way in Java. In other words, `VarIfDo_afterUp.dart` should be 
```
    if (x) n(); else {
      m();
      int i=0;
    }
```

As a rule moving any statement up and down, as well as down and up should result in an untouched code. It's not the case for `int i=0;` line in this example.

stevemessick:
You've restated the issue and added the obvious declaration of invertible
idempotency, which is implied in the issue. Do you have a question? I'm not
sure what the purpose of this comment is.

alexander-doroshko:
Sorry for the confusion, I looked at the incorrect tab in my browser and tested this PR against WEB-24986, but should have tested against WEB-24013. Taking another look. Sorry again.

alexander-doroshko:
Verified that it fixes the case from the 1st comment to the WEB-24013. Pushing.

alexander-doroshko:
P.S. Got exception trying to move `foo();` up:
```
main() {
  foo();
}
```

PR #466, opened 2017-01-13
Fix exception thrown when moving statement from minimal main

stevemessick:
@alexander-doroshko Fixes the exception you found earlier today.


alexander-doroshko:
👍

PR #468, opened 2017-01-17
Address Dart Generator issue WEB-24496, fixes to == and hashCode generation

jwren:


Dart/testData/generate/EqualsAndHashCode1.txt
```diff
@@ -5,18 +5,18 @@ class EqualsAndHashCode1 {
...
+              str == other.str &&
+              v == other.v &&
+              i == other.i;
 
   @override
-  int get hashCode {
-    return str.hashCode ^ v.hashCode ^ i.hashCode;
-  }<caret>
+  int get hashCode =>
+      super.hashCode ^
```
Adamovskiy: Now hashCode uses `super`, but `==` is no - leads to equals-hashCode contract violation. Should be added to `==` too.

jwren: Understood.  Would you be okay with doing the runtime check without having another option in the dialog?  I'd rather not have the additional option.

alexander-doroshko: As I understand @Adamovskiy he' asking to add `super == other &&` to `==()` implementation. Looks like it should be fine for all cases, no need in option. 
@Adamovskiy, is it ok?

Adamovskiy: @alexander-doroshko is right. It would be nice to have optional `runtimeType == other.runtimeType` to prevent subclasses, but `super == other` in `==()` is always required if `super.hashCode` is in `hashCode`.

alexander-doroshko: Thanks!
I think we can go without any options for now.
Two checks `other is ThisClass &&  runtimeType == other.runtimeType` seem to be redundant. It's enough to have the first one if we allow subclasses, and the second if we don't. Agree?
My preference is for `runtimeType == other.runtimeType` as it gives better guarantees to meet the transitivity requirement imposed by the equals contract.
`super == other` required in any case.

Adamovskiy: Isn't `other is ThisClass` is transitive as well? It is not symmetric. But, unlike `runtimeType`, it gives to analyzer type information for `other`. This is also the reason to leave both `other is ThisClass && runtimeType == other.runtimeType`. Case when behavior of this two implementations is not same - when class with fields has subclass with no extra fields. I am not sure, but think that in most cases is some kind of proxy, witch should be treated as instance of original class. Symmetricity problem in this case is solved by delegating `proxy.==(other)` to `other.==(this)`. So I think it is better to leave just `other is ThisClass`, if there is no option for now.

alexander-doroshko: Agree, keeping `other is ThisClass` to give analyzer a hint sounds reasonable.

Quick glance at [this article](http://www.angelikalanger.com/Articles/JavaSolutions/SecretsOfEquals/Equals-2.html) made me think that we are safer with `runtimeType == other.runtimeType`.

jwren:
PTAL.  runtimeType check added in this PR, and I did a rebase to keep the master branch in intellij-pliugins/ clean.

alexander-doroshko:
Please take a look at the 1st comment by Alexey. Equals-hashCode contract violation is caused by using super in hashCode, but not using in ==.

> I did a rebase to keep the master branch in intellij-pliugins clean.

👍

jwren:
Thanks Alex... PTAL... I think that was the last missing piece?

alexander-doroshko:
@jwren @Adamovskiy It appeared that the task is a bit more complex than we thought. 
We can't call `super.hashCode` and `super.==` if it results in calling `Object.hashCode` and `Object.==`. Because [implementation](https://api.dartlang.org/stable/1.21.1/dart-core/Object/operator_equals.html) in `Object` class is
```
bool operator ==(other) => identical(this, other);
```
which means all the following fields comparison just an unreachable code.

We need to check superclasses hierarchy, search for `hashCode` and `==` implementation presence in super classes (all except `Object`), and add `super.hashCode` and `super.==` to the generated code only if it won't result in calling `Object.hashCode` and `Object.==`. 

@jwren How fo you update the PR without introducing new commits? It's cool! :)

jwren:
@alexander-doroshko For this level of resolution, does some non-Object override == or hashCode, any reason not to use the Dart analysis server?  Starting with some content in DartHierarchyUtil, it looks like utility methods already exist for the majority of what I need to do.

@alexander-doroshko interactive rebase:
    git rebase -i HEAD~3
a "--force" push is required since you are re-writing history (be aware that some git repos, such as in dart prevent this)

Some docs: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)

alexander-doroshko:
> any reason not to use the Dart analysis server

I think using server is the right way to go (and actually the only way)

> interactive rebase

Great! There's even an integration with 'interactive rebase' and 'force push' features in IntelliJ IDEA, though I didn't try. AFAIK force push is forbidden for common branches in our repos, but is welcome for custom ones.

jwren:
PR updated again.  While we check to see if a super class implements == or hashCode, the check doesn't look at the implementations of the methods.  If we want this, we can follow up.

alexander-doroshko:
Thanks, pushed.

I also moved a request to the server outside of a write action - in `beforeInvoke()` method.

Also I've found a bug in `doesSuperclassOverrideEqualEqualAndHashCode()` implementation: it iterates not only super classes, but also implemented interfaces and mixins. You need to iterate the list respecting `TypeHierarchyItem.getSuperclass()`. PR is welcome :).

alexander-doroshko:
@jwren, will you have time to fix super classes iteration? Otherwise, I can fix it myself.

jwren:
Do we want to check for interfaces?  While an interface's implementation doesn't mean that when our class is called the == or hashCode in the interface will be called.

alexander-doroshko:
> Do we want to check for interfaces? 

No, and that's the bug we need to fix as they are checked now. We need to iterate the list respecting `TypeHierarchyItem.getSuperclass()`.

jwren:
Got it, yes, I'll create a PR today

jwren:
Follow up: https://github.com/JetBrains/intellij-plugins/pull/484

PR #469, opened 2017-01-20
Use balloon instead of modal query

stevemessick:
@alexander-doroshko This changes the UI for the server bug reporter to use a non-modal balloon instead of the modal query.  It plays nicely with the standard IntelliJ error balloon: they stack and disappear in the order they appear. You can make it go away faster by clicking the X or you can click the link to open the github issue tracker in your web browser.

Thanks for changing the Dart SDK to be project-specific!


alexander-doroshko:
Cool!

I also added writing the error to the Event Log tool window. If user misses the balloon he/she still can click the link in the Event Log:
image

stevemessick:
Great idea, thanks!

alexander-doroshko:
@stevemessick Since error balloon is not so annoying now as modal dialog may be increase or remove MAX_DISRUPTIONS_PER_SESSION?
MIN_DISRUPTION_TIME might be not needed as well.

stevemessick:
Good idea. I sent you a new PR.

PR #470, opened 2017-01-24
Newlining

stevemessick:
@alexander-doroshko Make the Dart indenter be sensitive to whether it is doing a wholesale reformat or a local indentation computation. It should reduce GC churn slightly, too.

This gives us the framework to resolve other discrepancies between what the formatter does and what the user expects while typing code, once we identify them.

This PR should get cherry-picked into the current plugin release, if possible.

/cc @devoncarew


alexander-doroshko:
Great, pushed!
Need to check how it merges into 163. There may be problems both with the Platform API and with other commits in the Dart plugin that also should be included.

PR #474, opened 2017-02-02
Fix for https://youtrack.jetbrains.com/issue/RUBY-18536

holgerbrandl:
The PR implements a new structure view model that shows sections in a cascading tree. This seems to most natural way to visualize the structure of a Markdown file.

Before: 
image

WIth PR:
image

Let me know if there are any questions

### DIFF
--- a/markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureElement.java
+++ b/markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureElement.java
@@ -1,30 +1,34 @@
 package org.intellij.plugins.markdown.structureView;
 
+import com.intellij.ide.IdeBundle;
 import com.intellij.ide.structureView.StructureViewTreeElement;
+import com.intellij.ide.structureView.impl.common.PsiTreeElementBase;
 import com.intellij.ide.util.treeView.smartTree.SortableTreeElement;
-import com.intellij.ide.util.treeView.smartTree.TreeElement;
 import com.intellij.navigation.ItemPresentation;
+import com.intellij.navigation.LocationPresentation;
 import com.intellij.navigation.NavigationItem;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.impl.source.PsiFileImpl;
 import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.tree.TokenSet;
-import com.intellij.util.ArrayUtil;
 import org.intellij.plugins.markdown.lang.MarkdownElementTypes;
 import org.intellij.plugins.markdown.lang.MarkdownTokenTypeSets;
-import org.intellij.plugins.markdown.lang.psi.impl.MarkdownBlockQuoteImpl;
-import org.intellij.plugins.markdown.lang.psi.impl.MarkdownCompositePsiElementBase;
 import org.intellij.plugins.markdown.lang.psi.impl.MarkdownFile;
-import org.intellij.plugins.markdown.lang.psi.impl.MarkdownListItemImpl;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
-public class MarkdownStructureElement implements StructureViewTreeElement, SortableTreeElement {
+import static org.intellij.plugins.markdown.lang.MarkdownTokenTypeSets.*;
+
+public class MarkdownStructureElement extends PsiTreeElementBase<PsiElement> implements  SortableTreeElement, LocationPresentation {
+
   private static final ItemPresentation DUMMY_PRESENTATION = new MarkdownBasePresentation() {
+
     @Nullable
     @Override
     public String getPresentableText() {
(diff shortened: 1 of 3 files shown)


holgerbrandl:
@valich I guess you'd be the one to judge it. :-)

holgerbrandl:
Potentially the h[1-6] could be removed from the ItemPresentation and the headers could be whitespace-trimmed as well. This would improve structure view popup (meta-f12) search-as-you-type, which still feels a bit dodgy at the moment.

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureElement.java
```diff
@@ -117,28 +103,63 @@ private ItemPresentation getPresentationImpl() {
...
+    final List<MarkdownStructureElement> childrenElements = new ArrayList<MarkdownStructureElement>();
+
+    PsiElement nextSibling = myElement instanceof MarkdownFile ?
+            myElement.getFirstChild().getFirstChild() :
+            myElement.getNextSibling();
+
+    PsiElement maxContentLevel = null;
+
+    while (nextSibling != null) {
+      if (nextSibling != myElement && isSameLevelOrHigher(nextSibling, myElement)) {
```
valich: Can the check `nextSibling != myElement` fail at all? I see that nextSibling is initialized as a child or sibling to `myElement` and only moves further..

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureViewFactory.java
```diff
@@ -38,10 +43,16 @@ protected boolean isSuitable(PsiElement element) {
         return false;
       }
       final PsiElement parent = element.getParent();
-      if (MarkdownStructureElement.hasTrivialChild(parent)) {
+      if (hasTrivialChild(parent)) {
```
valich: Is this acceptable after your change? I feel like it may ignore some elements incorrectly.

valich:
@holgerbrandl Also have you signed our [Contributor Agreement](https://www.jetbrains.com/agreements/cla/) yet? If not, please sign it, it's necessary for us to accept external contributions.

holgerbrandl:
> 1. The older commits (concerning the previous PR) must be removed because they will mess the commit history;

Done.

> 2. I am not sure this code works well in the following case:
> ` * # Header1`

Indeed this not yet supported. But I've never seen those in real-life markdown.

>Continuing the previous concern, what do you think about including lists and tables in the structure? I think there may be some scenarios where is may be useful (changelogs, api documentation, smth more?)

Personally I'd think that the view should just show some high-level structure abstraction and not the complete DOM. Similar to the java-structure which stops on method level and does not list for loops and statements. I've tried to include more types but then the structure-view becomes immediately very cluttered and looses a lot of its appeal.

> There are no tests. 

Indeed, no tests. I'm  curious about how to test a structure-view. 

> There are some codestyle concerns (like the lack of @Nullable/@NotNull annotations) 

I've added them to all new methods where possible.

> is `element != null` really necessary?

Well spotted. No it's not needed and I've removed it.

Concerning `isSuitable` implementation:
> Is this acceptable after your change? I feel like it may ignore some elements incorrectly.

It took me a while to realize that `isSuitable` is needed to navigate to structure from source. Indeed the first version of this PR broke that functionality. It should be fine now (see 2nd commit). Except if the user moves the caret to the end of the file, in which case the argument to `org.intellij.plugins.markdown.structureView.MarkdownStructureViewFactory.MarkdownStructureViewModel#findAcceptableElement` seems to become `null` and there's no obvious way to find an acceptable parent in such a situation.

Concerning Item presentation:
I've did as suggested in my earlier comment and inverted the ItemPresentation to fix the structure popup keyboard search.
image

I've just signed the Contributor Agreement.

Feel welcome if you've further questions or suggestions.

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureViewFactory.java
```diff
@@ -29,19 +36,21 @@ public MarkdownStructureViewModel(@NotNull PsiFile psiFile, @Nullable Editor edi
...
-      }
-      if (!MarkdownStructureElement.PRESENTABLE_TYPES.contains(element.getNode().getElementType())) {
-        return false;
-      }
-      final PsiElement parent = element.getParent();
-      if (MarkdownStructureElement.hasTrivialChild(parent)) {
-        return false;
+    protected Object findAcceptableElement(PsiElement element) {
+      // walk up the psi-tree until we find an element from the structure view
+      while (element != null && !PRESENTABLE_TYPES.contains(element.getNode().getElementType())) {
```
holgerbrandl: I could imagine that there's is some utility method for this type of upward-directed psi-traversal, but I could not find it.

valich: There is `com.intellij.psi.util.PsiTreeUtil#getParentOfType` and one may use `MarkdownHeaderImpl.class` as an element class

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureElement.java
```diff
@@ -38,107 +42,149 @@ public String getLocationString() {
...
+
+  private boolean isSameLevelOrHigher(@NotNull PsiElement psiA, @NotNull PsiElement psiB) {
+    IElementType typeA = psiA.getNode().getElementType();
+    IElementType typeB = psiB.getNode().getElementType();
+
+    return headerLevel(typeA) <= headerLevel(typeB);
+  }
+
+
+  private int headerLevel(@NotNull IElementType curLevelType) {
```
valich: One may get header level from the `org.intellij.plugins.markdown.lang.psi.impl.MarkdownHeaderImpl#getHeaderNumber` (needs visibility escalating)

valich:
> This in particular means that MarkdownStructureElement extends PsiTreeElementBase to make the Location (ie the actual section header strings) searchable (see com/intellij/ide/util/FileStructurePopup.java:929)

Whoa, that's a bigger change! My tears flow looking at our API sometimes but let's live with this.. It's just more difficult to check the changes in one commit in such situation. (BTW I'd prefer having two — one "refactoring" commit to make our class use base implementation and the second with the real logic changes; since you've squashed them already there's nothing to do :) )

I am writing tests for the structure view presentation right now; If everything goes right, I'll just merge your commit and put the tests right alongside.
Also I can do a little cleanup (yellow code, stdlib usage) and resolve my comments above myself; I can also leave it to you if you prefer but that'll take more time, obviously.

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureViewFactory.java
holgerbrandl: Since the psi-tree generated by the current markdown parser is more a flat list than a tree, I think `getParentOfType` won't find the section element. But I may be wrong.

holgerbrandl:
>My tears flow looking at our API sometimes but let's live with this.. It's just more difficult to check the changes in one commit in such situation

Indeed it felt a bit odd to require an implementation and not an interface in `FileStructurePopup` to allow for searchable locations.

> you've squashed them already
You asked for squashing, and I did as much as I could. :-)

valich:
@holgerbrandl I've managed to implement the tests. The tricky part was to include locationStrings in the output. See 2a0ce3b for details if you'll need to implement that in your plugin.

Also I've decided to find headers located in quotes/lists. This is a rare case indeed but I trust that we should support every sensible case given the price is not too high. After a couple of tries I've found quite short and sensible solution, see 7e73c77. If you think something is wrong or you have any further improvements to suggest, you're welcome :)

Thanks for the contribution!

markdown/src/org/intellij/plugins/markdown/structureView/MarkdownStructureViewFactory.java
valich: I did not get your question then. Well, there are either `getParentOfType` or `getNext/PrevSiblingOfType`, more complex traversal is hand-written usually..
There are some nice utilities like `com.intellij.psi.SyntaxTraverser`, but they are still quite general

holgerbrandl:
I like your TRANSPARENT_CONTAINERS to allow for headers in lists without cluttering the structure view with ol and ul. Looks really good, and works well.

Totally unrelated: also many thanks for the the gradle configuration. It took me a while to realize that it actually allows to run your plugin against ij-trunk-snapshot, which is really cool and makes  the frustrations I had in the past when trying to run it with my public-release ij go away.

PR #475, opened 2017-02-04
add an icon for the dart analysis view decorated with an error

devoncarew:
- add an icon for the dart analysis view decorated with an error

This could be used to badge the Dart Analysis tool window icon when there are analysis errors.

@alexander-doroshko


alexander-doroshko:
I've found `AllIcons.Nodes.ErrorMark`. This is what `LayeredIcon.create(DartIcons.Dart_13, AllIcons.Nodes.ErrorMark)` looks like on standard resolution monitors:
image
Looks a bit more contrast to me. Which one do you find better?

devoncarew:
`AllIcons.Nodes.ErrorMark` sgtm, esp. as it's designed as a badge, and would look similar to other uses in IntelliJ. Closing...

PR #477, opened 2017-02-06
Allow named args to be moved like exprs

stevemessick:
@alexander-doroshko Extending the previous PR to accept named arguments as well as list expression elements. I'm not planning to do it for positional args, but we can talk about that if you like.


alexander-doroshko:
Pushed!
I'm not a big user of it in Dart, so you may set priorities as you think would be right.
Note that I'm reviewing 'move statement' functionality a bit less thoroughly than other PRs as it would require too much time and dozens of corner case checks.

PR #478, opened 2017-02-06
Make newline after bracket like newline after brace

stevemessick:
@alexander-doroshko Move the closing right bracket down two lines after a newline, just like a closing right brace.


alexander-doroshko:
super!

PR #480, opened 2017-02-08
fold exception traces

devoncarew:
Fold exception traces in the console. This takes potentially very long traces, and folds the `#0 ...` lines.

before:
screen shot 2017-02-07 at 3 04 53 pm

after:
screen shot 2017-02-07 at 4 27 16 pm

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartConsoleFolding.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartConsoleFolding.java
@@ -14,16 +14,23 @@ import org.jetbrains.annotations.Nullable;
 import java.io.File;
 import java.util.List;
 import java.util.Locale;
+import java.util.regex.Pattern;
 
 public class DartConsoleFolding extends ConsoleFolding {
 
   private static final String DART_MARKER = SystemInfo.isWindows ? "\\bin\\dart.exe " : "/bin/dart ";
   private static final String TEST_RUNNER_MARKER = "pub.dart.snapshot run test:test -r json "; // see DartTestRunningState.startProcess()
 
+  //  #1      main (file:///Users/foo/projects/bar/tool/generate.dart:30:3)
+  private static final Pattern EXCEPTION_PREFIX = Pattern.compile("#\\d+ +[^ ].*");
+
   @Override
   public boolean shouldFoldLine(@NotNull final String line) {
     // fold Dart VM command line created in DartCommandLineRunningState.createCommandLine() together with the following "Observatory listening on ..." message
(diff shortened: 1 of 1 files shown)


jwren:
LGTM

Thanks for the screenshots.  The next piece here would be something to decode and annotate the error codes from the Dart VM.

alexander-doroshko:
LGTM!

Will it be difficult to do the same without regexp? This regexp looks simple and unlikely could cause performance problems. Just as a rule we try to avoid regular expressions in console filters for performance reasons.

Also, I think we can add some heuristics to make sure that this is a Dart stack frame. For example, it seems that number of spaces + number of chars in the frame index is always 7
```
Stack Overflow
#0      main (file:///C:/dart_projects/untitled7/bin/main.dart:6:5)
#13     main (file:///C:/dart_projects/untitled7/bin/main.dart:7:3)
...
#18408  main (file:///C:/dart_projects/untitled7/bin/main.dart:7:3)
```
Also, we could probably check `contains("dart")`.

devoncarew:
Yup, happy to re-write w/o a regex.

`18,408` is a lot of frames! May adjust this a bit to make sure we collapse stack overflows well.

alexander-doroshko:
> 18,408 is a lot of frames! May adjust this a bit to make sure we collapse stack overflows well.

VM already does it. Ellipsis in my stack is written by the VM, not by me, I removed just a few lines from it :)

devoncarew:
OK, updated. This now:

- no longer uses regexs
- handles the `...` in a stack overflow (it will collapse the whole set of stacks instead of treating it as two separate stacks
- handles the (recently added) text `<asynchronous suspension>`
- will always display the first 8 frames, and start collapsing after that

alexander-doroshko:
Thanks! 
Tracking state in `ConsoleFolding` is a bit fragile, as there are no guarantees about how its methods are called. Fore example `myFrameCount` value becomes unreliable if there is more than one active process emitting console output. Unfortunately I do not know a better solution.

devoncarew:
Ah, good to know re: state and `ConsoleFolding`. If it does become an issue, we could remove the logic for always showing prefixes -

alexander-doroshko:
> we could remove the logic for always showing prefixes -

sorry?

devoncarew:
The state is mostly used for tracking how many trace lines we've seen, so that we can choose to start folding after the first 8 lines (to always show the prefix of the whole stack trace). This is nice, but not critical, so we could lose it if it's causing problems.

alexander-doroshko:
Yup.
I also [added](https://github.com/JetBrains/intellij-plugins/commit/0d3f61373569225e907e421ec3d3a8c783f9b79a) 2 usages of `myFrameCount` to avoid false ellipsis capturing.

PR #482, opened 2017-02-12
change the label of the dart format action; add a keybinding

devoncarew:
- change the label of the dart format action; we use `dartfmt` now, in order to reduce any confusion about which formatter will run dartfmt
- add a keybinding. The normal formatting keybinding on a mac is cmd-alt-L. I looked at using shift-cmd-alt-L for the dart formatter, but unfortunately that binding is already used (to open a formatting dialog). I used ctrl-alt-L; this was not bound to anything, and mirrors the similar organize imports action's binding (ctrl-alt-O). On a mac, the action does respond to the keybinding (and the normal IntelliJ formatter responds to the cmd-alt-L). In the keybinding pres dialog however, I do see that IntelliJ thinks the dartfmt action is bound to both ctrl-alt-L and cmd-alt-L. Not sure how to clear that up.

Hopefully address much of https://youtrack.jetbrains.com/issue/WEB-25401.

screen shot 2017-02-12 at 2 26 23 pm

@jwren @alexander-doroshko

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -231,7 +231,7 @@
             text="Restart Dart Analysis Server" description="Restart Dart Analysis Server">
     </action>
     <action id="Dart.DartStyle" class="com.jetbrains.lang.dart.ide.actions.DartStyleAction"
-            text="Reformat with Dart Style" description="Format your Dart code using the dart_style formatter">
+            text="Reformat Code with dartfmt" description="Format your Dart code using dartfmt (the Dart Style formatter)">
       <add-to-group group-id="CodeFormatGroup" anchor="last"/>
       <add-to-group group-id="EditorPopupMenu" relative-to-action="EditorPopupMenu1" anchor="after"/>
       <add-to-group group-id="ProjectViewPopupMenuModifyGroup" anchor="before" relative-to-action="$Delete"/>
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -234,19 +234,19 @@ dart.sort.members.action.name=Sort Members in Dart File
 dart.sort.members.action.name.ellipsis=Sort Members in Dart files...
 dart.sort.members.action.description=Sort members in your Dart files
 dart.sort.members.hint.failed=Failed to sort members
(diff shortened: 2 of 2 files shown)


alexander-doroshko:
`Ctrl+Alt+L` is the default for standard Reformat Code action on Windows/Linux, so it won't work, sorry.

Having more than 5000 actions in IntelliJ code base it's a big problem to decide which of them should have a shortcut by default and which shortcut would be the best. While there are still some tricky shortcuts unassigned on each OS there should be huge arguments to assign it to a plugin-specific action. With the current Dart plugin adoption, I'm afraid we have to live without any Dart-specific shortcuts out-of-the-box.

Dart Style -> dartfmt renaming looks good to me. Isolated commit would be great.
Cc @pq as I think it was [his suggestion](https://youtrack.jetbrains.com/issue/WEB-14817) for the 'Dart Style' action name. Probably 2 years ago we thought that 'dartfmt' word was going to be replaced with 'Dart Style', but looks like it didn't happen.

devoncarew:
> Having more than 5000 actions in IntelliJ code base it's a big problem to decide which of them should have a shortcut by default

Gotcha, but that does still leave us with the problem that it's more work then it should be to run `dartfmt`. Can we scope the keybinding to just a dart editor? I'm not tied to any particular solution but would like to make it easier to run dartfmt.

jwren:
I do not believe that keybindings can be scoped in IntelliJ.  Do other editors like Atom for instance, have this functionality?

I am fine with the renaming to include dartfmt, but am glad that "Dart Style" is still included, there will be users who call it dart_style because simply because that is what it has been named thus far in the tool.  I have always seen "dartfmt" as the unix/cli name, not the name of the formatter, myself.

As for next steps here.  Currently users have been able to bind a keyboard action to the action if they really want.  They can also setup a macro to run the formatter when the save action runs.  I have been told that both are too cumbersome and won't be used.  It turns out that only power users do these kinds of configurations.

What users really want is for the save action to be automatically be bound to the dartfmt action.  I realize that this is too big of an ask as well @alexander-doroshko, but would it be possible to have an option in the settings to "Automatically auto format with dartfmt on save", and if that is enabled, "Automatically auto sort imports on save", etc?

devoncarew:
OK, I walked this PR back a bit; it no longer includes any keybindings, it just tweaks the text for the dartfmt action.

screen shot 2017-02-26 at 9 19 04 pm

@alexander-doroshko, ptal, thanks!

PR #483, opened 2017-02-13
display casual frames when debugging

devoncarew:
- display casual frames when debugging

This upgrades the service protocol lib to expose the new casual frame info, adds marker frames on the stack at async boundaries, and renders older ('casual') frames in italics - we have location info for these frames, but not variable information.

screen shot 2017-02-12 at 3 30 52 pm

We want a tiny bit more time for the implementation to bake - which is why there's a compile time constant to enable/disable. But I don't expect there to be any big changes.

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
@@ -13,6 +13,7 @@ import com.intellij.xdebugger.evaluation.XDebuggerEvaluator;
 import com.intellij.xdebugger.frame.XExecutionStack;
 import com.intellij.xdebugger.frame.XStackFrame;
 import com.jetbrains.lang.dart.DartFileType;
+import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartAsyncMarkerFrame;
 import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceEvaluator;
 import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceStackFrame;
 import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceValue;
@@ -33,6 +34,9 @@ public class VmServiceWrapper implements Disposable {
 
   public static final Logger LOG = Logger.getInstance(VmServiceWrapper.class.getName());
   private static final long RESPONSE_WAIT_TIMEOUT = 3000; // millis
+
+  // TODO: Remove this compile time flag once the VM implementation and wire protocol are final.
+  private static final boolean RENDER_CAUSAL_FRAMES = true;
 
   private final DartVmServiceDebugProcess myDebugProcess;
   private final VmService myVmService;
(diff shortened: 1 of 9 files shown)


alexander-doroshko:
```
java.lang.NullPointerException
	at org.dartlang.vm.service.element.Frame.getKind(Frame.java:43)
	at com.jetbrains.lang.dart.ide.runner.server.vmService.VmServiceWrapper$7.lambda$received$0(VmServiceWrapper.java:359)
```

devoncarew:
Thanks, I can look into that.

devoncarew:
(updated to fix an NPE)

alexander-doroshko:
Another NPE with today's nightly Dart SDK build :(
```
java.lang.NullPointerException
	at org.dartlang.vm.service.element.CodeRef.getName(CodeRef.java:45)
	at com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceStackFrame.customizePresentation(DartVmServiceStackFrame.java:57)
```

devoncarew:
Ah, the enum names changed

devoncarew:
Updated UI:

screen shot 2017-02-14 at 6 51 54 am

I now see an issue (w/ the recent VM CL) w/ the reported location of async frames. We should hold off on this PR until that's resolved.

devoncarew:
> I now see an issue (w/ the recent VM CL) w/ the reported location of async frames

Nevermind, pilot error on my part. This is good to go (modulo your lgtm!).

alexander-doroshko:
Looks great now!
Is the API still experimental or we can remove the `RENDER_CAUSAL_FRAMES` constant?

BTW we've recently announced [similar feature for Java](https://blog.jetbrains.com/idea/2017/02/intellij-idea-2017-1-eap-extends-debugger-with-async-stacktraces/) :)

devoncarew:
Awesome docs! I wonder if we could get a similar blurb when this work hits?

Re: the availability, I think we want a bit more time for the implementation to bake. The protocol is unlikely to change. I think that the timing for EAP works well for us for this feature - it will give us some burn-in time. /cc @johnmccutchan re: timing and burn-in

alexander-doroshko:
Sure, we'll add a screenshot to the 'New WebStorm EAP is out' blog post when SDK 1.23-dev builds start.

alexander-doroshko:
Pushed!

I also tried an example from [WEB-15765](https://youtrack.jetbrains.com/issue/WEB-15765):
```
import 'dart:async' show Future;
import 'package:stack_trace/stack_trace.dart';

main() {
  Chain.capture(() => new Future(() => someFunc()));
}

someFunc() => new Future(() => someOtherFunc());

someOtherFunc() {
  print(new Chain.current().terse); // breakpoint here
}
```
Any change to get more informative stack for this case too? Currently I get the usual stack (today's SDK 1.23.0-edge.8538d8):
image

devoncarew:
The VM work is based on using the `async` and `await` keywords. It adds hidden local vars to a frame to allow the VM to link back to previous async invocations. As I understand, manual use of Futures and `then()`'s won't get the async stack frame treatment.

devoncarew:
> Sure, we'll add a screenshot to the 'New WebStorm EAP is out' blog post when SDK 1.23-dev builds start.

👍 great!

johnmccutchan:
A couple of bits of context:

1) We are investigating what it would take to track causal async stacks for legacy `.then`, `.catchError` code paths. Support today is restricted to `async` functions. 

2) Running with `package:stack_trace` and native support in the VM results in weird stack traces. `package:stack_trace` will need to be updated to understand the new stack traces the VM outputs.

alexander-doroshko:
Thanks for the info!

Curious, is it technically possible to get local vars / args / `this` fields for the stack frames that are behind the asynchronous gap?

johnmccutchan:
@alexander-doroshko I have future work planned around that.

PR #484, opened 2017-02-15
In the == and hashCode Dart generator, consider only superclasses, not interfaces.

jwren:


alexander-doroshko:
See documentation for [GetTypeHierarchyConsumer.computedHierarchy](https://github.com/JetBrains/intellij-plugins/blob/d0fcca865b56466415f13763cb3c517830c8a708/Dart/thirdPartySrc/analysisServer/com/google/dart/server/GetTypeHierarchyConsumer.java#L38). There are no guarantees about the order of the items in list. It makes no sense to iterate the returned list.

I've found correct superclass iteration in `DartServerTypeHierarchyTreeStructure` and [reused](https://github.com/JetBrains/intellij-plugins/commit/011e716cb651d85757ce6bddd4caabf84f2ba73f) it. I also added a test that uses real Analysis Server.

PR #487, opened 2017-02-24
pass PUB_ENVIRONMENT into pub invocations

devoncarew:
- pass PUB_ENVIRONMENT into pub invocations

@kevmoo @alexander-doroshko @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -148,6 +148,14 @@ public class DartAnalysisServerService implements Disposable {
 
   @NotNull private final EvictingQueue<String> myDebugLog = EvictingQueue.create(DEBUG_LOG_CAPACITY);
 
+  public static String getClientId() {
+    return ApplicationNamesInfo.getInstance().getFullProductName().replaceAll(" ", "-");
+  }
+
+  private static String getClientVersion() {
+    return ApplicationInfo.getInstance().getApiVersion();
+  }
+
   private final AnalysisServerListener myAnalysisServerListener = new AnalysisServerListenerAdapter() {
 
     @Override
@@ -1404,8 +1412,8 @@ public class DartAnalysisServerService implements Disposable {
       myServerSocket =
         new StdioServerSocket(runtimePath, StringUtil.split(vmArgsRaw, " "), analysisServerPath, StringUtil.split(serverArgsRaw, " "),
                               debugStream);
-      myServerSocket.setClientId(ApplicationNamesInfo.getInstance().getFullProductName().replace(' ', '_'));
-      myServerSocket.setClientVersion(ApplicationInfo.getInstance().getApiVersion());
+      myServerSocket.setClientId(getClientId());
+      myServerSocket.setClientVersion(getClientVersion());
 
       final AnalysisServer startedServer = new RemoteAnalysisServerImpl(myServerSocket);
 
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
```diff
@@ -178,6 +178,8 @@ private static void doPerformPubAction(@NotNull final Module module,
 
     try {
       if (ourInProgress.compareAndSet(false, true)) {
+        command.withEnvironment("PUB_ENVIRONMENT", "IntelliJ-IDEA");
```
jwren: I think we want to use what we use for the Analysis Server:

myServerSocket.setClientId(ApplicationNamesInfo.getInstance().getFullProductName().replace(' ', '_'));

Perhaps the DartAnalysisServerService should have a static method getClientId which returns ApplicationNamesInfo.getInstance().getFullProductName().replace(' ', '_') for this location, the location below and for the usage in the analysis server?

Perhaps we should call a method

devoncarew: @kevmoo, do we want a catch-all for anything IntelliJ, or more specific client ids?

kevmoo: More specific is good – assuming it's not a huge set of things.

This will help us understand what tools are being used.

kevmoo: Also: please pass through the IDE process environment, if it exists.

devoncarew:
Makes sense re: the above; will update the PR.

devoncarew:
(updated)

Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
```diff
@@ -64,6 +65,16 @@
 
   private static final AtomicBoolean ourInProgress = new AtomicBoolean(false);
 
+  public static String getPubEnvironmentVariable() {
+    String existingVar = System.getenv("PUB_ENVIRONMENT");
```
kevmoo: I see `PUB_ENVIRONMENT` in a string 3 times – could be [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself)'d up.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -148,6 +148,14 @@
 
   @NotNull private final EvictingQueue<String> myDebugLog = EvictingQueue.create(DEBUG_LOG_CAPACITY);
 
+  public static String getClientId() {
+    return ApplicationNamesInfo.getInstance().getFullProductName().replaceAll(" ", "-");
```
alexander-doroshko: Previously spaces were substituted with an underscore char, now with a hyphen. I do not care, just wanted to make sure it is intentional.

Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
```diff
@@ -59,11 +60,23 @@
...
 
   private static final AtomicBoolean ourInProgress = new AtomicBoolean(false);
 
+  public static String getPubEnvValue() {
+    String existingVar = System.getenv(PUB_ENV_VAR_NAME);
+    if (existingVar == null) {
+      return DartAnalysisServerService.getClientId();
+    }
+    else {
+      return existingVar + ":" + DartAnalysisServerService.getClientId();
```
alexander-doroshko: Path separator is different on Windows. I guess you mean `File.pathSeparatorChar`.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
devoncarew: Yup, intentional :)

Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java
devoncarew: I don't think so; this is to create a new, longer env var token, so that if we're running in another context - like on a travis bot - we get a different token. It could then we used to differentiate between user usage and bot usage.

(so the separator token doesn't have much meaning here, just something to concatenate with)

PR #488, opened 2017-02-24
badge the analysis view when the analyzer is busy

devoncarew:
We've had some feedback that the progress view when analyzing is too noisy. This is specifically when used with the new analysis driver, which - given changes to some highly referenced libraries - can re-compute the error state for large sets of libraries. This looks to the user like making small changes in some files produces a lot of analysis.

Some tweaks to how we present status to address this:
- badge the analysis view when the analyzer is busy; this uses the same mechanism as when the `Run` or `Debug` tool windows are active
- we show the first analysis reported by the analysis server using IntelliJ's progress mechanism; later progress is ignored. We reset state every time the analysis server restarts.

screen shot 2017-02-24 at 12 20 29 pm

@alexander-doroshko @jwren


devoncarew:
(also /cc @scheglov)

jwren:
LGTM

PR #489, opened 2017-02-25
add a dart drop frame action

devoncarew:
- add a dart drop frame action
- fix https://youtrack.jetbrains.com/issue/WEB-25346

screen shot 2017-02-24 at 4 11 48 pm

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/actions/DartPopFrameAction.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/actions/DartPopFrameAction.java
@@ -1,0 +1,51 @@
+
+package com.jetbrains.lang.dart.ide.runner.actions;
+
+import com.intellij.icons.AllIcons;
+import com.intellij.openapi.actionSystem.*;
+import com.intellij.openapi.project.DumbAware;
+import com.jetbrains.lang.dart.DartBundle;
+import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceStackFrame;
+import org.jetbrains.annotations.NotNull;
+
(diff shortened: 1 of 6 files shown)


devoncarew:
(updated to handle async frames)

Dart/src/com/jetbrains/lang/dart/ide/runner/actions/DartPopFrameAction.java
```diff
@@ -0,0 +1,51 @@
...
+    if (ActionPlaces.isMainMenuOrActionSearch(e.getPlace()) || ActionPlaces.DEBUGGER_TOOLBAR.equals(e.getPlace())) {
+      e.getPresentation().setEnabled(enabled);
+    }
+    else {
+      e.getPresentation().setVisible(enabled);
+    }
+  }
+
+  static DartVmServiceStackFrame getStackFrame(AnActionEvent e) {
+    Component component = PlatformDataKeys.CONTEXT_COMPONENT.getData(e.getDataContext());
```
alexander-doroshko: Searching for the stack frame from context this way doesn't work if the frames list is not focused. For example, if a user is exploring Variables View. Need to do similar to Java's action.

devoncarew: Ah, good to know. I had some trouble getting the current frame using the technique the Java action did; perhaps because of how this action is contributed?

alexander-doroshko: Truly, I didn't check Java's action path. But we should find a way to fix ours.

PR #491, opened 2017-02-27
Show a tooltip w/ analysis errors on run

devoncarew:
When running a Dart app, if there are analysis errors in the containing module, show a tooltip on the Dart Analysis view telling the users about the issues.

screen shot 2017-02-26 at 8 43 30 pm

The user can enable / disable this by adding or removing the Validate Dart analysis pre-launch task:

screen shot 2017-02-26 at 8 43 05 pm

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -205,6 +205,7 @@
     <runConfigurationProducer implementation="com.jetbrains.lang.dart.ide.runner.test.DartTestRunConfigurationProducer"/>
 
     <programRunner implementation="com.jetbrains.lang.dart.ide.runner.DartRunner"/>
+    <stepsBeforeRunProvider implementation="com.jetbrains.lang.dart.ide.runner.DartValidateBeforeRunTaskProvider" id="dartValidateBeforeRun" />
 
     <localInspection bundle="com.jetbrains.lang.dart.DartBundle" key="outdated.dependencies.inspection.name"
                      groupName="Dart" enabledByDefault="true" level="WARNING" language="Dart"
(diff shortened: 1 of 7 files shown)


alexander-doroshko:
Sorry for the delay. I'm thinking.

devoncarew:
No worries, I'm happy to finish this up, just wanted to put up my 95% solution before I went on break.

devoncarew:
@alexander-doroshko, this is ready for review now! I'll squash this into one commit.

devoncarew:
This needs to be reviewed to verify that we are adding this to a new launch config. A possible resolution if we're not is to add the validation to the default Dart launch config.

devoncarew:
We could also just always show a balloon, with some links to disable more permanently if the user chooses.

alexander-doroshko:
We've discussed the feature internally and found a separate Before Run Task to be a too heavy solution.
Showing tool window balloon without a check box / before run task is probably affordable. For those who want to disable it - there's already a way to control all notifications like this. The feature is not well-known though, it is in Settings menu of the Event Log tool window. We may add links like 'Dismiss for this session' or 'Dismiss forever' or both to the balloon. The latter could programmatically trigger corresponding Event Log setting.

One more consideration: we could use a more accurate scope of files for which we care about errors. For example when a file from bin folder is launched we care about errors from lib and bin folders, but not from test, tool, examples, etc. Some similar scope calculation is done in `DartResolveScopeProvider`, may be it will be possible to reuse it.

devoncarew:
Thanks, will close this PR and work on a different one. My takeaways are that:
- we don't need to use the before run task functionality
- we should just show a balloon on every run if there are errors
- there are existing framework mechanisms to make sure the balloons can be filtered
- we could choose to put a `don't show again` link in the balloon
- we could be more precise about which errors apply to the app

PR #494, opened 2017-03-01
Update Dart Analysis Server sources.

pq:
* picks up added `defaultArgumentListString` property in the new `CompletionSuggestion` API

@alexander-doroshko

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
+++ b/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
@@ -122,7 +122,7 @@ public class AnalysisError {
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
-    String code = jsonObject.get("code").getAsString();
+    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
     Boolean hasFix = jsonObject.get("hasFix") == null ? null : jsonObject.get("hasFix").getAsBoolean();
     return new AnalysisError(severity, type, location, message, correction, code, hasFix);
   }
--- a/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
+++ b/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
@@ -122,7 +122,7 @@ public class AnalysisError {
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
-    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
+    String code = jsonObject.get("code").getAsString();
     Boolean hasFix = jsonObject.get("hasFix") == null ? null : jsonObject.get("hasFix").getAsBoolean();
     return new AnalysisError(severity, type, location, message, correction, code, hasFix);
   }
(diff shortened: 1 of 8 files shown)


Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
```diff
@@ -122,7 +122,7 @@ public static AnalysisError fromJson(JsonObject jsonObject) {
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
-    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
+    String code = jsonObject.get("code").getAsString();
```
alexander-doroshko: This will throw NPE with older SDKs

pq: OK, will fix.

pq:
It looks like I missed a few API changes...  Stand-by for update.

alexander-doroshko:
The rest looks safe.
I guess these sources are generated. Is it possible to fix anything somewhere to make it generate backward-compatible code?

pq:
> It looks like I missed a few API changes... Stand-by for update.

Done.

> Is it possible to fix anything somewhere to make it generate backward-compatible code?

Great question.

cc @scheglov @bwilkerson @jwren for thoughts.

bwilkerson:
> Is it possible to fix anything somewhere to make it generate backward-compatible code?

Maybe. We only have one version of the spec to work from when generating, so that will limit us to some extent, but depending on what you need to do it might be enough.

There is also an opportunity to have it generate more Java code so that we need to do less when the spec is updated. Jaime and I discussed this when we were first writing the generator, but the work was not finished.

The code to generate the Java support code is in these two files:
https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/spec/codegen_java.dart
https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/spec/codegen_java_types.dart

alexander-doroshko:
> Is it possible to fix anything somewhere to make it generate backward-compatible code?

For the reference: I was asking about [this code](https://github.com/dart-lang/sdk/blob/66b5248bf937ef1202f04b106c63730e57a008b8/pkg/analysis_server/tool/spec/generated/java/types/RequestError.java#L83) that throws NPE with older SDKs
```
  public static RequestError fromJson(JsonObject jsonObject) {
    String code = jsonObject.get("code").getAsString();
    ...
  }
```

bwilkerson:
Then, yes, it should be possible to fix the way it generates field accesses.

devoncarew:
Perhaps just marking `code` as optional?

bwilkerson:
The field 'code' is not optional and it never has been. Marking it as optional would be the wrong solution (and a breaking change to the API).

@alexander-doroshko Have we run into a situation where server sent back a RequestError that did not have a code associated with it? If so, we need to fix the server. Either way, should we just always be defensive (always check for null, even when the field is required)?

alexander-doroshko:
@bwilkerson Sorry, I gave a wrong link. I meant [AnalysisError.code](https://github.com/dart-lang/sdk/blob/66b5248bf937ef1202f04b106c63730e57a008b8/pkg/analysis_server/tool/spec/generated/java/types/AnalysisError.java#L125) introduced in [this commit](https://github.com/dart-lang/sdk/commit/e7e205f00b24faf9ee5cdf0baf21b8a0a04f8c3a). Older SDKs do not have analysis error code.

bwilkerson:
Ok. Then for backward compatibility we should always check for null, even for required fields, because they might not have been required in previous versions. That seems reasonable to me.

alexander-doroshko:
We've already had [this discussion](https://github.com/JetBrains/intellij-plugins/pull/399/files/efe77c62f84dd2fcfe71426d4d5f40cba8a2ba55#r75862948) half a year ago and at that point decided just to stay with a couple of manually modified lines :).

bwilkerson:
I had forgotten that discussion. However, it might be worth reconsidering. Seems like that decision is leading to errors and that it might be safer to modify the generator. But I'm fine with either decision.

alexander-doroshko:
Merged

pq:
Thanks!

PR #495, opened 2017-03-02
Support for linked edits of default args in Dart completions.

pq:
* updates to the latest version of the `CompletionSuggestion` API
* adds a template builder and linked edits for default argument lists (when present)

See: https://github.com/flutter/flutter-intellij/issues/553

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -5,6 +5,9 @@ import com.intellij.codeInsight.completion.*;
 import com.intellij.codeInsight.completion.util.ParenthesesInsertHandler;
 import com.intellij.codeInsight.lookup.LookupElement;
 import com.intellij.codeInsight.lookup.LookupElementBuilder;
+import com.intellij.codeInsight.template.TemplateBuilderFactory;
+import com.intellij.codeInsight.template.TemplateBuilderImpl;
+import com.intellij.codeInsight.template.impl.TextExpression;
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.highlighter.HtmlFileType;
 import com.intellij.injected.editor.VirtualFileWindow;
@@ -12,6 +15,7 @@ import com.intellij.lang.Language;
 import com.intellij.lang.html.HTMLLanguage;
 import com.intellij.lang.injection.InjectedLanguageManager;
 import com.intellij.lang.xml.XMLLanguage;
+import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Pair;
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -347,4 +349,25 @@ else if (elementKind.equals(ElementKind.TOP_LEVEL_VARIABLE)) {
       return null;
     }
   }
+
+  @Override
+  public void beforeCompletion(@NotNull final CompletionInitializationContext context) {
```
alexander-doroshko: As far as I see this method has been moved without any changes inside. Has it been done on purpose? I'd prefer to avoid such changes as they pollute code annotation (I mean right-click gutter -> Annotate) and make it harder to track file history.

pq: I just invoked the formatter (which maybe is configured to re-arrange?)  I thought this was the canonical thing to do but maybe I don't have the right settings?

alexander-doroshko: Ok, clear. We do not have the policy to rearrange the code before commit, only to reformat. To switch it off invoke Cmd+Alt+Shift+L, uncheck Rearrange, press Enter. Next time you may use standard dialog-less Cmd+Alt+L.

pq: Updated settings.  Thanks!

pq: And sorry for the added noise!

alexander-doroshko:
The change seems to break `DartServerCompletionTest.testFunctionWithArgsInvocation`. Other tests in `DartServerCompletionTest` blink sometimes but I do not know whether it is caused by this change. Can you take a look? Adding a test for the new functionality would also be great if possible.

pq:
Quick update: one test is fixed but I'm puzzling through `testFunctionAsArgument`.  It passes when run solo but fails consistently in the suite which makes me think I've introduced a timing issue. Investigating...

alexander-doroshko:
> It passes when run solo but fails consistently in the suite

The same without your change. No idea so far.

pq:
> The same without your change. No idea so far.

Was just going to say that. :)  It's a puzzle!

pq:
FWIW: I've confirmed that this does appear to do the right think in the IDE so it may be more a matter of hardening the test.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -264,19 +243,48 @@ private static LookupElement createLookupElement(@NotNull final Project project,
...
+              ParenthesesInsertHandler.WITH_PARAMETERS.handleInsert(context, item);
+              // Show parameters popup.
+              final Editor editor = context.getEditor();
+              final PsiElement psiElement = lookupObject.getElement();
+
+              // Insert argument defaults if provided.
+              final String argumentListString = suggestion.getDefaultArgumentListString();
+              if (argumentListString != null) {
+                final Document document = editor.getDocument();
+                int offset = context.getSelectionEndOffset() - 1;
```
alexander-doroshko: At this point caret is already between parens, you can take `editor.getCaretModel().getOffset();`.
With the current code it is possible to get malformed code when completing over the existing text using Tab:
image

alexander-doroshko:
@pq There's one case that should be thought over and fixed. When a user is completing using **Tab** over the existing method call with parameters. What would you expect to happen?
image

pq:
Thanks @alexander-doroshko .  To be honest, I'm not sure what should happen here (except to say that I'm sure that what I'm currently doing is NOT ideal)!  Incidentally, Java doesn't handle this particularly well either.  Do you have any thoughts on the most consistent (and unsurprising) behavior?

alexander-doroshko:
I see 2 ways: replace all old arguments with the new live template or keep old arguments and skip inserting args placeholder. In Java there is no live template with arguments placeholders, so it keeps old args and places the caret after opening paren. We could probably do the same. Do you think it could be better in Java?

pq:
I'm not sure what would be better for Java but note that it's not useful to invoke a completion from within a method invocation.  I wonder what would make it more useful?  And in general, what folks expect to have happen in this case?  In the short term, I think I favor:

> keep old arguments and skip inserting args placeholder

What's the best way to detect we're in this case?  I'm guessing I could map the caret location back to see if we're in a `PsiElement`?

Thanks for the help!

alexander-doroshko:
>What's the best way to detect we're in this case?

`char at caret != ')' ` seems to work perfectly, see `ParenthesesInsertHandler.WITH_PARAMETERS`

pq:
Sorry for the slow follow-up on this.  I did some experimenting and it looks like the more ideal behavior would require a bit of doing (unless I'm missing a shortcut).  It also looks like, while odd, the current behavior is consistent with what happens in Java.

screen shot 2017-03-07 at 9 02 21 am
screen shot 2017-03-07 at 9 02 27 am

Do you object strongly to landing it as is or do you see any easy improvements?

Thanks in advance!

alexander-doroshko:
These screenshots are for Enter, no problems with it. My concerns were about Tab :)

alexander-doroshko:
Merged

pq:
\o/

Thank you @alexander-doroshko !

alexander-doroshko:
@pq We can't be sure that 100% of users will like the feature. Would be great to add an option for it in the Smart Keys page. As an example see `AngularJSConfigurationPage`, `WebEditorOptionsProvider` and other usages of `EditorSmartKeysConfigurableEP`.
image

pq:
SGTM.

FYI @devoncarew @mit-mit

PR #496, opened 2017-03-10
Add surround-with-brackets for named args (to support Flutter)

stevemessick:
@alexander-doroshko I limited this to only being available when the selected expression is a named argument. It *will* make code red, unless it is already red because the user changed `child:` to `children:` first. And yes, we could be even more specific and check the name of the argument but I don't want to be too restrictive -- I don't have enough Flutter expertise to know if it could be useful for other arguments.

The one thing I don't like about this currently is that it adds bracket-surrounding at position #3 (for me). I'd prefer it at the bottom so it doesn't change the existing order.

/cc @devoncarew @pq

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/surroundWith/expression/DartWithBracketsExpressionSurrounder.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/surroundWith/expression/DartWithBracketsExpressionSurrounder.java
@@ -1,0 +1,33 @@
+package com.jetbrains.lang.dart.ide.surroundWith.expression;
+
+import com.intellij.psi.PsiElement;
+import com.jetbrains.lang.dart.psi.DartNamedArgument;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class DartWithBracketsExpressionSurrounder extends DartWithExpressionSurrounder {
+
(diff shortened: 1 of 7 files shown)


Dart/src/com/jetbrains/lang/dart/ide/surroundWith/expression/DartWithBracketsExpressionSurrounder.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Override
+  protected String getTemplateText(PsiElement expr) {
+    String text = expr.getText();
+    int nlIndex = text.lastIndexOf('\n');
+    if (nlIndex < 0) {
+      return "[" + expr.getText() + "]";
+    }
+    Matcher matcher = Pattern.compile("\\n", Pattern.MULTILINE).matcher(text);
+    String newText = matcher.replaceAll("\n  ");
```
alexander-doroshko: Indents handling seems to be not needed, IDE auto-reformats the code after the action. I [simplified](https://github.com/JetBrains/intellij-plugins/commit/60bdb59e08d2b4e8584a3d8bd0cd82022a998dd1) this and tests still pass. Let me know if I missed something.

stevemessick: Simpler code is always better, thanks. I'll try to remember `textContains()`, too.

PR #498, opened 2017-03-22
make the errors/warnings/hints display more consistent

devoncarew:
- make the errors/warnings/hints display more consistent

before (error / warning / hint):

screen shot 2017-03-22 at 12 40 30 pm

after:

screen shot 2017-03-22 at 12 39 36 pm

- we now show warnings with a yellow underline
- hints have a similar presentation to dart errors and warnings (an underline, no background color changes)

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/highlight/DartSyntaxHighlighterColors.java
+++ b/Dart/src/com/jetbrains/lang/dart/highlight/DartSyntaxHighlighterColors.java
@@ -91,9 +91,9 @@ public class DartSyntaxHighlighterColors {
   public static final TextAttributesKey ERROR =
     createTextAttributesKey(DART_ERROR, CodeInsightColors.ERRORS_ATTRIBUTES);
   public static final TextAttributesKey WARNING =
-    createTextAttributesKey(DART_WARNING, CodeInsightColors.WEAK_WARNING_ATTRIBUTES);
+    createTextAttributesKey(DART_WARNING, CodeInsightColors.WARNINGS_ATTRIBUTES);
   public static final TextAttributesKey HINT =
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
- `WEAK_WARNING_ATTRIBUTES` are too subtle in the default color scheme
- `NOT_USED_ELEMENT_ATTRIBUTES` not only add an underline, text foreground is also affected. And in default color scheme ONLY foreground changes, there's no underline. Hints become looking like a comment or unused code.

Here's before and after from Dart's Colors&Fonts page in Preferences:
image

Also there are other color themes, both bundled and also dozens of custom ones. I'm pretty sure that transition from `highlight like an error` to `highlight like a weak warning`  and from `warning` to `unused symbol` would be very unexpected to users.

devoncarew:
Hmm, OK, updated to map errors to errors, warnings to warnings, and hints to weak warnings.

This gives us the benefit of distinguishing between errors and warnings in the text (which we weren't doing before), and of hints having slightly less visual emphasis than warnings. Here's the preference page w/ this PR:

screen shot 2017-03-23 at 1 38 18 pm

screen shot 2017-03-23 at 1 38 26 pm

devoncarew:
@alexander-doroshko, I discussed this with @lukechurch in person. He is ok with this change (errors shown as errors, warnings as warnings) given the upcoming change to strong mode as the default. Happy to talk more by email or VC if you like.

This PR has been out for awhile; not sure if it can land as is or needs to be updated / redone.

alexander-doroshko:
Ok, pushed manually. Text also updated, PTAL: https://github.com/JetBrains/intellij-plugins/commit/b3fd9acaa697af69eac962ec2e488eb314ecbd25

devoncarew:
Looks great!

PR #499, opened 2017-03-22
General cleanup in Dart plugin test package com.jetbrains.dart.analysisServer

jwren:
Add missing overrides in com.jetbrains.dart.analysisServer, move setUp(), getBasePath() and getTestDataPath() to top of classes, and in that order.


jwren:
@alexander-doroshko

PR #500, opened 2017-03-24
In the Dart Plugin, add AnalysisError.correction information into the Dart Analysis panel.

jwren:
screenshot from 2017-03-24 13 03 44

@devoncarew @bwilkerson @alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
@@ -4,6 +4,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.registry.Registry;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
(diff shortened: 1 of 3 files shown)


devoncarew:
This takes up a lot of visual space, and the info is less important than the description and location. Are there other ways we could present this info? Perhaps in the status line when an error is selected?

bwilkerson:
Given that some of our errors are now being formatted with multiple lines, perhaps a pop-up when the user hovers over a message that contains both the whole error message (formatted nicely) and the correction message.

alexander-doroshko:
@bwilkerson Do yu have examples of multiline errors? Can `'\n'` be simply replaced with space for them?

Agree that we should try to make correction not to take much (any?) of precious space. Raw ideas: show it as a tooltip of the error/warning/hint icon; or as a tooltip of the whole line; or add a new column of icons with correction shown on hover. More ideas welcome.

bwilkerson:
The new error messages are related to type inference, which tends to be hard for people to understand. I'm not sure whether they're actually committed yet, but here's an example of what one looks like:
```
Couldn't infer type parameter 'T extends MenuItem<T>'.

Tried to infer 'MenuItem<dynamic>' for 'T extends MenuItem<T>' which doesn't work:
  Type parameter 'T extends MenuItem<T>' declared to extend 'MenuItem<MenuItem<dynamic>>'.
The type 'MenuItem<dynamic>' was inferred from:
  Parameter 'subMenu' declared as     'MenuModel<T>'
                      but argument is 'MenuModel<MenuItem<dynamic>>'.

Consider passing explicit type argument(s) to the generic.
```
We could replace the newlines with spaces, but that's not going to be readable a a single line. I'm guessing that the best approach will be to use the first line of the message in the Dart Analysis view and have a way of seeing the rest of the message.

alexander-doroshko:
>I'm guessing the best approach will be to use the first line of the message in the Dart Analysis view and have a way of seeing the rest of the message

Or we can keep protocol as is and handle it at the client side and take your suggestion: show the first line in the table, full text (probably together with correction) on hover.

bwilkerson:
Yes, that's what I was suggesting, so it sounds like we agree. :-)

jwren:
Follow up with PR with information in tooltip is here: https://github.com/JetBrains/intellij-plugins/pull/501

PR #501, opened 2017-03-27
In the Dart Plugin, add AnalysisError.correction information into the tooltip of the Dart Analysis errors and warnings

jwren:
I made an attempt at adding the information into the tooltips in the DartAnnotator, but the AnnotationHolderImpl makes a call to XmlStringUtil.escapeString(message) which removes any added '/n' or '\<br/\>'.

two


jwren:
https://youtrack.jetbrains.com/issue/WEB-23801

alexander-doroshko:
Thanks, merged! 
BTW did you have a chance to test it with a real multiline message?

alexander-doroshko:
Hm, it seems to be not working if the error text is wider than the cell:
image

alexander-doroshko:
@jwren can you take a look?

jwren:
I took a look but came up short, this behavior comes from within JLabel & JComponent.  There to not appear to be any places to easily override the behavior, although I did try.

PR #502, opened 2017-03-30
tweaks to dartdoc hover rendering

devoncarew:
- remove double blank lines from the the dartdoc tooltip text
- remove the leading `Signature` text
- convert symbol references to markdown code format (`[foo]` ==> ``` `foo` ```)

@alexander-doroshko 

screen shot 2017-03-29 at 1 28 56 pm

screen shot 2017-03-29 at 6 11 16 pm

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -80,7 +80,6 @@ public class DartDocUtil {
     final StringBuilder builder = new StringBuilder();
     builder.append("<code>");
     if (signature != null) {
-      builder.append("<b>Signature:</b> ");
       if (signatureIsHtml) {
         builder.append(signature);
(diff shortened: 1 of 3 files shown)


devoncarew:
It looks like this does break several unit tests - taking a look.

devoncarew:
- added fixes for DartServerDocUtilTest and DartDocUtilTest

alexander-doroshko:
Brackets are sometimes a part of hyperlink markup. With this change, we loose for example a link in the `DateTime` class, its last line is 
```
the [intl](http://pub.dartlang.org/packages/intl) package.
```

Also smth is wrong with `Object` docs:
image

devoncarew:
OK, will update the regex.

devoncarew:
Updated. I removed the handling of `[foo]`, but kept some of the whitespace improvements. It looks like the markdown library in use does not correctly handle some `[foo](links)` in any case - the rendering of Object doesn't look correct on master.

PR #504, opened 2017-03-30
Add support for edit.getStatementCompletion

stevemessick:
@alexander-doroshko This is the beginning of the Complete Statement work. The server modifications are pretty straight-forward. I would appreciate you verifying that the call to `awaitForLatchCheckingCanceled()` is correct. I'm just guessing that the last argument (`false`) is needed to stop error checking. The editor hook-up is basic at this point. No support for identifier completion terminated with statement completion has been defined yet. However, we already know one case where that would be useful, so I'm planning to do that eventually.

This line needs to be added to plugin.xml to make it work:
```xml
    <lang.smartEnterProcessor language="Dart"
                              implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerStatementCompletionProcessor"/>
```

I have not committed the DAS code to do recognize the edit.getStatementCompletion request yet.

I have tested that the round-trip works, using a fixed response from the server.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -949,6 +949,29 @@ public class DartAnalysisServerService implements Disposable {
     return results;
   }
 
+  public SourceChange edit_getStatementCompletion(@NotNull final VirtualFile file, final int _offset) {
+    final String filePath = FileUtil.toSystemDependentName(file.getPath());
+
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return null;
+    }
+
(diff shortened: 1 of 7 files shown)


alexander-doroshko:
The last argument of `awaitForLatchCheckingCanceled` must be `true`, or (better) use its 3-parameters version. Otherwise, there's a risk to suspend server events processing and hang the EDT for the whole `timeoutInMillis` period. The only place where `false` is passed is the one where I haven't figured out how to rewrite it yet, it needs to be fixed.
I'll take a look at the rest of the code later, sorry, not feeling well today.

alexander-doroshko:
Sorry for delay. The rest looks good if we assume that it is ok to block the EDT for up to 100ms. The trick with `ProcessCancelledException` doesn't work for this case so we unblock either on server response or on timeout.

PR #506, opened 2017-03-31
add support for awaiter frame traces

devoncarew:
- add support for awaiter frame traces
- prefer displaying awaiter traces if available; then prefer a sync causal traces, else show regular vm (sync) traces

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
@@ -37,9 +37,6 @@ public class VmServiceWrapper implements Disposable {
   public static final Logger LOG = Logger.getInstance(VmServiceWrapper.class.getName());
   private static final long RESPONSE_WAIT_TIMEOUT = 3000; // millis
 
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java
```diff
@@ -369,16 +366,13 @@ public void received(final Stack vmStack) {
         ApplicationManager.getApplication().executeOnPooledThread(() -> {
           InstanceRef exceptionToAddToFrame = exception;
 
-          ElementList<Frame> elementList;
-          if (RENDER_CAUSAL_FRAMES) {
+          ElementList<Frame> elementList = vmStack.getAwaiterFrames();
```
jwren: With the VM protocol, do we check that the user is at some version of the SDK, or just assume that they are?

devoncarew: For this particular change, older versions of the protocol won't send in await frames (or async causal frames). Those fields will just not be there when we go to get them out of the json. So we can degrade gracefully here w/o version checks.

(similarly, even VMs which do support sending the new frames types don't send at paused locations where they're not necessary - purely sync code)

PR #513, opened 2017-04-10
Stmt complete

stevemessick:
@alexander-doroshko This hooks up statement completion. The latest dev-branch of the SDK includes support for it.

There are two changes relative to what you saw before. First, obviously, the plugin.xml change is included. Second, it actually sets the editing cursor to the location given in the completion result, something I forgot to do earlier.

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -160,6 +160,8 @@
     <completion.contributor language="Dart" implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerCompletionContributor"/>
     <completion.contributor language="HTML" implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerCompletionContributor"/>
     <lookup.charFilter implementation="com.jetbrains.lang.dart.ide.completion.DartCharFilter"/>
+    <lang.smartEnterProcessor language="Dart"
+                              implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerStatementCompletionProcessor"/>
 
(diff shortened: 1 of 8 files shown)


alexander-doroshko:
Is there anything server can already do? If so adding a test would be great. `myFixture.performEditorAction(IdeActions.ACTION_EDITOR_COMPLETE_STATEMENT)` will help to do the trick.

Is it true that there's no way to call server outside of the write action? Then before landing this PR I need to refactor `DartAnalysisServerService` a bit. As for now, there's a huge risk to ignore server response and hang UI for the whole timeout. It happens if server sends `computedErrors` notification between request and response, and with the new analysis driver, it is very likely.

stevemessick:
I can add a test. It will work if it is run using the latest 1.24.0-edge... SDK. Should I bump the analyzer version so the test can check if the SDK supports the new request?

I'm fine with holding this PR open until the 1.24 DEV build is out. I need it for my own testing but the service isn't ready for general use yet.

> Is it true that there's no way to call server outside of the write action?

I'm not sure what you are asking. Do you want this to be a two-stage operation, where it requests the action when the user types s-c-enter, then processes the result asynchronously? I don't know if the editor framework supports that. How would it work in identifier completion? (I have not implemented that yet, but intend to.)

alexander-doroshko:
> run using the latest 1.24.0-edge

That's ok. I run these tests only locally and I know how to interpret false failures.

> I'm not sure what you are asking.

Ok, I see that `SmartEnterProcessor.process()` is called inside a write action and see no way to avoid it. Two-stage async operation doesn't sound good to me. So we have to review all server response and notification handlers and make sure they do not take write action. `onErrorsUpdated()` definitely needs to be rewritten. I can take a look few days later.

PR #516, opened 2017-04-11
Stmt complete

stevemessick:
@alexander-doroshko Here's the statement completion support with a couple tests. We still have the write action issue to deal with.

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -160,6 +160,8 @@
     <completion.contributor language="Dart" implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerCompletionContributor"/>
     <completion.contributor language="HTML" implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerCompletionContributor"/>
     <lookup.charFilter implementation="com.jetbrains.lang.dart.ide.completion.DartCharFilter"/>
+    <lang.smartEnterProcessor language="Dart"
+                              implementationClass="com.jetbrains.lang.dart.ide.completion.DartServerStatementCompletionProcessor"/>
 
     <resolveScopeProvider implementation="com.jetbrains.lang.dart.resolve.DartResolveScopeProvider"/>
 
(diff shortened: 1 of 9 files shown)


alexander-doroshko:
Thanks for the tests!
Let's suspend until read/write locks story is refactored.

alexander-doroshko:
Sorry for delay, pushed now!

A couple of notes:
- it seems to work only if `if` or `while` is the last statement in the method. Otherwise, I guess the following statement is parsed as `if`-body and server decides not to add braces. I think it should add braces anyway and make the statement to be outside of `if`. In Java it does this way.
- I noticed that sometimes server returns `SourceChange` with zero `SourceFileEdits` or with one `SourceFileEdit` that has a single `SourceEdit` with **empty string** as `replacement`. In other words, `noop` is sometimes received. Is it expected or should be fixed at server side? If expected then it should be handled at client side and in case of `noop` `DartServerStatementCompletionProcessor.process()` must return `false` so that the action is handled elsewhere.
- Idea about the API. Is it possible that SourceChange contains more than one SourceFileEdit? I guess no, it is not expected to touch any file except the current one. Maybe change `GetStatementCompletionConsumer#computedSourceChange` to `GetStatementCompletionConsumer#computedEdit` like we already have, for example, in `OrganizeDirectivesConsumer#computedEdit`?

stevemessick:
Thanks, Alex! Yes, at this time the code relies on parsing errors to determine what completion to offer. This needs a lot of work, but I'm focused on getting the editing in place right now. Eventually, there should only be a noop if the code encounters an error, but I haven't got that far yet, so I'm not surprised you're seeing that. Thanks for the tip about `OrganizeCirectivesConsumer`; I'll take a look.

stevemessick:
@alexander-doroshko SourceChange includes the position of the cursor after the edit has been applied. SourceFileEdit does not. So, we need to keep using SourceChange since positioning the cursor is one of the key features of this operation.

alexander-doroshko:
Sounds reasonable.

PR #517, opened 2017-04-11
Initial Dart Web App launch configuration, currently commented out

jwren:


Dart/src/com/jetbrains/lang/dart/ide/runner/web/DartWebRunConfigurationType.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  protected DartWebRunConfigurationType() {
+    super("DartWebRunConfigurationType", DartBundle.message("runner.web.configuration.name"),
+          DartBundle.message("runner.web.configuration.description"),
+          DartIcons.Dart_web);
+    addFactory(new ConfigurationFactory(this) {
+      @Override
+      public String getName() {
+        return "Dart Web App"; // compatibility
```
alexander-doroshko: Compatibility? ;).

alexander-doroshko:
> currently commented out

doesn't look like this :)

Anyway, it is too early to include it in the Dart plugin. Need to understand what we want first.

alexander-doroshko:
Looks like we can close this.

jwren:
Yes, thanks.

PR #519, opened 2017-04-18
show errors when running a Dart app

devoncarew:
Show errors when running a Dart app:

- add functionality to Dart launches to display errors on run. We display tool window notifications on the Dart Analysis view and focus an error in the editor
- the user can set an option in the preferences to not show these notifications
- the code is written in such a way that the general API is available to other plugins (are there errors for a launch / display the errors for a launch)

screen shot 2017-04-18 at 11 59 17 am

screen shot 2017-04-18 at 11 51 34 am

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -45,6 +45,7 @@ import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFileSystemItem;
 import com.intellij.psi.search.FilenameIndex;
+import com.intellij.psi.search.SearchScope;
 import com.intellij.util.Alarm;
 import com.intellij.util.Consumer;
 import com.intellij.util.PathUtil;
@@ -612,6 +613,14 @@ public class DartAnalysisServerService implements Disposable {
   @NotNull
   public List<DartServerData.DartError> getErrors(@NotNull final VirtualFile file) {
     return myServerData.getErrors(file);
+  }
+
+  public List<DartServerData.DartError> getErrors(@NotNull final Module module) {
+    return myServerData.getErrors(module);
+  }
+
+  public List<DartServerData.DartError> getErrors(@NotNull final SearchScope scope) {
+    return myServerData.getErrors(scope);
   }
 
   @NotNull
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
@@ -6,11 +6,14 @@ import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleUtilCore;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.impl.source.resolve.ResolveCache;
+import com.intellij.psi.search.SearchScope;
 import com.intellij.util.SmartList;
 import gnu.trove.THashMap;
 import org.dartlang.analysis.server.protocol.*;
@@ -178,6 +181,34 @@ public class DartServerData {
     if (hasChanges) {
       forceFileAnnotation(file, false);
     }
+  }
+
+  @NotNull
+  List<DartError> getErrors(@NotNull final Module module) {
+    final List<DartError> errors = new ArrayList<>();
+
+    for (String path : myErrorData.keySet()) {
+      final VirtualFile file = LocalFileSystem.getInstance().findFileByPath(path);
+      if (file != null && ModuleUtilCore.moduleContainsFile(module, file, false)) {
+        errors.addAll(myErrorData.get(path));
+      }
+    }
+
+    return errors;
+  }
+
+  @NotNull
+  List<DartError> getErrors(@NotNull final SearchScope scope) {
+    final List<DartError> errors = new ArrayList<>();
+
+    for (String path : myErrorData.keySet()) {
+      final VirtualFile file = LocalFileSystem.getInstance().findFileByPath(path);
+      if (file != null && scope.contains(file)) {
+        errors.addAll(myErrorData.get(path));
+      }
+    }
+
+    return errors;
   }
 
   @NotNull
@@ -398,7 +429,6 @@ public class DartServerData {
     }
   }
 
-
   public static class DartRegion {
     protected int myOffset;
     protected int myLength;
@@ -460,6 +490,10 @@ public class DartServerData {
 
     public String getSeverity() {
       return mySeverity;
+    }
+
+    public boolean isError() {
+      return mySeverity == "ERROR";
     }
 
     @Nullable
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
@@ -19,6 +19,10 @@ import com.intellij.execution.runners.ExecutionUtil;
 import com.intellij.ide.projectView.ProjectView;
 import com.intellij.ide.projectView.impl.ProjectViewPane;
 import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationGroup;
+import com.intellij.notification.NotificationListener;
+import com.intellij.notification.NotificationType;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.openapi.application.ApplicationManager;
@@ -45,6 +49,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
+import javax.swing.event.HyperlinkEvent;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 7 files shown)


alexander-doroshko:
Is there a need for a project service? It looks stateless, so probably replaceable with simple static methods.

devoncarew:
No, I mostly just wanted a singleton that could own a project, but I could re-write to not use a service.

alexander-doroshko:
Thinking about wording. We don't prevent the app from being launched, do we? "Error launching" text seems to bee too severe.

devoncarew:
When it was under development, I did use the text `Launching app` or `Launching <launch-name>`. But it didn't seem very descriptive about why the notification was showing, esp. given that we're always showing errors.

The errors may not be fatal to the app launch however, so `Error launching app` won't always be correct. `Issues with app launch`? `Analysis issues with <launch-name>`?

alexander-doroshko:
I don't know. Maybe smth like 'Errors in code may cause the app crash' :)

devoncarew:
updated to:

screen shot 2017-04-18 at 11 59 17 am

devoncarew:
notes: perhaps a first-class way in the balloon to `don't show this again`

devoncarew:
or `configure`

alexander-doroshko:
LGTM! Will merge tomorrow.

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
```diff
@@ -181,6 +184,34 @@ void computedImplemented(@NotNull final String filePath,
   }
 
   @NotNull
+  List<DartError> getErrors(@NotNull final Module module) {
+    final List<DartError> errors = new ArrayList<>();
+
+    for (String path : myErrorData.keySet()) {
+      final VirtualFile file = LocalFileSystem.getInstance().findFileByPath(path);
+      if (file != null && ModuleUtilCore.moduleContainsFile(module, file, false)) {
+        errors.addAll(myErrorData.get(path));
```
alexander-doroshko: `myErrorData` is accessed in multiple threads. I'm going to surround the whole loop with `synchronized(myErrorData)`.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
```diff
@@ -172,6 +181,44 @@ public static DartProblemsView getInstance(@NotNull final Project project) {
...
+
+  public void clearNotifications() {
+    if (myNotification != null) {
+      myNotification.expire();
+      myNotification = null;
+    }
+  }
+
+  private void showNotification(@NotNull NotificationType notificationType,
+                                @NotNull Project project,
```
alexander-doroshko: `Project` parameter seems to be redundant here and in 2 methods above, we have `myProject` field in hands. I'm on it, no actions needed.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -614,6 +615,14 @@ public void dispose() {
     return myServerData.getErrors(file);
   }
 
+  public List<DartServerData.DartError> getErrors(@NotNull final Module module) {
```
alexander-doroshko: Don't you mind me removing `getErrors(Module)` here and in `DartServerData`? You always can use `module.getModuleContentScope()` instead.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunnerParameters.java
```diff
@@ -151,6 +154,22 @@ public VirtualFile getDartFileOrDirectory() throws RuntimeConfigurationError {
     return dartFile;
   }
 
+  @Nullable
+  public Module getModule(@NotNull Project project) {
```
alexander-doroshko: This method is unused. May I remove it?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
devoncarew: Sounds good! A few of the public APIs are unused by this PR - they are more there for completeness, to make the API a bit more robust possible future client needs.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunnerParameters.java
devoncarew: sounds good; this must have been from before I moved to using the global scope

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunningState.java
```diff
@@ -126,7 +127,27 @@ public void onTextAvailable(final ProcessEvent event, final Key outputType) {
...
+    final Project project = getEnvironment().getProject();
+    try {
+      final DartRunConfiguration dartRunConfiguration = (DartRunConfiguration)getEnvironment().getRunProfile();
+      final VirtualFile launchFile = dartRunConfiguration.getRunnerParameters().getDartFileOrDirectory();
+
+      if (getEnvironment().getRunnerAndConfigurationSettings() == null) {
+        DartExecutionHelper.displayIssues(project, launchFile);
+      }
+      else {
+        String launchTitle = "Analysis issues with " + getEnvironment().getRunnerAndConfigurationSettings().getName();
```
alexander-doroshko: `getEnvironment().getRunnerAndConfigurationSettings().getName()` seems to be equivalent to `dartRunConfiguration.getName()` which has NPE-safe signature. Rewriting it makes `DartExecutionHelper.displayIssues(Project, VirtualFile)` unused. Are you ok with me deleting it?

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunnerParameters.java
devoncarew: sounds good (this might have been from before I moved to using the global scope)

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunningState.java
devoncarew: Yes (sounds like we'll always have a launch title available).

Dart/src/com/jetbrains/lang/dart/ide/runner/DartExecutionHelper.java
```diff
@@ -0,0 +1,115 @@
...
+
+    if (errors.isEmpty()) {
+      return;
+    }
+
+    final DartProblemsView problemsView = DartProblemsView.getInstance(project);
+
+    final String content;
+    if (errors.size() == 1) {
+      content = errors.get(0).getMessage() + " (<a href=\"issues\">show</a>)";
```
alexander-doroshko: How about omitting 'show' link if tool window is not minimized?

Also, I'm going to try to handle href appending and click handling to a single method. Currently, we add a link here, but handle in a different class which is harder to read.

devoncarew: That makes sense re: putting the link creation and handling behavior in one place.

Re: the `show` link, having fewer things in that notifications is better. If you can determine when the tool window is showing, sgtm.

Dart/src/com/jetbrains/lang/dart/ide/runner/DartExecutionHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    // Show a notification on the dart analysis tool window.
+    problemsView.showErrorNotification(project, launchTitle, content, icon);
+
+    // Jump to and highlight the first error (order unspecified) in the editor.
+    DartServerData.DartError error = errors.get(0);
+    final VirtualFile errorFile = LocalFileSystem.getInstance().findFileByPath(error.getAnalysisErrorFileSD());
+    if (errorFile != null) {
+      final OpenFileDescriptor descriptor = new OpenFileDescriptor(project, errorFile, error.getOffset());
+      descriptor.setScrollType(ScrollType.MAKE_VISIBLE);
+      descriptor.navigate(true);
```
alexander-doroshko: Currently, there's no chance to disable jumping to the first error. Even turning the balloon off won't help. For Java, there's an option in Preferences. How about:
- do not jump to error in 2017.1.x;
- think about optional jumping for 2017.2 and give it a longer run during EAP cycle?
I'm afraid some users may find editor switching irritating :).

devoncarew: Hmm, I would like to be able to jump to the error in the Flutter plugin. Some ideas:

- this be off by default in the Dart plugin but available for callers to configure
- we could read the state of the balloon configuration and infer from that what the user would like (if set to not display the balloon, we would not jump to the error position)
- we could add a preference setting now

> give it a longer run during EAP cycle?

Happy to have some sort of EAP / burn in cycle for the feature; would be nice if we could do that in a way that the Flutter plugin could opt-into a faster rollout.

Dart/src/com/jetbrains/lang/dart/ide/runner/DartExecutionHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    if (errors.size() == 1) {
+      content = errors.get(0).getMessage() + " (<a href=\"issues\">show</a>)";
+    }
+    else {
+      content = errors.size() + " " + StringUtil.pluralize("issue", errors.size()) + " found. (<a href=\"issues\">show</a>)";
+      problemsView.showErrorNotification(project, launchTitle, content, icon);
+    }
+
+    // Show a notification on the dart analysis tool window.
+    problemsView.showErrorNotification(project, launchTitle, content, icon);
```
alexander-doroshko: `problemsView.showErrorNotification()` can be called twice, see 4 lines above :)

devoncarew: Hmm, I think you're going to merge a fairly edited version of this PR, but I'll push a fix for this issue for completeness -

Dart/src/com/jetbrains/lang/dart/ide/runner/server/DartCommandLineRunningState.java
alexander-doroshko: ok, will merge this part as is with TODO to make opt-in/out easier before going public

alexander-doroshko:
Pushed with follow-up as discussed above. Before cherry-picking to 2017.1.x or going public as 2017.2 EAP I think we need to sort out the following. Let's continue by email.
- Easier opt-out for the notifications.
- Reconsider default for 'jump to first error', make easy opt-in/opt-out.
- Doesn't work when starting 'all tests in a folder'.
- Usability nit. I see `5 issues found. (show)`, click `show` and see a table with 150 issues. Not clear why there were only 5 issues in the notification and where they are in the opened table.
- Similar usability nit in case of one error. I see `A function body must be provided (show)`, click `show` and, again, I see a table with 150 issues without any idea where `A function body must be provided` error is.

alexander-doroshko:
[landed](https://github.com/JetBrains/intellij-plugins/commits/master/Dart)

PR #520, opened 2017-04-24
Dart support for trailing commas after named args (WIP).

pq:
Adds:

* support for trailing commas after named args in constructors (see: https://github.com/flutter/flutter-intellij/issues/551)

* control via a new Dart code generation preference

screen shot 2017-04-24 at 4 16 17 pm

Besides wordsmithing, the test needs work as this bit in `DartServerCompletionContributor`

    final DartComponent declaration = DartResolveUtil.findConstructorDeclaration(newExpression);

yields `null` (whereas in practice it does not).  Do I need to provoke a resolve?

@alexander-doroshko


alexander-doroshko:
I see that similar options for other languages are in the Smart Keys page (where you've already added `Insert default argument values in completions`). Not an ideal place, but looks more consistent with other technologies and definitely better than 'Code Generation'. This option is about code completion, not about code style/code generation.

Are you saying it doesn't work yet? No guarantees that `findConstructorDeclaration()` has no bugs or that it is a suitable method for this task. So what should be done before thorough review and landing?

P.S. Please rebase on master.

P.P.S. Do you have more background than mentioned in flutter/flutter-intellij#551? It is marked as a blocker there, but without knowing the background it looks just like a minor questionable feature, which has not been asked for years and has not been implemented in other similar languages. It's about typing a comma... how can it be blocking anything ;-)?
BTW does it save keystrokes? Looks like you'll need to press Right Arrow instead of Comma key. Or is the purpose of the feature some other than saving keystrokes?
From the point of view of this feature what is the difference between named args and standard / positional args?

pq:
Thanks for the feedback.

cc @mit-mit for background and justification for the blocker tag.

As for where options should live, I'll migrate.  (Aside: the fact that I forgot where the other options were is either testament to my spaciness or the fact that it's not obvious where these things should go! 😉  )

Why would `findConstructorDeclaration() ` not be suitable?

Any thoughts about getting resolution working in the tests?

alexander-doroshko:
> Why would findConstructorDeclaration() not be suitable?

I'm no answering yet, I'm asking :). I didn't dig into the code yet. So it it suitable? ;)

> Any thoughts about getting resolution working in the tests?

In `DartServerCompletionTest.java` it should work the same as in real run-time, this test uses real analysis server.

pq:
> I'm no answering yet, I'm asking :). I didn't dig into the code yet. So it it suitable? ;)

Well, I *thought* so and assumed the code was vetted since it was in the plugin.

> In DartServerCompletionTest.java it should work the same as in real run-time, this test uses real analysis server.

Right but the issue is not in what I get from the server, it's in the PSI model.  Specifically,

    DartResolveUtil.findConstructorDeclaration(newExpression);

returning `null` in the test but not "in the wild".

Maybe this speaks to the unsuitability of the method?  Since I don't know my way around the PSI element model, I could use some guidance on that...

alexander-doroshko:
You may see PSI using 'View PSI Structure' action. And get general PSI tree idea looking in `dart.bnf` file.

pq:
Thanks!  That plugin was really useful (and the reason I got this far! 👍 )

alexander-doroshko:
I'll be able to help tomorrow if anything remains not clear. Feel free to prepare questions.

alexander-doroshko:
And still curious about the rationale :).

pq:
Updated to a smart key insert preference.

screen shot 2017-04-25 at 3 10 20 pm

pq:
More insight into the failing test.

screen shot 2017-04-25 at 3 30 10 pm

Any idea why the `REFERENCE_EXPRESSION` is resolving to `null` in `com.jetbrains.dart.analysisServer.DartServerCompletionTest#testTrailingCommas`?

Dart/src/com/jetbrains/lang/dart/ide/application/options/DartCodeStyleGenerationConfigurable.form
```diff
@@ -47,6 +44,25 @@
...
               </grid>
+              <grid id="b7063" layout-manager="GridLayoutManager" row-count="1" column-count="1" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+                <margin top="0" left="0" bottom="0" right="0"/>
+                <constraints>
+                  <grid row="1" column="0" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
+                </constraints>
+                <properties/>
+                <border type="etched" title="Auto-insert trailing commas"/>
+                <children>
+                  <component id="da9d2" class="javax.swing.JCheckBox" binding="myInsertTrailingCommasInConsCheckBox">
```
alexander-doroshko: why is this file touched?

Dart/src/com/jetbrains/lang/dart/ide/application/options/DartCodeStyleGenerationConfigurable.java
```diff
@@ -55,6 +57,7 @@ public String getHelpTopic() {
 
   public void reset(CodeStyleSettings settings) {
     myInsertOverrideAnnotationCheckBox.setSelected(settings.INSERT_OVERRIDE_ANNOTATION);
+    myInsertTrailingCommasInConsCheckBox.setSelected(DartCodeInsightSettings.getInstance().TRAILING_COMMAS_AFTER_CONS_ARGS);
```
alexander-doroshko: and this one?

Dart/src/com/jetbrains/lang/dart/ide/formatter/settings/DartCodeStyleSettingsProvider.java
```diff
@@ -9,6 +9,7 @@
 import org.jetbrains.annotations.NotNull;
 
 public class DartCodeStyleSettingsProvider extends CodeStyleSettingsProvider {
+
```
alexander-doroshko: this empty line finally appears to be the only change in this file :)

Dart/src/com/jetbrains/lang/dart/ide/application/options/DartCodeStyleGenerationConfigurable.form
pq: Whoops.  Forgot to revert.  Will fix.

Dart/src/com/jetbrains/lang/dart/ide/application/options/DartCodeStyleGenerationConfigurable.java
pq: And here!

Dart/src/com/jetbrains/lang/dart/ide/formatter/settings/DartCodeStyleSettingsProvider.java
pq: Will fix.

pq:
Hey @alexander-doroshko.  Sorry for the noise.  I only partly backed out my changes when I moved to a smart key preference.  Should be a little tidier now. 👍

pq:
Closing in favor of pushing into server as per conversation in flutter/flutter-intellij#551.

PR #523, opened 2017-05-17
Modify the Dart Analysis Server rename test to make it not blink.

jwren:
@alexander-doroshko


alexander-doroshko:
I'm ok with the change. It was testing a corner case which is probably not so important.
Just curious, why did it make the test flaky?

PR #524, opened 2017-05-22
sort the stagehand templates

devoncarew:
Sort the stagehand templates lexically by name. (the order that stagehand returns them in is undefined; currently it's sorted by template id).

@alexander-doroshko


alexander-doroshko:
Sure, I'm ok with it.
But aren't there plans to introduce any semantic sorting at stagehand end? We'll lose it then.

devoncarew:
No, not really. The stagehand cli ui sorts and displays the templates by template ID. That makes sense for the tool since it only displays ids and descriptions (in IntelliJ, we display labels and descriptions). The `--machine` mode has no real ordering, just the template definition order.

PR #525, opened 2017-05-22
Verify that the Dart VM & Dart Analysis Server files are where the Dart Plugin thinks they should, before invoking.

jwren:


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1405,6 +1406,26 @@ private void startServer(@NotNull final DartSdk sdk) {
...
+      else if (!runtimePathFile.canExecute()) {
+        dasStartupErrorMessage = "the Dart VM file is not executable at location: " + runtimePath;
+      }
+      else if (!dasSnapshotFile.canRead()) {
+        dasStartupErrorMessage = "the Dart Analysis Server snapshot file is not readable at location: " + analysisServerPath;
+      }
+      if (!dasStartupErrorMessage.isEmpty()) {
+        LOG.warn("Failed to start Dart analysis server: " + dasStartupErrorMessage);
+        stopServer();
+      }
```
alexander-doroshko: @jwren shouldn't `return;` be here?

jwren: Yes, fixed and merged into the commit history.

alexander-doroshko:
Thanks, it was already merged (+the Windows-specific fix) at this moment :).

PR #527, opened 2017-06-01
Enable third party IntelliJ plugins to add AnalysisServer listeners to the Analysis Server being used by the Dart Plugin for IntelliJ

jwren:
@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/publicApi/ThirdPartyUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/publicApi/ThirdPartyUtil.java
@@ -1,0 +1,42 @@
+package com.jetbrains.lang.dart.publicApi;
+
+import com.google.dart.server.AnalysisServerListener;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
+import com.jetbrains.lang.dart.ide.actions.DartStyleAction;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+/**
+ * This class is an access point for third party plugins to integrated Dart Plugin features in downstream plugins.
+ */
+@SuppressWarnings("unused")
+public class ThirdPartyUtil {
+
+  /**
+   * Run dartfmt on the specified dart files.
+   */
+  public static void runDartfmt(@NotNull final Project project, @NotNull final List<VirtualFile> dartFiles) {
+    DartStyleAction.runDartfmt(project, dartFiles);
+  }
+
+  /**
+   * For some {@link Project}, add the {@link AnalysisServerListener}. If the {@link Project} has a running service, the listener is added
+   * immediately, otherwise it is added when the server is started.
+   */
+  public static void addAnalysisServerListener(@NotNull final Project project,
+                                               @NotNull final AnalysisServerListener serverListener) {
+    DartAnalysisServerService.addAnalysisServerListener(project, serverListener);
+  }
+
+  /**
+   * For some {@link Project}, remove the {@link AnalysisServerListener}. If the {@link Project} has a running service, the listener is
+   * removed immediately.
+   */
+  public static void removeServerListener(@NotNull final Project project,
+                                          @NotNull final AnalysisServerListener serverListener) {
+    DartAnalysisServerService.removeServerListener(project, serverListener);
+  }
+}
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
No objections. Curious about the use cases.

Dart/src/com/jetbrains/lang/dart/publicApi/ThirdPartyUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import com.jetbrains.lang.dart.ide.actions.DartStyleAction;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+/**
+ * This class is an access point for third party plugins to integrated Dart Plugin features in downstream plugins.
+ */
+@SuppressWarnings("unused")
+public class ThirdPartyUtil {
```
alexander-doroshko: Do you feel strong about the need for such a class?
I consider everything not private within Dart plugin code to be the public API. If we try to gather everything that is already used by Flutter plugin in this class it will already become enormous and harder to use.

jwren:
I care most about just having the functionality.  Having a common API location for documentation feels like a natural addition.  With the "source as documentation" pattern followed by JetBrains, it felt like the right thing to be able to delete the "// leave this here and public for 3rd party plugins"- as the methods are now invoked in a class named "ThirdPartyUtil".

@skybrian @devoncarew 

As for use cases: I'm not completely sure yet I have a few ideas nothing in the short term, but I wanted to get this in for 2017.2.

skybrian:
> "With the "source as documentation" pattern followed by JetBrains

This is a terrible idea and we shouldn't follow it.

alexander-doroshko:
Text like 'Run dartfmt on the specified dart files' doesn't add much value for a method with signature `runDartfmt(dartFiles)`, but in general case, documentation is welcome anywhere, not only in a dedicated class :).

I mean I'm not sure there's benefit in having 2 methods in different classes that are 100% equivalent, like `DartStyleAction.runDartfmt()` and `ThirdPartyUtil.runDartfmt()` or `ThirdPartyUtil.addAnalysisServerListener()` and `DartAnalysisServerService.addAnalysisServerListener()`. There are already >50 of other classes/methods from the Dart plugin used in the Flutter plugin, I don't think we need to mirror them in a class like `ThirdPartyUtil`.

Dart/src/com/jetbrains/lang/dart/publicApi/ThirdPartyUtil.java
skybrian: I agree that a single "Util" class with unrelated static methods bundled into it isn't good in the long term. It would be better to put the methods for each domain in separate classes. Although this API would consist of Java classes and not a protocol, following the examples of the [Flutter Daemon protocol](https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode) or [Analysis server protocol](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html) seems like a good idea.

Also, if someone is going to take this on, it should we be with the long-term goal of not having an enormous API surface between the Dart plugin and other plugins. Instead we'd review what we have and try to come up with a smaller, stable, less brittle public API with no weird hacks so it will work in the long term. (And in particular, untangling monstrosities like DartVmServiceDebugProcessZ.)

But I don't feel strongly about starting this project now. I won't have time to work on it any time soon, and if nobody is volunteering to work on it, perhaps it is better to wait?

skybrian:
Okay, yes, I agree that comments that don't add value shouldn't be written. A comment that says "this is used in another plugin" seems like a good warning since they are different code bases and show usages won't work, but being systematic about it would be better.

In the Flutter plugin we are just using anything public if it's needed functionality, including very dubious hacks, but this makes things brittle so it would be better to clean that up. (But this requires a new release of the Dart plugin with the new API so it is a slower process.)

alexander-doroshko:
So I suggest to keep as is :). 
`addAnalysisServerListener()` method looks good in `DartAnalysisServerService` class and `runDartfmt()` looks good in `DartStyleAction` (and would be even better if we rename it to `DartfmtAction` but probably it's too late?)

As for `DartVmServiceDebugProcessZ` - I'm totally for its removal. I suggest to do it now, before 2017.2 release. Let me know if I can help.

alexander-doroshko:
And re: the Dart plugin API that could be improved - let's do it now, before 2017.2.

skybrian:
Just to clarify what I was thinking, if we were doing the domain thing we'd still have the publicApi package, but there would be one class in that package for each domain that's considered part of the public API.

If we go with existing classes then I think that's putting off extracting a public API for now.

alexander-doroshko:
Merged manually!
I made methods not static to be on a par with others, standard usage is `DartService.getInstance(project).doSmth()`.

jwren:
Thank you @alexander-doroshko

PR #529, opened 2017-06-07
tweak the text for whole file formatting

devoncarew:
@alexander-doroshko, minor tweak to the text for whole-file formatting.


alexander-doroshko:
This text was suggested by WebStorm tech writer. She also suggests to optionally add 'the' before 'dartfmt'. Why do you think it's better without 'tool' word? I'll ask tech writer again with your arguments :).

devoncarew:
I think you either want 'use the dartfmt tool when' or 'use dartfmt when' . Either would work - just 'use dartfmt tool when' seems like it's missing a word when you read it.

alexander-doroshko:
Thanks! We'll go with 'the' then.

PR #530, opened 2017-06-14
work to support configuring exception breakpoints

devoncarew:
Work to support configuring exception breakpoints:

- add a new breakpoint type, `DartExceptionBreakpointType`; this shows up in the breakpoints dialog
- have the UI for it allow two settings, `break on uncaught` and `break on all` exceptions
- plumb the UI info through to vm service calls, including listening for changes to the setting while the app is running

@alexander-doroshko 

I do see issues where changing the value from `break on uncaught` to `break on all` and back doesn't kick out an event. Changing the breakpoint state from enabled to disabled does fire events.

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -203,6 +203,7 @@
     <console.folding implementation="com.jetbrains.lang.dart.ide.runner.server.DartConsoleFolding"/>
 
     <xdebugger.breakpointType implementation="com.jetbrains.lang.dart.ide.runner.DartLineBreakpointType"/>
+    <xdebugger.breakpointType implementation="com.jetbrains.lang.dart.ide.runner.DartExceptionBreakpointType"/>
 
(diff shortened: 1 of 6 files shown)


alexander-doroshko:
@devoncarew Sorry, can it wait a week? I want to spend more time on the review than I have today.

devoncarew:
@alexander-doroshko, no problem, I was out as well. This will be great to have when it lands - it's something we've wanted for a while - but not having it isn't blocking anything.

alexander-doroshko:
@devoncarew Cool, pushed!
I've rewritten it a bit to be less hacky and to support changing the option on the fly, as well as to respect 'Suspend' check box, show icon when suspended and other minor tweaks. Please review/test: https://github.com/JetBrains/intellij-plugins/commit/5aab99d30ba2d2efa3ff8032bbc4933068799e4b

devoncarew:
I took a quick pass through the changes; thanks for the updates!

PR #535, opened 2017-06-22
use the BrowserUtil class

devoncarew:
In the AnalysisServerDiagnosticsAction class, switch to using the BrowserUtil class for opening urls. Locally when using webstorm, I see this exception:

```
Error opening Dart Analysis Server diagnostics page: com.intellij.ide.browsers.BrowserLauncherImpl$checkCreatedProcess$1 cannot be cast to java.lang.Runnable
```

and from googling, a suggestion was to switch to using the `BrowserUtil` class.

@alexander-doroshko


alexander-doroshko:
Ok, pushed.
Curious how to reproduce the issue. Maybe I need to fix `OpenDartObservatoryUrlAction.openInAnyChromeFamilyBrowser`.

PR #537, opened 2017-06-26
Postfix completion

stevemessick:
@alexander-doroshko Here's the postfix completion driver for Dart. It needs a version of the server with this change patched in to function:
https://codereview.chromium.org/2917943002/

There are a bunch of empty non-public classes defined to provide documentation. Ideally, all the docs would be obtained from the server, but that requires a custom build step. Doable, but not high priority IMO. The preferences/settings page for postfix completion lists all the templates defined for Dart and includes before and after samples. If that page is opened before any project has been created you'll only see one template, but the list will be updated when an analysis server is running.

I have not learned how to enable expansion of a postfix-! so for now .not is the only way to invoke that template.

I re-created the PR to include some cleaned-up files that I missed after making a change to the server protocol, along with some docs.


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1713,6 +1796,24 @@ private static boolean awaitForLatchCheckingCanceled(@NotNull final AnalysisServ
     }
   }
 
+  private void registerPostfixCompletionTemplates() {
+    ApplicationManager.getApplication().invokeLater(() -> {
+      // There must be a better way to initialize postfix templates!
+      for (int n = 20; n > 0; n--) {
+        if (DartPostfixTemplateProvider.initializeTemplates(this)) {
```
alexander-doroshko: @stevemessick, is this hack to make sure that the server is warmed up? Maybe catch the moment when initial 'pub list' or initial analysis is done (see around `serverStatus()` method)?

Also, as `DartPostfixTemplateProvider.initializeTemplates()` is called on server startup is there a need to call it again in `DartRemotePostfixTemplate.isApplicable()`?

stevemessick: The initialization only needs to be done in one place, if it is reliable. It wasn't before, but I think it is now.
```
1498684845069:Req:{"id"::"6","method"::"edit.listPostfixCompletionTemplates","clientRequestTime"::1498684843493}
1498684845073:Res:{"id"::"6","result"::{"templates"::[{"name"::"assert","key"::".assert","example"::"expr.assert -> assert(expr);"},{"name"::"fori","key"::".fori","example"::"limit.fori -> for(var i = 0; i < limit; i++) {}"},{"name"::"for","key"::".for","example"::"values.for -> for(var value in values) {}"},{"name"::"iter","key"::".iter","example"::"values.iter -> for(var value in values) {}"},{"name"::"not","key"::".not","example"::"bool.not -> !bool"},{"name"::"!","key"::"!","example"::"bool! -> !bool"},{"name"::"else","key"::".else","example"::"bool.else -> if (!bool) {}"},{"name"::"if","key"::".if","example"::"bool.if -> if (bool) {}"},{"name"::"nn","key"::".nn","example"::"expr.nn -> if (expr != null) {}"},{"name"::"notnull","key"::".notnull","example"::"expr.notnull -> if (expr != null) {}"},{"name"::"null","key"::".null","example"::"expr.null -> if (expr == null) {}"},{"name"::"par","key"::".par","example"::"expr.par -> (expr)"},{"name"::"return","key"::".return","example"::"expr.return -> return expr"},{"name"::"switch","key"::".switch","example"::"expr.switch -> switch (expr) {}"},{"name"::"try","key"::".try","example"::"stmt.try -> try {stmt} catch (e,s) {}"},{"name"::"tryon","key"::".tryon","example"::"stmt.try -> try {stmt} on Exception catch (e,s) {}"},{"name"::"while","key"::".while","example"::"expr.while -> while (expr) {}"}]}}
```
It takes a while for the server to receive the initialization request. Allowing more time for the request to be processed fixes it.

stevemessick:
This PR is obsolete and should be close without merging.

PR #538, opened 2017-06-28
Postfix completion, new branch

stevemessick:
@alexander-doroshko I got some git commands out-of-order and ruined the original branch. The old PR is still there in case you need to refer to it for something. This is a replacement for it that includes the modification to template initialization. As I mentioned in a reply comment in the old PR, the initialization time-out was too short so it was failing. I increased the wait time and everything works with a single initialization attempt now. I modified two files: DartAnalysisServerService.java and DartPostfixTemplateProvider.java.


Dart/resources/postfixTemplates/DartRemotePostfixTemplate/description.html
```diff
@@ -0,0 +1,5 @@
+<html>
```
alexander-doroshko: I'm getting an error because `before.dart.template` and `after.dart.template` are missing next to this file. I've added empty ones, let me know if you meant smth else.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -960,6 +968,80 @@ public void onError(final RequestError error) {
...
 
+  public boolean edit_isPostfixCompletionApplicable(VirtualFile file, int offset, String key) {
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return false;
+    }
+    final String filePath = FileUtil.toSystemDependentName(file.getPath());
+    final Ref<Boolean> resultRef = Ref.create();
+    final CountDownLatch latch = new CountDownLatch(1);
+    server.edit_isPostfixCompletionApplicable(filePath, key, offset, new IsPostfixCompletionApplicableConsumer() {
```
alexander-doroshko: I've added `final int offset = getOriginalOffset(file, _offset);`

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -960,6 +968,80 @@ public void onError(final RequestError error) {
...
+      @Override
+      public void isPostfixCompletionApplicable(Boolean value) {
+        resultRef.set(value);
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, POSTFIX_COMPLETION_TIMEOUT);
+    return resultRef.get();
```
alexander-doroshko: This may cause NPE because of unboxing. Fixed by initial `false` value for `Ref.create()`

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -960,6 +968,80 @@ public void onError(final RequestError error) {
...
+  }
+
+  public PostfixCompletionTemplate[] edit_listPostfixCompletionTemplates() {
```
alexander-doroshko: Added `@Nullable`

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -960,6 +968,80 @@ public void onError(final RequestError error) {
...
+        logError("edit_listPostfixCompletionTemplates()", null, error);
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, POSTFIX_INITIALIZATION_TIMEOUT);
+    return resultRef.get();
+  }
+
+  public SourceChange edit_getPostfixCompletion(@NotNull final VirtualFile file, final int _offset, final String key) {
```
alexander-doroshko: Added `@Nullable`

Dart/src/com/jetbrains/lang/dart/ide/template/postfix/DartRemotePostfixTemplate.java
```diff
@@ -0,0 +1,195 @@
...
+
+  @Override
+  public boolean isApplicable(@NotNull PsiElement context, @NotNull Document copyDocument, int newOffset) {
+    final Project project = context.getProject();
+    final PsiFile psiFile = context.getContainingFile();
+    final DartAnalysisServerService service = DartAnalysisServerService.getInstance(project);
+    String version = service.getSdkVersion();
+    Set<PostfixTemplate> templates = DartPostfixTemplateProvider.getTemplates(version);
+    boolean found = false;
+    for (PostfixTemplate temp : templates) {
```
alexander-doroshko: I've got an NPE here when switched to a different project with older Dart SDK. Marked `DartPostfixTemplateProvider.getTemplates(version)` as `@Nullable` and added null check here.

alexander-doroshko:
Looks great, pushed!

Minor tweaks applied.
And a couple of things not applied yet:
- Would be great to add a test if possible.
- It is theoretically possible that `edit_listPostfixCompletionTemplates()` blocks UI fo 5 seconds (because `invokeLater()` is used, so it runs in EDT). But looks like there's no need in running it in EDT. We could probably run it in pooled thread and add the corresponding sync to the  var `DartPostfixTemplateProvider.TEMPLATE_CACHE` that becomes multi-thread.

PR #539, opened 2017-06-29
Add dependency on xdebugger-impl

stevemessick:
@alexander-doroshko Not sure why this isn't already there.


alexander-doroshko:
Thanks! My bad, I forgot that I had to keep 2 iml files in sync. Pushed.

PR #540, opened 2017-06-30
prefer asyncCausalFrames, then regular sync frames

devoncarew:
- prefer asyncCausalFrames, then regular sync frames
- fix https://youtrack.jetbrains.com/issue/WEB-27420

@alexander-doroshko


devoncarew:
(Alex, I'm not at my work computer - I have not yet had a chance to test this)

PR #541, opened 2017-07-05
some cleanup to the run notification

devoncarew:
- fix https://youtrack.jetbrains.com/issue/WEB-26864
- remove the ability to jump to an error on run
- add a `don't show again` link
- make the `open view` link not show if the Dart Analysis view is already open
- do not show the number of issues in the notification (this could be inconsistent w/ the filter in the analysis view)

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
@@ -19,10 +19,9 @@ import com.intellij.execution.runners.ExecutionUtil;
 import com.intellij.ide.projectView.ProjectView;
 import com.intellij.ide.projectView.impl.ProjectViewPane;
 import com.intellij.ide.util.PropertiesComponent;
-import com.intellij.notification.Notification;
-import com.intellij.notification.NotificationGroup;
-import com.intellij.notification.NotificationListener;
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
@devoncarew I keep thinking about the balloon text and links. Currently, we have
image
How would you rate a bit updated one?
image
And if you click `Never show again` you see:
image

devoncarew:
Those changes look great!

I might tweak the first one from `Dart package 'foobar' contains analysis issues that may affect the execution of 'example.dart'.` to `'foobar' contains analysis issues that may affect the execution of 'example.dart'.` (i.e., drop the Dart package prefix).

alexander-doroshko:
Thanks! 
Btw is it better with apostrophes or without? Or mixed (as Dart package name is a single word, but RC name may contain spaces)?
With period at the end or without (especially given that the first symbol will likely be lowercase - it is a dart package name)?
Should links start with a capital letter or not?

alexander-doroshko:
Or maybe even shorter:
```
Analysis issues may affect the execution of 'My awesome RC name'.
```

devoncarew:
Perhaps `Module foo contains analysis issues that may affect the execution of example.dart.`?

devoncarew:
I terms of the period, I would just see if there's a typical pattern in IntelliJ for other notifications of this type :)

PR #542, opened 2017-07-17
Add analytics checkbox

devoncarew:
Add an analytics checkbox to control sdk analytics:
- update the analysis server protocol lib to include the new `analytics.enable(bool)`, `analytics.isEnabled` methods
- in `DartAnalysisServerService.java`, add code to get and set the setting, and to cache it if the server is not available, so we can still show the checkbox in the UI
- add a checkbox to the preference page

screen shot 2017-07-14 at 5 49 20 pm

@alexander-doroshko


devoncarew:
As per an off-line discussion, closing this for now while we iterate and seek another solution.

PR #544, opened 2017-07-20
Update FlutterUtil in Dart plugin to continue to prevent the Dart plugin from contributing to modules where the Flutter plugin also contributes

jwren:
This fixes https://github.com/flutter/flutter-intellij/issues/1225


jwren:
@alexander-doroshko

alexander-doroshko:
Pushed!
Will it make sense to also add `isFlutterPluginInstalled()` check to the `isFlutterModule()` implementation?

alexander-doroshko:
You are right, `isFlutterPluginInstalled()` shouldn't be a part of `isFlutterModule()` implementation, additional check is done by clients when needed.

PR #545, opened 2017-07-21
Add a project level preference in the DartAnalysisServerService to enable and disable the DAS general subscription: "analyzed files"

jwren:
@alexander-doroshko


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -315,6 +319,17 @@ public void serverStatus(@Nullable final AnalysisStatus analysisStatus, @Nullabl
     }
   };
 
+  public static boolean doEnableAnalyzedFilesSubscription() {
+    PropertiesComponent properties = PropertiesComponent.getInstance();
```
alexander-doroshko: `PropertiesComponent.getInstance()` returns app-level component. If you want this setting to be per-project (I think so reading the title) use `PropertiesComponent.getInstance(Project)` (you'll have to make it not static then).

Cosmetics: how about renaming this method to `isAnalyzedFilesSubscriptionEnabled()`? BTW do you need it to be public?

jwren: Yes, I changed it yesterday from being a project level setting to a global setting, I squashed the commits together so you can't see the history.

I figured that if someone is enabling/disabling it they probably want it configured everywhere, also by not having it be a project setting, the UI will be easier and cleaner to implement.  I changed the source yesterday morning but the GitHub web UI was down so I couldn't change the text in the PR, updating it now.  Updating the method name now.  I think it does need to be public to be called by third party plugins.

jwren:
PR updated.

alexander-doroshko:
Pushed with a minor [tweak](https://github.com/JetBrains/intellij-plugins/commit/e5a34beaa692fa646b2470be2dfe2d6826769d02) not to store the default value.

Just realized that the default behavior for Google devs is dramatically changed with this PR. Are you sure this is what you want? With this change, you get rid of 'Dart Packages' lib for internal projects, all the dependencies in internal projects are treated as non-project files. You loose Go To File / Go To Class / Go To Symbol functionality for dependencies, [highlight usages at caret](https://youtrack.jetbrains.com/issue/WEB-27223) feature and much more. Curious what has changed that you decided to switch off the long-living feature by default?

alexander-doroshko:
So, with this change I think the client will have to:
- If `"dart.projects.without.pubspec"` is `true` and `isAnalyzedFilesSubscriptionEnabled()` is `false` (which is now by default) then remove previously generated 'Dart Package' library (by calling `DartFileListener.updateDependenciesOnDartPackagesLibrary(project, Condition.FALSE, library)`) on project opening.
- and also remove the lib when calling `setEnableAnalyzedFilesSubscription(false)` (again, only if `"dart.projects.without.pubspec"` is `true`)
- when calling `setEnableAnalyzedFilesSubscription(true)` you may want to restart Analysis Server or suggest user to (again, only if `"dart.projects.without.pubspec"` is `true`).

jwren:
Thanks Alex for putting these notes into this PR.

PR #549, opened 2017-09-20
initial plumbing for closing labels support

devoncarew:
@alexander-doroshko, this is some initial plumbing for 'closing labels' support. These are labels displayed inline into the editor which, for code like Flutter's build() method, can help users match the final line of a nested constructor call with the starting line. More info in https://github.com/flutter/flutter-intellij/issues/1252 and related.

We have a bit more work on the analysis server side in terms of:
- making the notifications less noisy (code changes can produce more of these notification than is absolutely necessary)
- and producing fewer notifications - we create more in a build() method than we probably should

But for the IntelliJ side, I could use some advice for this PR. It looks like in order to produce an Inlay, you start with an Editor instance, call `InlayModel getInlayModel()`, and use that returned object to create `Inlay`s. Where would be a good place in the Dart plugin to look for where I should be getting the Editor instances?

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -224,6 +224,11 @@ public class DartAnalysisServerService implements Disposable {
     }
 
     @Override
+    public void computedClosingLabels(@NotNull final String filePath, List<ClosingLabel> labels) {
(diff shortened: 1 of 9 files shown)


alexander-doroshko:
I think when you get `computedClosingLabels()` notification you may schedule a EDT activity that would update inlays for the corresponding editor(s) (`FileEditorManager.getEditors(file)`).

alexander-doroshko:
Merged

PR #551, opened 2017-10-05
some tweaks to the dart method refactoring dialog

devoncarew:
This PR makes a small tweak to the refactor method dialog so that the `getter` option is not selected by default when entering the dialog. It's a bit confusing to users (who don't necessarily read all the options) that sometimes the dialog creates getters and sometimes not. Without the user explicitly specifying it, we now default to creating methods (not getters).

@alexander-doroshko


alexander-doroshko:
LGTM.
// Cc @scheglov as the original author of these lines.

scheglov:
LGTM

PR #552, opened 2017-10-06
Use already known step annotation when extracting regex pattern 

jhorstmann:
...instead of relying on cucumber package names

**As** the creator of an alternative cucumber implementation
**I want** my implementation to be supported by intellij
**So that** my users are happy

The cucumber-java plugin currently uses two different ways to identify annotated step definitions:

 - In `CucumberJavaExtension` it correctly uses the `StepDefAnnotation` meta-annotation to find all annotated methods
 - For matching steps against patterns, `CucumberJavaUtil` relies on the package name of the methods annotation

I would like the plugin to support additional annotations, which are still annotated by `StepDefAnnotation`, but reside in a different package than the cucumber ones. The PR solves this by reusing the discovered annotation class from the step definition.

Please note that this is the first time I'm touching an IntelliJ plugin and I'm still struggling with running the unit tests.

There are also still several usages of the package-based logic, which I was not able to replace yet.

### DIFF
--- a/cucumber-java/src/org/jetbrains/plugins/cucumber/java/CucumberJavaExtension.java
+++ b/cucumber-java/src/org/jetbrains/plugins/cucumber/java/CucumberJavaExtension.java
@@ -52,7 +52,7 @@ public class CucumberJavaExtension extends AbstractCucumberJavaExtension {
       if (annotationClass.isAnnotationType()) {
         final Query<PsiMethod> javaStepDefinitions = AnnotatedElementsSearch.searchPsiMethods(annotationClass, dependenciesScope);
         for (PsiMethod stepDefMethod : javaStepDefinitions) {
-          result.add(new JavaStepDefinition(stepDefMethod));
+          result.add(new JavaStepDefinition(stepDefMethod, annotationClass));
         }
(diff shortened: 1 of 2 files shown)


avokin:
The request a little bit outdated but I would like to merge it. Could you provide us signed agreement?
See instruction [here](https://www.jetbrains.com/agreements/cla/)

jhorstmann:
I signed the Contributor Agreement now and am looking into the merge conflicts. It seems the changes from IDEA-183813 touched the same parts.

I'll have to familiarize myself a bit with the code again, but it might be the only needed change is to remove the check for `isCucumberStepAnnotation` in `CucumberJavaUtil.getCucumberStepAnnotation`. The `annotationClassName` parameter in `JavaStepDefinition` is already determined to be a step definition annotation.

avokin:
Jörn, I've tried to resolved conflicts and after it was done, I found out that there is nothing to cherry pick - I've recently done almost the same. So I propose to close this PR.

By the way, recently I've added support for Zuchini framework: 
83f9f4b8551f05ece88c6d53699ff4637e062faf

PR #553, opened 2017-10-10
update the icons we use for test outline elements

devoncarew:
@alexander-doroshko, this updates how we determine want icons to use for `group` and `test` items in the outline view (it uses newly undeprecated element kinds), and switches the icons that we use. Here's the latest:

screen shot 2017-10-10 at 2 25 12 pm

This depends on some soon-to-land changes in the analysis server (https://dart-review.googlesource.com/c/sdk/+/12841).

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/structure/DartStructureViewElement.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/structure/DartStructureViewElement.java
@@ -1,5 +1,6 @@
 package com.jetbrains.lang.dart.ide.structure;
 
+import com.intellij.icons.AllIcons;
 import com.intellij.ide.structureView.StructureViewTreeElement;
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
Thanks, much better this way!
Are you aware of any other Outline nodes that Analysis Server reports as `ElementKind.UNKNOWN`?

devoncarew:
That was the only one to date. We were fortunate in that there were exact nodes to use for package:test (they just needed to be undeprecated).

We'll likely send some future nodes as unknown however - specifically nodes for widgets in a flutter build() method. There was some discussion here yesterday that Elements don't map well to how we want to use the structure view for some use cases, like showing ctor invocations for build() methods, and likely in the future, providing the right info to allow light mutation of the tree.

Hopefully something we can talk about in the meeting today :) /cc @scheglov @bwilkerson

PR #554, opened 2017-10-12
add new, final, const keywords

devoncarew:
- add `new`, `final`, `const` keywords to the Dart colors and fonts pref page

@alexander-doroshko, this seems to highlight fine w/o the `<angle></angle>` wrappers. Do I need them?

Also, any objections if I follow this PR up with one that reformats the dart code snippet a bit, so it looks like something that may have gone through dartfmt?


alexander-doroshko:
The purpose of this preview is to have at least one example of each color setting, to be compact and preferrably to be valid Dart code. 
What problem this PR solves? There's already plenty of keyword examples.
Reformatting will make it 4 times longer, it won't fit any screen.

devoncarew:
I didn't see examples of `new` and `const`; likely not necessary to have all keywords, but those stuck out as missing.

Compactness makes sense. From perusing some of the other language pages (java, groovy, kotlin) it seems like they use a bit more whitespace than the Dart page, and it is a bit disconcerting to me in places in the Dart page where it's clearly not following dartfmt conventions (I guess it doesn't need to be 100% accurate - perhaps just avoid clear differences?).

alexander-doroshko:
I'm sure much more keywords are missing. I mean one keyword at all (any) would be enough for the purpose of being able to preview and configure the color. 

If it's possible to make it look better without making much longer then ok. Otherwise, for the purpose of colors configuration, this page looks pretty good to me.

devoncarew:
OK, I'll see if there's something which really stands out -

PR #555, opened 2017-10-18
add icons for ctor and function invocations

devoncarew:
- prep for two new element kinds showing in the outline - constructor and function invocations. These aren't sent by the analysis server yet, but we anticipate using them for the outline for flutter build methods.

Currently using icons that are slight variants on the class and method icons, badged so that they look slightly different; open to other ideas :)

screen shot 2017-10-18 at 12 59 19 pm

@alexander-doroshko


alexander-doroshko:
I forgot to say, [pushed](https://github.com/JetBrains/intellij-plugins/commit/b4c28be2d2ee37b37f35dd77b164798680b2367d#diff-35900d987289a83af101d18a9f7c3501) yesterday.

PR #557, opened 2017-11-03
Fix for Dart/Flutter plugin issue https://github.com/flutter/flutter-intellij/issues/1449

jwren:
Fix for https://github.com/flutter/flutter-intellij/issues/1449, and a change to the Dart Plugin and Flutter Plugin to check for the flutteriness of a pubspec.  We now look to see if the individual pubspec declares flutter when determining if a notification should appear for a pubspec, _instead_ of asking the module if some pubspec.yaml declares flutter.

Nit: when comparing a file name to "pubspec.yaml" the two plugins are not consistent in _not_ ignoring the case of the filename.

Corresponding PR: https://github.com/flutter/flutter-intellij/pull/1468

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
@@ -4,10 +4,20 @@ import com.intellij.ide.plugins.PluginManager;
 import com.intellij.openapi.extensions.PluginId;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleType;
+import com.intellij.openapi.vfs.VirtualFile;
 import com.jetbrains.lang.dart.sdk.DartSdk;
 import com.jetbrains.lang.dart.sdk.DartSdkLibUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.yaml.snakeyaml.DumperOptions;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+import org.yaml.snakeyaml.nodes.Tag;
+import org.yaml.snakeyaml.representer.Representer;
+import org.yaml.snakeyaml.resolver.Resolver;
+
+import java.io.IOException;
+import java.util.Map;
 
 public class FlutterUtil {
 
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -38,4 +48,60 @@ public static boolean isFlutterModule(@NotNull final Module module) {
...
   }
+
+  /**
+   * Returns true if passed pubspec declares a flutter dependency.
+   * <p/>
+   * This method intentionally matches the contents in PubRoot.declaresFlutter in the Flutter plugin, which ensures that
+   * the same logic is used between the two plugins to avoid issues like this:
+   * https://github.com/flutter/flutter-intellij/issues/1445
+   */
+  public static boolean doesPubspecDeclaresFlutter(@NotNull final VirtualFile pubspec) {
```
pq: => doesPubspecDeclareFlutter or maybe `isPubspecDeclaringFlutter`?

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
```diff
@@ -38,4 +48,60 @@ public static boolean isFlutterModule(@NotNull final Module module) {
   public static boolean isFlutterPluginInstalled() {
     return FLUTTER_PLUGIN_INSTALLED;
   }
+
+  /**
+   * Returns true if passed pubspec declares a flutter dependency.
```
pq: =>

Returns true if passed a pubspec that declares a flutter dependency.

?

alexander-doroshko:
LGTM!
Let me know whether I should push it or wait for the update according to Phil's comments.
Also, do we need to update other 3 usages of `isFlutterModule()`?

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
jwren: modified

Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
jwren: Changed

jwren:
@alexander-doroshko It depends on the context of the check.  In this case we wanted to know if some pubspec.yaml was flutter, but we were calling isFlutterModule(Module), which considers all pubspecs in the module.

In cases where we have some specific VirtualFile (instead of Module) it would make sense to follow this PRs lead.

PR #558, opened 2017-11-19
address an issue with the height of closing labels

devoncarew:
- ask the editor for the font ascent to use (to account for the user possibly changing the line height) - fix https://github.com/flutter/flutter-intellij/issues/1514

@alexander-doroshko


devoncarew:
Note that I don't see the `XDebuggerInlayUtil.MyRenderer` doing this, so I don't know how they're avoiding this same issue.

alexander-doroshko:
Pushed to 2017.3.1.
It appears that it is not `XDebuggerInlayUtil` that renders inline debugger data. `XDebuggerInlayUtil` is for some experimental feature that is disabled, so effectively it is not used at all.

brianegan:
Hello @alexander-doroshko :) I've just downloaded `2017.3.1 EAP`, but the labels are still appearing misaligned. Do you know if this fix has shipped with that version?

alexander-doroshko:
Hi, 
updated Dart plugin with the fix will be published later this week. It will be targeting 2017.3.1 EAP.

PR #559, opened 2017-12-03
avoid an exception from a @NotNull annotation

devoncarew:
- avoid an exception from a @NotNull annotation

While debugging, I saw this exception:

```
Argument for @NotNull parameter 'listElements' of com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.addListChildren must not be null
```

Instance.getElements() shouldn't return null, but it can in certain error circumstances. This moves the validation out of annotations.

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
@@ -377,11 +377,13 @@ public class DartVmServiceValue extends XNamedValue {
     });
   }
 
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
@devoncarew I'm ok with the change and have pushed it manually.

As far as I understand VM Service protocol doesn't allow sending `null` as `Instance.getElements()` for lists. Just wanted to note that with this change we lose the ability to catch protocol violation. If anybody cares about the root problem they might want to find other ways of catching it.

devoncarew:
Yup, I agree this is not fixing the issue at the source. I saw this when trying to expand a large array; I believe on the order of a 50k long byte array (representing text).

PR #560, opened 2017-12-27
error-prone plugin: update jar version to latest

ttsugriy:
Version 2.0.13 is more than 2 years old and misses a lot of useful checks.


chashnikov:
I've updated Error Prone to the latest version (2.2.0). This also required changing JVM arguments for the build process, see [my commit](https://github.com/JetBrains/intellij-plugins/commit/d7de6c7fcbf05bc8be416a2626e8adc0afaf07dc) for details.

ttsugriy:
Awesome, thanks Nikolay!

PR #561, opened 2018-01-17
use session.areBreakpointsMuted to help determine the break on exceptions mode

devoncarew:
- use session.areBreakpointsMuted to help determine the break on exceptions mode

This does address https://github.com/flutter/flutter-intellij/issues/1164, but will only help users of the Flutter plugin who have a version of the Dart plugin with this fix.

@alexander-doroshko


alexander-doroshko:
That's the best solution, thanks!

PR #562, opened 2018-01-24
Add APIs to DartAnalysisServerService to send arbitrary requests and process all responses.

scheglov:


scheglov:
@alexander-doroshko

PR #563, opened 2018-02-03
Icons and Colors for Dart/Flutter Completion Suggestions.

pq:
Adds colors and material icons to completions.

image

image

See: https://github.com/flutter/flutter-intellij/issues/454

Essentially this ports related support for editor annotations from the Flutter plugin; in the medium term we might update that to call back to the Dart Plugin and avoid duplication.

@devoncarew @alexander-doroshko


pq:
FWIW: the only really interesting bits are in:

* Dart/src/com/jetbrains/lang/dart/flutter/FlutterUtil.java
* Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java

devoncarew:
This is awesome! This has been asked for a bunch, and will help discovery of the material icons.

Not to derail this PR, but given the number of files here, and the fact that we do occasionally want to maintain them, I wonder if we shouldn't look into an extension point style mechanism first? The Dart plugin would provide a way to decorate code completion with icons, and the Flutter plugin could provide a specific implementation of that extension. I see that there is a `com.intellij.ide.IconProvider` interface, but not sure if it matches up with what we need (we'd need a callback interface that the Flutter plugin could implement, and compile against).

pq:
>  I wonder if we shouldn't look into an extension point style mechanism first?

I'm totally open to this.  Alternatively, I was wondering about moving icons into the flutter SDK (this would benefit other tool integrators) but maybe that's a hard sell?

alexander-doroshko:
An extension point sounds good to me!
I guess you'll benefit from being more flexible with the plugin updates if you keep maintaining the icons on the Flutter plugin end.

pq:
Extension point conversation kicked off here: https://youtrack.jetbrains.com/issue/WEB-31129

pq:
Closing in favor of:  https://github.com/flutter/flutter-intellij/pull/1968

Thanks again for the extension point @alexander-doroshko 👍

PR #566, opened 2018-02-14
Don't select the current line in 'Extract Method' when selection is empty.

scheglov:
Instead Dart Analysis Server [was enhanced](https://dart-review.googlesource.com/c/sdk/+/41520) to extract the smallest extractable expression that covers the caret.


alexander-doroshko:
@scheglov Thanks, merged.
Can we get info from the server about the code range that it is going to extract and set selection in the editor accordingly at the moment when modal dialog requesting for the method name is shown?

alexander-doroshko:
@scheglov Can you please review/test this [follow-up commit](https://github.com/JetBrains/intellij-plugins/commit/4cb339b1707175f8b1137b71797f242a85c0f1e8)?

PR #567, opened 2018-02-20
modify the analysis server status and feedback ui

devoncarew:
- change the feedback action icon so that isn't more clearly about providing feedback
- when the analysis server dies, show this via a tool window notification, rather than updating the feedback icon

screen shot 2018-02-19 at 6 30 17 pm

screen shot 2018-02-19 at 6 31 19 pm

@alexander-doroshko


alexander-doroshko:
Cc @stevemessick for his lgtm. Or have you already discussed it together?

BTW I don't remember why `DartFeedbackBuilder` was implemented as an extension point. Are you using it in an internal plugin? If so, don't you break compatibility by removing `DartFeedbackBuilder.showQuery()` method?

devoncarew:
> Or have you already discussed it together?

We haven't, but I should have cc'd him myself.

For context, this work motivated by reports that it's still not clear enough to the user when the analysis server has crashed.

> BTW I don't remember why DartFeedbackBuilder was implemented as an extension point. Are you using it in an internal plugin? If so, don't you break compatibility by removing DartFeedbackBuilder.showQuery() method?

We do have an internal implementation; I'll add the `showQuery` method back in.

alexander-doroshko:
>  I'll add the showQuery method back in.

Well, if your internal implementation doesn't use it then it's safe to delete.

devoncarew:
Looks like it overrides `prompt()` and `sendFeedback()`.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
```diff
@@ -203,15 +207,18 @@ public void clearNotifications() {
...
 
     if (myDisabledForSession) return;
 
     content = StringUtil.notNullize(content);
-    if (!content.endsWith("<br>")) content += "<br>";
-    content += "<br><a href='disable.for.session'>Don't show for this session</a>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;" +
-               "<a href='never.show.again'>Never show again</a>";
+    if (!terse) {
+      if (!content.endsWith("<br>")) content += "<br>";
+      content += "<a href='disable.for.session'>Don't show for this session</a>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;" +
```
alexander-doroshko: An empty line is added here on purpose, so I recovered it.

alexander-doroshko:
Thanks, pushed.

BTW don't you find that the new icon looks like a disabled action?
Maybe consider something like `AllIcons.General#Balloon`?

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblemsView.java
devoncarew: I removed for a slightly more compact display, but your call :)

devoncarew:
> BTW don't you find that the new icon looks like a disabled action?

The tool window's top toolbar does seem to use gray for their icons. I looked - in darcula they use `#8C8C8C` - which is what this icon uses. In the regular color scheme they use `#808080`, so this icon will be slightly off there. We could add a second set of icons for the regular theme.

You could try with `AllIcons.General#Balloon` but I think the color there would stand out against the other icons.

alexander-doroshko:
Ok, I'll keep using the icon from this PR.

PR #570, opened 2018-03-16
update the dart vm service library

devoncarew:
- update the dart vm service library

Nothing particularly critical to have here, but re-generating the library from the latest spec.

@alexander-doroshko


alexander-doroshko:
Thanks, merged!

PR #573, opened 2018-04-09
update the service protocol lib

devoncarew:
- update the service protocol lib to capture an update signature for `getSourceReport()` (currently not called by the Dart plugin)

@alexander-doroshko


alexander-doroshko:
Thanks!
Will you need to use it in Flutter plugin?

devoncarew:
We will eventually. If it makes its way back to 2017.3, then I'll use it. Otherwise I can use an untyped variant - `callServiceExtension()` - in the meantime.

alexander-doroshko:
pushed

PR #574, opened 2018-04-17
fix a race condition in the dart debugger against --preview-dart-2 vms

devoncarew:
We 're seeing a race condition when debugging against the dart vm when it's run with `--preview-dart-2`. The main isolate takes much longer to start up now - it's compiling the user's app with the common front end. We request isolates from the VM at startup, but now the main isolate is often in a pre-runnable state, where trying to set a breakpoint will fail.

To repro, create a command-line app, set a breakpoint, and hit run. The app will never auto-resume from start.

This fix:
- addresses an existing issue where the data in the `myIsolateIdToInfoMap` map would be replaced - we call `IsolatesInfo.addIsolate()` twice, and the 2nd time the info was being replaced.
- we do not try and set breakpoints on isolates that are not yet runnable. Those breakpoint attempts would fail; we instead wait for later lifecycle events (like runnable)

@alexander-doroshko


alexander-doroshko:
Thanks a lot, pushed!

PR #575, opened 2018-04-18
added wanneer as one of the dutch translations for when

ienemien:
Cucumber supports 'Wanneer' as a translation for 'When', but the Intellij cucumber plugin does not recognize 'Wanneer' as a keyword yet. I added 'Wanneer' as a Dutch translation for 'when'.

### DIFF
--- a/cucumber/resources/i18n.json
+++ b/cucumber/resources/i18n.json
@@ -554,7 +554,7 @@
 	  "scenario_outline": "Abstract Scenario",
 	  "examples": "Voorbeelden",
 	  "given": "*|Gegeven|Stel",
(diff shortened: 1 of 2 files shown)


JonathanBeekmanNL:
This seems like a good addition. "Wanneer" is a proper translation for "when", more expressive than "Als" in my opinion.

BasLeijdekkers:
Hi @ienemien,
Have you signed the contributor agreement? / Heb je de bijdragerovereenkomst ondertekend?
https://www.jetbrains.com/agreements/cla/

ienemien:
I hadn't yet, but just did that now, thank you!

PR #576, opened 2018-04-19
Dart Plugin- Append '.stagehand' and '.webdev' to the PUB_ENVIRONMENT environment variable to when pub is called.

jwren:
Additionally, append '.activate' if "pub global activate" is being called for stagehand or webdev command.


jwren:
IDE[.stagehand].[activate] would be easier, but then you wouldn't know which IDE.

The DartPubActionBase.getPubEnvValue() gives some IDE name, so we'll have results like the following: <WebStorm | Community_IDEA | Ultimate_IDEA>[.stagehand].[activate].  This way you'll be able see which percent use WebStorm versus other IntelliJ IDEs.

PR #582, opened 2018-05-17
Use 'execution.getSuggestions' to request completion in the 'Evaluate' dialog.

scheglov:
https://github.com/dart-lang/sdk/commit/76a30dd594e06c20fd9be0e2dc8198e7241744a2 provides a basic implementation of this API in Analysis Server. Not all features are implemented at the moment, I will work them in parallel. But even just support for local variables seems useful.

I don't know yet how to access runtime types of variables, someone with better knowledge of debugger API might give hints. @jacob314

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -1605,14 +1605,13 @@ public class DartAnalysisServerService implements Disposable {
     }
   }
 
-  public RuntimeCompletionResult execution_getSuggestions(@NotNull final VirtualFile file,
-                                                          @NotNull final String code,
-                                                          final int _offset,
+  @Nullable
+  public RuntimeCompletionResult execution_getSuggestions(@NotNull final String code,
+                                                          final int offset,
                                                           @NotNull final VirtualFile contextFile,
                                                           final int contextOffset,
                                                           @NotNull final List<RuntimeCompletionVariable> variables,
                                                           @NotNull final List<RuntimeCompletionExpression> expressions) {
-    final String filePath = FileUtil.toSystemDependentName(file.getPath());
     final String contextFilePath = FileUtil.toSystemDependentName(contextFile.getPath());
 
     final AnalysisServer server = myServer;
@@ -1622,7 +1621,6 @@ public class DartAnalysisServerService implements Disposable {
 
     final CountDownLatch latch = new CountDownLatch(1);
     final AtomicReference<RuntimeCompletionResult> refResult = new AtomicReference<>();
-    final int offset = getOriginalOffset(file, _offset);
     server.execution_getSuggestions(
       code, offset,
       contextFilePath, contextOffset,
@@ -1636,7 +1634,7 @@ public class DartAnalysisServerService implements Disposable {
 
         @Override
         public void onError(RequestError error) {
-          logError("execution_getSuggestions()", filePath, error);
+          logError("execution_getSuggestions()", contextFilePath, error);
           latch.countDown();
         }
       });
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -1729,7 +1729,6 @@ public class DartAnalysisServerService implements Disposable {
       }
 
       final DebugPrintStream debugStream = str -> {
-        if (!str.contains("{\"event\":\"analysis.errors\"")) System.out.println("debugStream: " + str);
         str = str.substring(0, Math.min(str.length(), MAX_DEBUG_LOG_LINE_LENGTH));
         synchronized (myDebugLog) {
           myDebugLog.add(str);
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -94,6 +94,7 @@ import java.io.File;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class DartAnalysisServerService implements Disposable {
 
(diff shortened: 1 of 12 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1604,6 +1605,46 @@ public void execution_deleteContext(@NotNull final String contextId) {
     }
   }
 
+  public RuntimeCompletionResult execution_getSuggestions(@NotNull final VirtualFile file,
```
alexander-doroshko: I guess we can get rid of `file` argument. AFAIK dummy files always have normalized line breaks, so no need to convert offset.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -75,15 +77,28 @@ public DartServerCompletionContributor() {
...
+               final Project project = originalFile.getProject();
+
+               final CompletionSorter sorter = createSorter(parameters, originalResultSet.getPrefixMatcher());
+               final String uriPrefix = getPrefixIfCompletingUri(parameters);
+               final CompletionResultSet resultSet = uriPrefix != null
+                                                     ? originalResultSet.withRelevanceSorter(sorter).withPrefixMatcher(uriPrefix)
+                                                     : originalResultSet.withRelevanceSorter(sorter);
+
+               // If completion is requested in "Evaluate Expression" dialog, use runtime completion.
+               if (originalFile.getContext() != null) {
```
alexander-doroshko: `originalFile.getContext() != null` does not necessarily mean that we are inside Evaluate Expression window. There are many other cases when it happens. For example when editing a breakpoint condition.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -139,6 +148,40 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
+    final String code = dummyDocument.getText();
+    final int codeOffset = parameters.getOffset();
+
+    final DartAnalysisServerService das = DartAnalysisServerService.getInstance(project);
+    final RuntimeCompletionResult completionResult =
+      das.execution_getSuggestions(originalFile.getVirtualFile(),
+                                   code, codeOffset,
+                                   contextFile, contextOffset,
+                                   Lists.newArrayList(), Lists.newArrayList());
+    if (completionResult.suggestions != null) {
```
alexander-doroshko: I've got NPE here

scheglov: You probably have a bit older SDK, which does not implement this request yet.
So, we get a timeout and return null.
Of course this is still a bug in this PR - we need to check for null.

I will mark the method nullable as well.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
scheglov: OK, it is probably misnomer then, but the feature is still useful, I think?
It seems to work in breakpoint conditions too.

We will not able to give runtime types of variables and expressions, but it is OK, they are designed to be optional - if we send empty lists for them.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
scheglov: Ah, OK.
I will remove it.
And for `logError` is is better to use `contextFile`.

jacob314:
Here's some ugly example code showing going from a string expression to an DartVmServiceValue to the type name and file location of the value.
 XDebugSession session = XDebuggerManager.getInstance(project).getCurrentSession();
+                     XDebuggerEvaluator evaluator = session.getDebugProcess().getEvaluator();
+                     String expressionToResolve = e.g. "foo" or "foo.bar";
+                     List<SourceEdit> editList = new ArrayList();
+                     final DartVmServiceValue[] vmServiceValue = { null };
+                     if (evaluator != null && expressionToResolve != null) {
+                       System.out.println(" TORESOLVE=#" + expressionToResolve);
+                       evaluator.evaluate(expressionToResolve, new XDebuggerEvaluator.XEvaluationCallback() {
+                         @Override
+                         public void errorOccurred(@NotNull String errorMessage) {
+                           System.out.println("got error:" + errorMessage);
+                           latch.countDown();
+                           // This is fine, the expression just couldn't be evaluated. Continue.
+                         }
+
+                         @Override
+                         public void evaluated(@NotNull XValue result) {
+                           if (result instanceof DartVmServiceValue) {
+                             vmServiceValue[0] = (DartVmServiceValue) result;
+                           }
+                           result.computeTypeSourcePosition((@Nullable XSourcePosition sourcePosition) -> {
+                             VirtualFile file = sourcePosition.getFile();
                               // The analyzer appears to do fine if you import a part instead of the library so this worked for my test.
+                             System.out.println("File = " +file.getPresentableUrl());
// Rest of this snippet is irrelevant as you wouldn't be using imports.
+                             importedElements.add(new ImportedElements(file.getPresentableUrl(),                              latch.countDown();
+                           });

Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/RuntimeCompletionExpressionType.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public static final RuntimeCompletionExpressionType[] EMPTY_ARRAY = new RuntimeCompletionExpressionType[0];
+
+  public static final List<RuntimeCompletionExpressionType> EMPTY_LIST = Lists.newArrayList();
+
+  /**
+   * The path of the library that has this type. Omitted if the type is not declared in any library,
+   * e.g. "dynamic", or "void".
+   */
+  private final String libraryPath;
```
jacob314: For easy integration with the observatory it would be convenient if this functioned ok if a path to the actual source file defining the class was ok rather than strictly the library path. it appears the analyzer does fine with imports to part files so I would expect that isn't much of an issue to support and it is just a matter of documenting that this can be a path to the library or the actual part file in the library defining the type.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
jacob314: For some runtime expressions we can still do a good job even when not paused. See  https://github.com/JetBrains/intellij-plugins/pull/583 for a related CL to enable evaluating expressions when not paused at a breakpoint.

scheglov: Yes, when Alexander told me that context is set also for breakpoint conditions, I tried and can get good completion there too.

scheglov:
I'd like to focus on completion using static types in this PR.
Using runtime types of variables and expressions can be added later.

Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/RuntimeCompletionExpressionType.java
scheglov: Does Observatory provide only the file that defines a class, but not the library?

We can change it to the file path, analyzer knows which libraries include which parts.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
alexander-doroshko: `if (originalFile instanceof DartExpressionCodeFragment)` looks safer to me. Do you mind if I change it this way?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1604,6 +1605,44 @@ public void execution_deleteContext(@NotNull final String contextId) {
...
+      new GetRuntimeCompletionConsumer() {
+        @Override
+        public void computedResult(RuntimeCompletionResult result) {
+          refResult.set(result);
+          latch.countDown();
+        }
+
+        @Override
+        public void onError(RequestError error) {
+          logError("execution_getSuggestions()", contextFilePath, error);
```
alexander-doroshko: I've wrapped it with `if (!RequestErrorCode.UNKNOWN_REQUEST.equals(error.getCode())) {}` to be old SDK-compatible

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
scheglov: Sounds good to me.

alexander-doroshko:
Great feature, thanks! Merged.

jacob314:
Tthere is an easy enough way to get the library via the observatory
protocol so ignore my previous suggestion.

PR #583, opened 2018-05-19
Support evaluate in Dart when not paused at a breakpoint.

jacob314:
Previously you could only evaluate expressions when paused at a breakpoint. Due to the Dart threading model it is completely reasonable to evaluate expressions even when not paused at a breakpoint and this is particularly useful for UI debugging tools such as the Flutter inspector.

The design is to use the currently selected file to indicate what library to evaluate expressions in the context of. We use the first isolate in the application or the last isolate that was paused at a breakpoint.
It could be nice in the future to provide UI to select a library and UI to select an isolate.

I didn't see any existing tests testing the related functionality evaluating code at a breakpoint but if there are tests that I just missed let me know and I'll add analogous tests for this functionality.

I've added clumsy code to resolve finding the library given a part file. If there is a cleaner solution using the analysis server, @scheglov please let me know.

Screenshot showing evaluating expressions in the context of a library on a server app that isn't stopped at a breakpoint.
image


jacob314:
@alexander-doroshko can you please take a look. I believe I don't have permissions to request an intellij-plugins reviewer.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceEvaluatorOnLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public boolean isCodeFragmentEvaluationSupported() {
+    return true;
+  }
+
+  @Override
+  public void evaluate(@NotNull final String expression,
+                       @NotNull final XEvaluationCallback callback,
+                       @Nullable final XSourcePosition expressionPosition) {
```
alexander-doroshko: If you Alt+click or select and Alt+click something in the editor then `expressionPosition` is not `null`. Although it is probably the same as `myFile`, I think it would be more accurate to use it explicitly.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -576,6 +580,36 @@ private static boolean isDartPatchUri(@NotNull final String uri) {
...
+      } else {
+        // A breakpoint hasn't yet been hit but some isolates have been
+        // started. Use the first one.
+        isolateId = getIsolateInfos().iterator().next().getIsolateId();
+      }
+    }
+    Project project = getSession().getProject();
+    FileEditorManager manager = FileEditorManager.getInstance(project);
+
+    VirtualFile[] files = manager.getSelectedFiles();
```
alexander-doroshko: Speaking of the context file selection: I'm not sure there's an easy way to add special UI for this purpose. But it may make sense to move logic that takes the file from the current editor right to the `evaluate()` method. Users may switch editor tabs without closing the Evaluate Expression dialog.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceEvaluatorOnLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+                                         @NotNull final Project project) {
+    myDebugProcess = debugProcess;
+    myIsolateId = isolateId;
+    myFile = file;
+    myProject = project;
+  }
+
+  @Override
+  public boolean isCodeFragmentEvaluationSupported() {
+    return true;
```
alexander-doroshko: It's about multi-line field for the evaluated expression. Do you think we need it? If so, should it be enabled in `DartVmServiceEvaluator` as well?

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceEvaluatorOnLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+  public ExpressionInfo getExpressionInfoAtOffset(@NotNull final Project project,
+                                                  @NotNull final Document document,
+                                                  final int offset,
+                                                  final boolean sideEffectsAllowed) {
+    final PsiFile psiFile = PsiDocumentManager.getInstance(project).getPsiFile(document);
+    final PsiElement contextElement = psiFile == null ? null : psiFile.findElementAt(offset);
+    return contextElement == null ? null : DartVmServiceEvaluator.getExpressionInfo(contextElement);
+  }
+
+  private static VirtualFile getLibraryFile(PsiFile psiFile) {
```
alexander-doroshko: Feel free to reuse `DartResolveUtil.findLibrary()` instead, it seems to be equivalent.

jacob314: Thanks! I switched to the DartResolveUtil helper.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceEvaluatorOnLibrary.java
jacob314: I think the multi-line field is nice for entering complex expressions Dart VM handles multi line expressions fine although it could be a bit smarter. 
For example the following works:
image
But the following currently generates an error.
image
I'm also fine with keeping multi-line off until the second case is also supported.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
jacob314: Done. I agree that moving the calculation to right at the evaluate call is least surprising.

alexander-doroshko:
@jacob314, looks like this PR has not yet been updated

jacob314:
Patch coming shortly. I haven't addressed all the comments yet. Fixing a case where triggering an exception caused annoying breakpoint behavior.

jacob314:
Please take another look. I've uploaded the new patch.
You can now right click to "evaluate expression" even static class members and have everything just work. Mouse over of static class members also works. Everything from the library context is still in scope within a class context so there is no harm in doing this.
image
Note that you still can't hover over static class members when stopped at a breakpoint unless you are paused at a breakpoint in that class as the Dart vm service protocol doesn't yet handle that case.

I also refactored renaming the classes as DartVmServiceEvaluator and DartVmServiceEvaluatorInFrame where DartVmServiceEvaluatorInFrame subclasses DartVmServiceEvaluator as that reduces duplicate code and there is no need to keep the classes separate now that DartVmServiceEvaluator doesn't reference a specific library id.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceEvaluator.java
```diff
@@ -1,38 +1,176 @@
...
+                       @Nullable final XSourcePosition expressionPosition) {
+    final String isolateId = myDebugProcess.getCurrentIsolateId();
+    final Project project = myDebugProcess.getSession().getProject();
+    final FileEditorManager manager = FileEditorManager.getInstance(project);
+    PsiElement element = null;
+    PsiFile psiFile = null;
+    final List<VirtualFile> libraryFiles = new ArrayList<>();
+    // Turn off pausing on exceptions as it is confusing to mouse over an expression
+    // and to have that trigger pausing at an exception.
+    myDebugProcess.getVmServiceWrapper().setExceptionPauseMode(ExceptionPauseMode.None);
```
alexander-doroshko: I agree that with the current VM behavior we do need to disable pausing on exceptions during evaluation. Maybe we'd also ask VM Service guys to enable such behavior by default? VM knows when an exception happens during program execution and when - during evaluation, so it probably could always ignore exceptions on evaluation.

jacob314: Agreed. I've asked the VM Service team to provide a service protocol option to enable that behavior.

alexander-doroshko:
Pushed, thank you!

PR #584, opened 2018-06-11
Update Error-prone to 2.3.1

PromanSEW:


luengnat:
any plan to get this in and get it released?

PromanSEW:
@luengnat I think no, because they commit directly from IDEA, and don't monitor GitHub repository at all as I see.
😕

chashnikov:
Merged manually, thank you for the contribution!

PR #585, opened 2018-06-27
re-gen the analysis server lib

devoncarew:
@alexander-doroshko, it's been a while since we committed a new version of the analysis server generated files. This updates the lib. We may at some point want to look at how we're generating this, as there are patches to the generated files, and we're generating less than we could.


alexander-doroshko:
Thanks, pushed!

Just FYI: recently I fixed a very important problem in the analysis server lib: https://github.com/JetBrains/intellij-plugins/commit/aa15f2d23ef6ebcb25df784a979eee429416d575. It didn't handle some errors, which lead to temporary IDE freeze.

PR #589, opened 2018-10-03
New action for Angular Dart developers: Open Sibling Angular Dart File.

jwren:
@alexander-doroshko

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -263,6 +263,9 @@
             text="Sort members in Dart File" description="Sort members in your Dart code">
       <add-to-group group-id="CodeFormatGroup" anchor="after" relative-to-action="Dart.DartStyle"/>
     </action>
+    <action id="OpenSiblingAngularDartFile" class="com.jetbrains.lang.dart.ide.actions.OpenSiblingAngularDartFile"
+            text="Open Sibling Angular Dart File">
+    </action>
     <action id="Generate.Constructor.Dart" class="com.jetbrains.lang.dart.ide.generation.DartGenerateConstructorAction"
             text="Constructor">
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
We are going to discuss the feature internally. I'll keep you posted.

alexander-doroshko:
We think that such action in its current state is kind of contradictory an too specific.

On the one hand it is not discoverable for AngularDart developers. Given that there are <1% of AngularDart developers within IntelliJ/WebStorm user base, and perhaps only 10% of them will notice the feature and only 5% of those who notice will assign a custom shortcut. Without custom shortcut the feature is not useful because it doesn't save keystrokes. And for such a specific feature we definitely can't asign a default shortcut. So only maybe a dozen of users will eventually benefit from the feature.

On the other hand the action is visible/enabled for everyone (via Find Action). And it works somehow for everyone. In a regular web app it allows to jump between foo.scss and its compiled foo.css. As well as between bar.html and bar.css. People might have no idea what is Dart, and what is Angular Dart but the feature called "Open Sibling Angular Dart File" works somehow for them. And I guess the number of confused developers will be greater than the number of those who benefit from the feature.

Next, IntelliJ and WebStorm support hundreds of technologies. And half of these technologies assume that there may be sibling files with the same name and different extensions. [dart, html, scss, css] is definitely not the most popular set. So we should probably look for a general solution. And there are some internal discussions and experiments in this area. One idea is to contribute to the existing 'Go To Related File' feature. Others are experimenting with split (or combined) editors. Also, there's a 'Nested files' feature (configurable under gear in the Project View), so it's possible to group same-named files and there's a request to jump between them (though not implemented yet). 

So I'd rather we think about some general solution rather than very-very specific.

jwren:
No worries.  Typically I'd discuss, before implementing, but this work was already done so I just created a PR to share and discuss.

PR #590, opened 2018-10-04
Update README.md

devoncarew:
@alexander-doroshko

### DIFF
--- a/Dart/README.md
+++ b/Dart/README.md
@@ -43,7 +43,7 @@ Prerequisites:
 
 4. Open intellij-community project, compile it.
    Open File | Project Structure | Modules | `[+]` | Import Module, select intellij-plugins/Dart/Dart-community.iml.
-   In the same Project Structure dialog open the Dependencies tab of the `intellij.idea.community.main` module,
+   In the same Project Structure dialog open the Dependencies tab of the `intellij` > `idea.community` > `main` module,
    click `[+]` at the bottom (Mac) or right (Win/Linux) to add a module dependency on the `Dart-community` module.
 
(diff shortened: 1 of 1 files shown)


Dart/README.md
```diff
@@ -43,7 +43,7 @@ Prerequisites:
 
 4. Open intellij-community project, compile it.
    Open File | Project Structure | Modules | `[+]` | Import Module, select intellij-plugins/Dart/Dart-community.iml.
-   In the same Project Structure dialog open the Dependencies tab of the `intellij.idea.community.main` module,
+   In the same Project Structure dialog open the Dependencies tab of the `intellij` > `idea.community` > `main` module,
```
alexander-doroshko: Strictly speaking, the name of the module is `intellij.idea.community.main`. Tree presentation can even be disabled in the context menu (Flatten Modules). How about 
> open the Dependencies tab of the `intellij.idea.community.main` module (`intellij` > `idea.community` > `main` node)

or something else, both correct and clear?

devoncarew: Makes sense. I needed to do some digging in the tree structure to find the right module, so calling out both the name and the location would be useful

devoncarew:
updated -

alexander-doroshko:
Pushed, thanks!
BTW are there any updates to the VM Service files that we should take?

devoncarew:
Yes! They're on their way :)

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/3d5c204cc1d2582a12305a96232f2c4296a391bc

PR #595, opened 2018-11-01
upgrade the service protocol library to 3.10

devoncarew:
- upgrade the service protocol library to 3.10

This exposes the new `isolate.getScripts()` call (not yet available in shipping Dart SDKs).

@jwren @alexander-doroshko


jwren:
LGTM, thanks @devoncarew

PR #596, opened 2018-11-06
adjust the Dart analysis server crash report template

devoncarew:
- this updates the Dart crash report template slightly - it'll automatically apply the `area-analyzer` and `analyzer-crash-report` labels

@stereotype441, @alexander-doroshko, @jwren


alexander-doroshko:
I'm not sure this will work. Label management requires permissions that users (including me) do not have.

devoncarew:
Ah! It worked for me, but I do have the right permissions to add labels.

Perhaps we could look at using an explicit github issue template (https://help.github.com/articles/about-issue-and-pull-request-templates/). That does have downsides, in that once one is defined, we need to route people into one of a few templates.

PR #597, opened 2018-11-13
New package-scope analysis setting in the Dart Analysis view

jwren:
@alexander-doroshko @devoncarew @lambdabaa

PR for the initial scoped analysis.  This change in behavior is not enabled by default.

package-scoped-analysis

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -229,6 +229,12 @@ group.by.severity.description=Group by severity: errors in the top of the table,
 filter.problems=Filter
 filter.problems.description=Filter problems shown in the Dart Analysis tool window
 reset.filter=Reset filter
+analysis.server.settings=Analyzer settings
+analysis.server.settings.description=Dart Analysis Server settings tool window
+analysis.server.settings.title=Dart Analysis Server Settings
+analysis.server.settings.sdk.settings=Dart plugin settings
+analysis.server.settings.diagnostics=View analyzer diagnostics...
+analysis.server.settings.package.analysis=Scope analysis to the current package
 
 dart.sort.members.action.name=Sort Members in Dart File
 dart.sort.members.action.name.ellipsis=Sort Members in Dart files...
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
@@ -37,7 +37,7 @@ public class DartServerRootsHandler {
   private final List<String> myIncludedRoots = new SmartList<>();
   private final List<String> myExcludedRoots = new SmartList<>();
 
-  private DartProblemsViewSettings.ScopedAnalysisMode myScopedAnalysisMode = DartProblemsViewSettings.ScopedAnalysisMode.All;
+  DartProblemsViewSettings.ScopedAnalysisMode myScopedAnalysisMode = DartProblemsViewSettings.ScopedAnalysisMode.All;
 
   @Nullable private VirtualFile myCurrentFile;
 
@@ -188,7 +188,7 @@ public class DartServerRootsHandler {
 
     // TODO(jwren) follow up to extract method with duplicate code (from DartProblemsPresentationHelper)
     final VirtualFile pubspec = Registry.is("dart.projects.without.pubspec", false)
-                                ? DartBuildFileUtil.findPackageRootBuildFile(myProject, vFile, true)
+                                ? DartBuildFileUtil.findPackageRootBuildFile(myProject, vFile)
                                 : PubspecYamlUtil.findPubspecYamlFile(myProject, vFile);
 
     // temp var to make sure that value is initialized
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
@@ -5,11 +5,9 @@ import com.intellij.ProjectTopics;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.roots.ContentEntry;
-import com.intellij.openapi.roots.ModuleRootEvent;
-import com.intellij.openapi.roots.ModuleRootListener;
-import com.intellij.openapi.roots.ModuleRootManager;
+import com.intellij.openapi.roots.*;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.search.FilenameIndex;
@@ -17,10 +15,14 @@ import com.intellij.util.PathUtil;
 import com.intellij.util.SmartList;
 import com.intellij.util.io.URLUtil;
 import com.jetbrains.lang.dart.DartStartupActivity;
+import com.jetbrains.lang.dart.ide.errorTreeView.DartProblemsViewSettings;
 import com.jetbrains.lang.dart.sdk.DartSdk;
 import com.jetbrains.lang.dart.sdk.DartSdkLibUtil;
+import com.jetbrains.lang.dart.util.DartBuildFileUtil;
+import com.jetbrains.lang.dart.util.PubspecYamlUtil;
 import gnu.trove.THashSet;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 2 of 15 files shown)


Dart/src/com/jetbrains/lang/dart/DartBundle.properties
```diff
@@ -229,6 +229,12 @@ group.by.severity.description=Group by severity: errors in the top of the table,
 filter.problems=Filter
 filter.problems.description=Filter problems shown in the Dart Analysis tool window
 reset.filter=Reset filter
+analysis.server.settings=Analyzer settings
+analysis.server.settings.description=Dart Analysis Server settings tool window
+analysis.server.settings.title=Dart Analysis Server Settings
+analysis.server.settings.sdk.settings=Dart plugin settings
+analysis.server.settings.diagnostics=View analyzer diagnostics...
+analysis.server.settings.package.analysis=Enable Package-scoped analysis
```
devoncarew: `Scope analysis to the current package`?

devoncarew:
I'd suggest some slight changes to the UI:

`[ ] Scope analysis to the current package`
`Open Dart plugin settings`
---
`View analyzer diagnostics...`

alexander-doroshko:
I've played a bit without much reading code. I've been watching `analysis_setAnalysisRoots()` requests sent to the server. Here are the first comments:
- Convert paths to OS-dependent when adding them to `newIncludedRoots` collection. Nothing worked for me until I fixed it locally. 
- With the new option enabled the `newExcludedRoots` list is always empty. This way the Analyzer starts looking into `build`, `.idea`, `.pub`, `.dart_tool` and manually excluded folders. Shouldn't the list be the same as when the option is disabled?
- As I switch from one file to another within the same Dart package the whole list of errors in the Dart Analysis view sporadically disappears and appears again. (The new option is on, filters are off, no  `analysis_setAnalysisRoots()` calls during this experiment)
- Does the Platform give a chance to highlight errors/warnings right in the editor for non-project files?

Speaking about root by file detection:
- The fallback is the immediate enclosing folder - that's ok.
- The first attempt to guess the root I think should be trying to find `pubspec.yaml` in one of the parent folders, shouldn't it?
- The algorithm is looking for `lib` folder in some of the parent folders. Well, maybe that's the best guess we can make, I don't know pubspecless project specificity. But I'd be happy if it's possible to find some more specific distinctive features of the dart packages without pubspecs. Like, I don't know, both `lib` and `.dart_tool` folders presence and at the same time `BUILD` and `.packages` and `.gitignore` files presence.

jwren:
Thank you for testing, and for the comments. I'll go through and update the PR.

`build`, `.idea`, `.pub` comment is true, they should be "excluded".

Your comment about the `lib` logic- yes, we should try to use presence of other artifacts first, `.packages`, `.dart_tool`, etc, using the current logic as a fall back.

I'll update the PR soon.

jwren:
Fixing up this PR, see https://github.com/JetBrains/intellij-plugins/pull/626

jwren:
Inline wasn't trivial.  Responses in same bullet point format.

I didn't squash the two commits so that you can see the diff.

- Done
- Done, newExcludedRoots set now matches with and without the new setting, although I don't see where build, .pub, .dart_tool are included in this list
- Figured it out. This was due to isInIncludedRoots returning incorrectly for the Windows OS from DartProblemsView.setCurrentFile(), calling clearAll when it shouldn't.  A needed call to FileUtil.toSystemDependentName() in DartServerRootsHandler.isInIncludedRoots should fix the issue for Windows.
- I assume you are referencing the changes to DartAnnotator?  Yes, when opening a random dart file (some non-project file) on disk with this scoped analysis set to package-scope, the errors and warnings view as well as the errors in the editor are analyzed by the DAS and displayed.

- I agree
- Done- I found the code in DartProblemsPresentationHelper, and followed suite.  It is more consistent with the rest of the plugin, and assuming this looks good to you, I can follow up to extract the C&P into its own method in one of the utility classes (TODO added to this effect.)

alexander-doroshko:
> - although I don't see where build, .pub, .dart_tool are included in this list

Try any standard project with pubspec, you'll see where these excluded folders come from :)
> - the errors and warnings view as well as the errors in the editor are analyzed by the DAS and displayed

Great, now I see them. Something didn't work after the 1st iteration

> - I found the code in DartProblemsPresentationHelper, and followed suite. It is more consistent with the rest of the plugin, and assuming this looks good to you

If `BUILD` file presence is the marker you need then that's just great! You'll only need to tune the algorithm of searching pubspec/BUILD file a bit, as of now, it doesn't work for non-project files.

I'll start in depth code review next week. 

One more thought that doesn't require immediate attention. Once we've found `getEnclosingDartPackageDirectory()` and it happened to be outside of the project should we auto-exclude `.dart_tool`, `.pub` and `build` subfolders? We already do it for in-project packages so we might want to do it for non-project packages too.

Minor: there are 2 files that effectively do not contain any change. Feel free to remove them from the commit.

jwren:
I rebased to be up to date, took care of the minor comment (files modified in just a white space location), and sqaushed the commits together.

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
```diff
@@ -108,6 +138,55 @@ private void updateRoots() {
...
+
+  @NotNull
+  public String getEnclosingDartPackageDirectory(@NotNull final VirtualFile vFile) {
+    // temp var to make sure that value is initialized
+    final VirtualFile packageRoot;
+
+    // TODO(jwren) follow up to extract method with duplicate code (from DartProblemsPresentationHelper)
+    final VirtualFile pubspec = Registry.is("dart.projects.without.pubspec", false)
+                                ? DartBuildFileUtil.findPackageRootBuildFile(myProject, myCurrentFile)
+                                : PubspecYamlUtil.findPubspecYamlFile(myProject, myCurrentFile);
```
alexander-doroshko: > You'll only need to tune the algorithm of searching pubspec/BUILD file a bit, as of now, it doesn't work for non-project files.

jwren:
Fixed, thanks I thought that I tested that.  If fixed it by adding a boolean parameter to `DartBuildFileUtil.findPackageRootBuildFile(...)`

jwren:
Rebased and pushed to be merge-able (conflicts from StartupActivity-change resolved.)

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
alexander-doroshko: I see yellow code here. Did you mean `vFile`?

BTW shouldn't `findPubspecYamlFile()` logic depend on the analysis mode as well? Maybe both these methods could check for the current analysis mode without explicitly passed parameter?

alexander-doroshko:
I noticed that (with the feature enabled) analyzed root is changed if you navigate to SDK or 3rd-party package file. Is it expected? I thought analyzed scope shouldn't change if we navigate to a file from `fileIndex.isInLibraryClasses(file)`.

jwren:
That behavior doesn't surprise me.  I didn't consider/ test that before. I could see arguments both directions.  If I am looking into some imported package, wouldn't it make sense to see errors/warnings for that package, just like for libraries in my project?  I don't have a big preference here, I can see users asking for either choice here.  If you have a preference I'll experiment with it.

I suspect with things like this I'll be iterating on it a bit after users get their hands on it.

jwren:
Yes, I agree.  I played around with it, I made the fileIndex.isInLibraryClasses(file) change and rebased.

alexander-doroshko:
Cool, thanks! The PR is close to being merged. I have an idea about UI, I'll email you tomorrow. 

What do you think about `findPubspecYamlFile()`? Shouldn't its logic depend on the analysis mode, the same as  `findPackageRootBuildFile()`? Maybe both these methods could check for the current analysis mode without an explicitly passed parameter?

alexander-doroshko:
The use case: enable the 'package-scoped analysis' feature and open a file from some package outside of the project (not from SDK or dependencies, just from a separate dart project). 

To make it work for pubspecless projects you've added `doConsiderNonProjectFiles` argument to the `findPackageRootBuildFile()` method. But the feature still doesn't work for packages with pubspec because `findPubspecYamlFile()` doesn't work for non-project files.

So the suggestion is to make it work. And I thought it might be nicer from the code style point of view not to pass `doConsiderNonProjectFiles` as a parameter, but instead `findPackageRootBuildFile()` and `findPubspecYamlFile()` methods would somehow check the state of the 'package-scoped analysis' feature themselves.

jwren:
I just touched the PR again with some of the UI nits discussed in email.

As for the `findPubspecYamlFile()` question above... yes, I think.  The `findPubspecYamlFile()` is called by lots of places in the Dart Plugin, and assuming that we don't just want to have an equivalent `doConsiderNonProjectFiles` argument which we use sparingly (like I have done to date), I'm a bit worried about creating side effects.

I'll follow up tomorrow with some additional thoughts here.

alexander-doroshko:
I looked through the `findPubspecYamlFile()` usages, at first glance it looks safe to skip `fileIndex.isInContent(parent)` check if package-scoped analysis is enabled. I'll test thoroughly when it lands.

jwren:
Rebased and updated this PR with this change; do test it on your end.  Instead of squashing this last commit, I left it as it's own commit that you can squash if you want, but I wanted you to be able to see the change explicitly.

alexander-doroshko:
Thanks! I'll test and merge it tomorrow.

alexander-doroshko:
In progress...

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/15682f06af99ead5a08f33074d8a5a60cc9cad06
Landed with some minor changes! Please review and test it once again on your end.
The only issue I found so far was on the analysis server end: https://github.com/dart-lang/sdk/issues/35452
The biggest change I made was the current file and the feature state tracking. They were tracked both in `DartRootsHandler` and in `DartProblemsPresentationHelper`/`DartProblemsViewSettings`. I removed duplicating fields from `DartRootsHandler`.

PR #598, opened 2018-11-19
Introduced option to extract META-INF and OSGI-INF folder from generated jar file into modules output folder

lauritziu:
# Context of the pullrequest

We are currently working on a Eclipse E4 RCP application, which is completely build by Maven and Tycho and we would like to have support for that approach in
 Intellij because it is our preferred IDE. Eclipse in its own is a pure OSGI application.

The only plugin we've found to support the development of a OSGI based Eclipse application was the Osmorc plugin. Therefore we've choosen Osmorc to start the
development.

# Our development process

* have a Maven based module
* have a BND file in the root directory of each module
* use these [Osgi plugins](https://github.com/BestSolution-at/maven-osgi-plugin) to package and start a Eclipse application
* configure the Bnd maven plugin like this

```
<plugin>
    <groupId>biz.aQute.bnd</groupId>
    <artifactId>bnd-maven-plugin</artifactId>
    <version>3.3.0</version>
    <executions>
        <execution>
            <id>default-bnd-process</id>
            <goals>
                <goal>bnd-process</goal>
            </goals>
        </execution>
    </executions>
    <configuration>
        <skip>false</skip>
    </configuration>
</plugin>
```

* the plugin execution generates the Manifest and DS Component definitions under target/classes/META-INF
* this will be part of the final packaged jar file
* all that works like any other maven based module
* if the module/application is started, these generated information are considered and the bundle is useable in the launched OSGI framework
* if we start the application from Intellij itself, a maven run configuration is used where "Resolve Workspace" option is enabled.
* the whole bundle content comes from _target/classes_ directory

To support that approach we can benefit from the Osmorc Builder to start the Bnd generation process before the application is started. So our assumption is
that everything that is part at the bundle is located under _target/classes_.

# Plugin requirements for our development approach

1. let Bnd process the bnd file in each module and generate the stuff under _target/classes_

2. do that generation within the step "Build project" as part of a run configuration
3. let inspections work on the generated Manifest.mf files instead of the source Bnd file

With the changes provided by the pullrequest we can fulfill 1. and 2. but not 3. (is a lower priority)

# Customization of the Osmorc plugin to meet our needs

We introduced a new option in the facet configuration dialog for each module. This option is enabled by default and extracts everything under META-INF and
OSGI-INF from the bundle jar into the module output path. That is for Maven project _target/classes_.

With that in place we get our requirements fulfilled. Every build creates a bundle jar (if `private-package` instruction in the Bnd file references all
packages) and the Manifest.mf and DS component defintions are extacted and are visible at runtime.

# Questions

That is the way we've gone, but it is a good one? if yes, we would like to have the stuff included into the main development line to be in sync with the
upcoming changes of the Osmorc plugin. There are some other questions:

 * any idea to improve this approach?
 * Should this simple approach be converted (somehow) to something that matches the Bndtools idea better? Any hints for that?
 * How to use inspections for export/import packages if the Manifest is generated by Bnd? The main point we see here is that the current inspections are
 working on the plain Bnd file. A bnd file can have a lot of makros and these makros are generating, e.g. import package statements, at build time. How can
 an Intellij inspection use these generated information?
 * Why does the manifest generation behaves differently if using osmorc instead of maven bnd plugin? For Osmorc the `private-package` statement in the bnd  file is needed and need to reference all packages that should be analyzed.  With the maven bnd plugin it is not needed.

Thanks for your time to review the pullrequest.

### DIFF
--- a/osmorc/doc/getting-started.txt
+++ b/osmorc/doc/getting-started.txt
@@ -2,36 +2,76 @@
  Quick Instructions for new developers of Osmorc
 =======================================================
 
-1. Get the IntellIJ Open API:
-   - download: https://www.jetbrains.com/idea/download/index.html#kit
-   - unzip to some folder "devel/openapidoc"
+1. Get the latest IDEA EAP
+   - Download it from the EAP page https://www.jetbrains.com/idea/nextversion/ and install <idea-eap-folder>
+   - remove osmorc.jar and osmorc-jps-plugin.jar from the <idea-epa-folder>/Contents/plugins/osmorc
 
-2. Get the latest IDEA build (EAP) for JDK 1.5
-   - Download it from the EAP page
-   - unzip to some folder "devel/idea-jdk15"
-   - remove plugins/osmorc folder from that IDEA installation (if it exists)
+2. Launch IDEA EAP 
+   - Create New Project -> IntelliJ Platform Plugin
+   	- delete folder META-INF and src in the generated project
+   - Project SDK -> New IntellIj Platform SDK
+	   - Select <idea-eap-folder> and Finish
+   - Projectstructure -> Project Settings
+	   - Libraries: add Java libs
+		 + bndlib -> <idea-eap-folder>/Contents/plugins/osmorc/lib/biz*.jar
+		 + bundelor-all -> <idea-eap-folder>/Contents/plugins/osmorc/lib/bundlor-all.jar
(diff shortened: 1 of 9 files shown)


klikh:
Have you submitted a contributor license agreement? If not, please follow the steps described at http://www.jetbrains.com/agreements/cla/ to sign it.

christianheilmann:
Please close this pull request.

I have done a pull request with the same content on actual master:
https://github.com/JetBrains/intellij-plugins/pull/663

PR #599, opened 2018-11-20
url encode analyzer bug reports

devoncarew:
This url encodes the data we use to create a url for analyzer issue reports. This should hopefully address problems we've seen with analyzer bug reports without exception information.

- fix https://github.com/flutter/flutter-intellij/issues/2611

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
@@ -26,6 +26,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 
 public class DefaultDartFeedbackBuilder extends DartFeedbackBuilder {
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DefaultDartFeedbackBuilder.java
```diff
@@ -37,29 +38,32 @@ public String prompt() {
...
+        body += DartBundle.message("dart.feedback.file", file.getAbsolutePath());
       }
       catch (IOException e) {
         // ignore it
       }
     }
-    openBrowserOnFeedbackForm(urlTemplate, project);
+
+    //noinspection deprecation
+    openBrowserOnFeedbackForm(url + URLEncoder.encode(body), project);
```
alexander-doroshko: I'd rather we don't become the only client of this deprecated method in the whole IntelliJ codebase :)

alexander-doroshko:
Do you think this fixes both https://youtrack.jetbrains.com/issue/WEB-35878 and https://youtrack.jetbrains.com/issue/WEB-28443?

devoncarew:
Ah, I didn't see the youtrack issues (cc @stevemessick); yes, I think it addresses those.

PR #626, opened 2018-11-27
DartServerRootsHandler- have the resulting set from getExcludedPackageSymlinkUrls(module) be included in the excluded roots List.

jwren:
@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerRootsHandler.java
@@ -92,6 +92,10 @@ public class DartServerRootsHandler {
                 newExcludedRoots.add(FileUtil.toSystemDependentName(VfsUtilCore.urlToPath(excludedUrl)));
               }
             }
+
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Why do you think we need it?
As I can see 4 lines above we explicitly skip excluding such URLs. According to the commit message it was done on purspose. See https://youtrack.jetbrains.com/issue/WEB-16723.

jwren:
The method that computes this list is named "getExcludedPackageSymlinkUrls", after reading through it yesterday I was convinced that it was mistakenly not added to the excluded list.  Apparently the "excluded" in "getExcludedPackageSymlinkUrls" is meant to convey an exclusion from the exclusion list.... I might add a comment so I don't make this mistake again.

I almost wrote an email asking this question, but a PR was faster.  Thank you Alex.

alexander-doroshko:
`getExcludedPackageSymlinkUrls()` indeed returns a list of URLs that are excluded in terms of IntelliJ IDEA project model (before .packages era)). But we don't have to send this list to the Analysis Server because the Server knows about them itself and doesn't traverse millions of symlinks. Yes, we exclude excluded just to save IDE <-> server traffic.

PR #627, opened 2018-11-29
Change in Dart refactorings & assists to use write-ability of files, instead of their existence in the IJ notion of a project, in applying refactorings

jwren:
@alexander-doroshko -

Chatted with analyzer team (@scheglov @stereotype441) and we convinced ourselves that this is probably the best course of action.  Let me know if you have any questions.

Thanks Alex!

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
+++ b/Dart/src/com/jetbrains/lang/dart/assists/AssistUtils.java
@@ -154,9 +154,9 @@ public class AssistUtils {
   }
 
   @NotNull
-  public static Map<VirtualFile, SourceFileEdit> getContentFilesChanges(@NotNull final Project project,
-                                                                        @NotNull final SourceChange sourceChange,
-                                                                        @NotNull final Set<String> excludedIds)
+  private static Map<VirtualFile, SourceFileEdit> getContentFilesChanges(@NotNull final Project project,
+                                                                         @NotNull final SourceChange sourceChange,
+                                                                         @NotNull final Set<String> excludedIds)
     throws DartSourceEditException {
 
     final Map<VirtualFile, SourceFileEdit> map = Maps.newHashMap();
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
`VirtualFile.isWritable()` is effectively always `true` at least on Windows. By default neither Dart SDK nor 3-rd party packages are read-only.
Well, probably all Analysis Server bugs are fixed and it doesn't suggest to edit SDK/packages files anymore. But I'd be happy if we manage to find some more specific condition that will make required non-project files eligible for refactoring and at the same time will keep SDK/packages files protected. Any chance?

jwren:
Modified, rebased & squashed.

I talked with the analyzer developers about this, and yes, the [VirtualFile.isWritable()]  usage was a proxy for other information.  If it is not accurate for Windows, that is a problem.

How about the updated source?  Instead of disallowing refactorings completely if they have an edit outside of the project, it simply doesn't include that particular edit.  At the same time as this PR lands, I just created https://github.com/dart-lang/sdk/issues/35301 to have the analyzer folks fix it on their end in at least the same way.

When the analyzer has been updated in this way, we can revisit to remove this check, or we can leave it.

alexander-doroshko:
>  If it is not accurate for Windows, that is a problem.

It's accurate on any OS, it's just doesn't do what we want (according to its doc) if I'm getting the problem right.

> How about the updated source?

On the one hand, it looks safe (in terms that non-project files are not touched).
On the other hand, does it solve the initial problem? I thought you wanted to treat some non-project as being special and eligible for refactoring?

Sorry, I'm kind of blind when we are talking about special support for Google internal projects. I may be getting the problem wrong and have nothing to test the suggested solution.

jwren:
> > If it is not accurate for Windows, that is a problem.
> 
> It's accurate on any OS, it's just doesn't do what we want (according to its doc) if I'm getting the problem right.
> 
> > How about the updated source?
> 
> On the one hand, it looks safe (in terms that non-project files are not touched).
> On the other hand, does it solve the initial problem? I thought you wanted to treat some non-project as being special and eligible for refactoring?

Actually **not**-eligible for refactoring. Currently[1] the DAS returns a set of sources that are in, and potentially **not** in the project, the refactoring fails completely if there are any sources outside of the project.  This PR makes it so that only the set of changes in the project will be applied.


[1] - https://dart-review.googlesource.com/c/sdk/+/86040 - when this lands this PR will simply make the Dart plugin robust against the DAS in this way.

> 
> Sorry, I'm kind of blind when we are talking about special support for Google internal projects. I may be getting the problem wrong and have nothing to test the suggested solution.

alexander-doroshko:
Ah, got it, thanks! Will push in the morning.

PR #629, opened 2018-12-12
trim text from VM stack names

devoncarew:
- trim some undesired text from the front of VM stack frames

screen shot 2018-12-11 at 10 46 21 am

@alexander-doroshko


jwren:
LGTM

PR #633, opened 2019-01-16
[WEB-23645] Show package info in Dart code completion

lambdabaa:
This patch updates the Dart plugin to use CompletionSuggestion#elementUri,
added to analysis server in https://dart-review.googlesource.com/c/sdk/+/88744
and https://dart-review.googlesource.com/c/sdk/+/89441, to provide class
package info whenever a class is being suggested and the elementUri is available.

6lvcsg5a504

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -322,8 +322,9 @@ public class DartServerCompletionContributor extends CompletionContributor {
     final String lookupString = suggestion.getCompletion();
     LookupElementBuilder lookup = LookupElementBuilder.create(lookupObject, lookupString);
 
-    if (suggestion.getDisplayText() != null) {
-      lookup = lookup.withPresentableText(suggestion.getDisplayText());
+    String displayText = suggestion.getDisplayText();
+    if (displayText != null) {
+      lookup = lookup.withPresentableText(displayText);
     }
 
     // keywords are bold
@@ -448,6 +449,11 @@ public class DartServerCompletionContributor extends CompletionContributor {
           });
         }
       }
+    }
+
+    // If this is a class, try to show which package it's coming from.
+    if (element != null && element.getKind().equals(ElementKind.CLASS) && suggestion.getElementUri() != null) {
+      lookup = lookup.appendTailText(" (" + suggestion.getElementUri() + ")", true /* grayed */);
     }
 
     // Use selection offset / length.
--- a/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/CompletionSuggestion.java
+++ b/Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/CompletionSuggestion.java
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+ * Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
  * for details. All rights reserved. Use of this source code is governed by a
  * BSD-style license that can be found in the LICENSE file.
  *
- * This file has been automatically generated.  Please do not edit it manually.
+ * This file has been automatically generated. Please do not edit it manually.
  * To regenerate the file, use the script "pkg/analysis_server/tool/spec/generate_files".
  */
 package org.dartlang.analysis.server.protocol;
@@ -58,6 +58,12 @@ public class CompletionSuggestion {
    * displayed text should be different than the completion. Otherwise it is omitted.
    */
   private final String displayText;
+
+  /**
+   * The URI of the element corresponding to this suggestion. It will be set whenever analysis server
+   * is able to compute it.
+   */
+  private final String elementUri;
 
   /**
    * The offset, relative to the beginning of the completion, of where the selection should be placed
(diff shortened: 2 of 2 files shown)


alexander-doroshko:
👍 
Merged: https://github.com/JetBrains/intellij-plugins/commit/41cb252cfd71703f8341a086e2ddef314dfd1163

alexander-doroshko:
@lambdabaa I noticed that for local files URI is too long and looks not as useful as in case of `dart:` and `package:` imports:
image
What about making it shorter? 
I'm not sure whether this should be done on server end or on client end. 
// cc @scheglov

lambdabaa:
> What about making the text in parentheses shorter?

I think it's a good idea. We had talked about only including the package `package:foo` instead of the package and library `package:foo/bar.dart`. I think it would probably make sense to do this server-side for consistency across clients, wdyt @scheglov?

scheglov:
I think that just `package:foo` is not enough, there might be multiple libraries in the same package, exporting symbols with the same name. For example `analyzer` :-)

I also noticed that `file://` URIs are too long, mostly when you write some tests. Given that we use `file://` URIs only for relative imports, inside `test`, or `bin`, and add new `import` directive with relative imports - we should display relative URIs as well. I plan to add optional `displayUri` field to `IncludedSuggestionSet`, so that `file://` libraries will get nicer URIs to display.

alexander-doroshko:
> I plan to add optional `displayUri` field to `IncludedSuggestionSet`, so that `file://` libraries will get nicer URIs to display.

👍

scheglov:
With https://github.com/dart-lang/sdk/commit/0ef30fb86e02c8103394e67c0360bd1d416a3a12, https://github.com/dart-lang/sdk/commit/56356d426af75f14477d6853c4fd6ce023cb6735, and the corresponding quick hack for Dart plugin (which is trivial, so I will not send PR, so please implement it), we can show relative file URIs.

image

alexander-doroshko:
@scheglov Shouldn't `getDisplayUri()` be also added to `CompletionSuggestion`? Currently, we use `CompletionSuggestion.getElementUri()` to show in completion popup. See https://github.com/JetBrains/intellij-plugins/commit/41cb252cfd71703f8341a086e2ddef314dfd1163#diff-b788dc0bc2f2b9f218821dc4d4802bfaR368
Assuming that clients _do not have to_ subscribe to `CompletionService.AVAILABLE_SUGGESTION_SETS` having `CompletionSuggestion.getElementUri()` is probably the only way to get this functionality.

alexander-doroshko:
BTW it might be not too late to remove `CompletionSuggestion.getElementUri()` from the protocol completely if it has no use. 

You might also want to remove `CompletionSuggestion.getImportUri()`, I'm not sure it has any value. Or at least mark it as deprecated.

scheglov:
I removed `CompletionSuggestion.getImportUri` from the protocol and DAS in https://github.com/dart-lang/sdk/commit/5e2529c82e11150dbc2448f0f43aa4fcea16177b.

We lean toward removing `elementUri` completely. As I understand, it was added to solve the same problem that `CompletionService.AVAILABLE_SUGGESTION_SETS` is solving, and it was added recently. So, probably it is OK to get provide this data only when the new approach is used.

alexander-doroshko:
Cool!
What do you think about `getDisplayUri()`? Should only `IncludedSuggestionSet` have it or `CompletionSuggestion` as well?

scheglov:
Only `IncludedSuggestionSet` should have it.

I did not separate mentally `elementUri` and `displayUri` in `CompletionSuggestion`.

alexander-doroshko:
Retrospective:
- We added `CompletionSuggestion.elementUri` to the protocol to have URIs in parentheses in completion list
- We decided that `CompletionSuggestion.elementUri` sometimes look not so good in completion list (too long `file://...` URIs)
- To fix it we added `IncludedSuggestionSet.elementUri`. It works only if `CompletionService.AVAILABLE_SUGGESTION_SETS` subscription is enabled for those items that come via `IncludedSuggestionSet`.

So that's why 2 questions came to my mind:
- Remove `CompletionSuggestion.elementUri`?
- Add `CompletionSuggestion.displayUri` for those completion items that come via `CompletionSuggestion`, not via `IncludedSuggestionSet`?

alexander-doroshko:
Retrospective:
- We added `CompletionSuggestion.elementUri` to the protocol to have URIs in parentheses in completion list
- We decided that `CompletionSuggestion.elementUri` sometimes look not so good in completion list (too long `file://...` URIs)
- To fix it we added `IncludedSuggestionSet.displayUri`. It works only if `CompletionService.AVAILABLE_SUGGESTION_SETS` subscription is enabled for those items that come via `IncludedSuggestionSet`.

So that's why 2 questions came to my mind:
- Remove `CompletionSuggestion.elementUri`?
- Add `CompletionSuggestion.displayUri` for those completion items that come via `CompletionSuggestion`, not via `IncludedSuggestionSet`?

scheglov:
If `CompletionService.AVAILABLE_SUGGESTION_SETS` subscription is enabled, we get suggestions for all top-level declarations, excluding only the library where completion is requested (where we might consider to not display any URI as a signal of this to the user). So, for all interesting cases `IncludedSuggestionSet.displayUri` or `AvailableSuggestionSet.uri` is enough. There are also non-declarations (keyword, code blocks) or non-top-level completions (`myObject.^`), but for these URI is not interesting.

If `CompletionService.AVAILABLE_SUGGESTION_SETS` subscription is *not* enabled, we have the state we were about a month ago, not a huge loss.

alexander-doroshko:
Ok, thanks!

scheglov:
And... it gone.
https://github.com/dart-lang/sdk/commit/266ce73e289bf5a07af2b95b96a9d88f9e57815f

PR #636, opened 2019-02-11
call toString() via invoke()

devoncarew:
- call `toString()` via `invoke()`

After the Dart 2 CFE refactor, `invoke()` is much faster than `evaluate` (and in fact, `invoke` was added after the refactor in order to make these simple cases of evaluation faster).

The final version of `invoke` was only added to the sdk ~5 months ago (https://github.com/dart-lang/sdk/commit/7f9a9feae181573cca609fd1b8760d7992176699), so we may not be able to land thie PR right away.

@alexander-doroshko


alexander-doroshko:
Maybe check SDK version to choose how to call `toString()`?

devoncarew:
That would work; I'm probably won't have cycles for this PR for a while, but can circle back when I do.

devoncarew:
We should do an SDK version check in this PR.

devoncarew:
Updated!

- the `VmService` class can now report the version of the protocol supported by the current runtime (`runtimeVersion`)
- we use that info in `callToString` to either delegate to 'invoke' or 'eval' depending on whether the protocol version is pre or post 3.11
- tested manually against both a current runtime, and verified that the older code path does also return correct results

I'll need to upstream the change to the `VmServiceBase` class.

alexander-doroshko:
Thanks, merged.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -320,30 +320,32 @@ private boolean computeListPresentation(@NotNull final XValueNode node) {
...
+        if (toStringInstanceRef.getKind() == InstanceKind.String) {
+          String string = toStringInstanceRef.getValueAsString();
+          // default toString() implementation returns "Instance of 'ClassName'" - no interest to show
+          if (string.equals("Instance of '" + myInstanceRef.getClassRef().getName() + "'")) {
+            noGoodResult();
           }
           else {
-            noGoodResult(); // unlikely possible
+            if (toStringInstanceRef.getValueAsStringIsTruncated()) {
+              string += "...";
```
alexander-doroshko: I've changed this line to 
`addFullStringValueEvaluator(node, toStringInstanceRef);`
just like in other similar places. It adds `...View` link to see full value in a popup.

PR #637, opened 2019-02-14
Issue 35680 - Dart code completion suggestions from not yet imported libs

lambdabaa:
This change is associated with the analysis server protocol changes from
https://github.com/dart-lang/sdk/commit/d4b8e0696df14a7834b2f89659b7e3fc41fd3c91
which @scheglov has added support for on the server side. At a high
level, the flow has client send server a subscription for the available
suggestion sets. Server will then send availableSuggestions notifications
that give the client pending completions to cache. At completion time,
we include a couple new fields in CompletionSuggestion which allow us to
identify which of the cached suggestions should be included (in addition
to the default ones).

TODO: Either in this CL or a follow-up we need to update the client to respect relevance tags.

screenshot from 2019-02-14 15-43-01

Note that before this change, you would need to have imported `dart:io` and `dart:html` to get these suggestions.

/cc @alexander-doroshko @bwilkerson

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -2,7 +2,9 @@
 package com.jetbrains.lang.dart.analyzer;
 
 import com.google.common.collect.EvictingQueue;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.dart.server.*;
 import com.google.dart.server.generated.AnalysisServer;
@@ -186,6 +188,11 @@ public class DartAnalysisServerService implements Disposable {
     }
 
     @Override
+    public void computedAvailableSuggestions(@NotNull List<AvailableSuggestionSet> changed, @NotNull int[] removed) {
+      myServerData.computedAvailableSuggestions(changed, removed);
+    }
+
+    @Override
     public void computedErrors(@NotNull final String filePathSD, @NotNull final List<AnalysisError> errors) {
       final String fileName = PathUtil.getFileName(filePathSD);
 
@@ -275,9 +282,13 @@ public class DartAnalysisServerService implements Disposable {
                                    final int replacementOffset,
                                    final int replacementLength,
                                    @NotNull final List<CompletionSuggestion> completions,
+                                   @NotNull final List<IncludedSuggestionSet> includedSuggestionSets,
+                                   @NotNull final List<String> includedSuggestionKinds,
                                    final boolean isLast) {
       synchronized (myCompletionInfos) {
-        myCompletionInfos.add(new CompletionInfo(completionId, replacementOffset, replacementLength, completions, isLast));
+        myCompletionInfos.add(
+          new CompletionInfo(completionId, replacementOffset, replacementLength, completions, includedSuggestionSets,
+                             includedSuggestionKinds, isLast));
         myCompletionInfos.notifyAll();
       }
     }
@@ -441,7 +452,8 @@ public class DartAnalysisServerService implements Disposable {
 
   public void addCompletions(@NotNull final VirtualFile file,
                              @NotNull final String completionId,
-                             @NotNull final CompletionSuggestionConsumer consumer) {
+                             @NotNull final CompletionSuggestionConsumer consumer,
+                             @NotNull final CompletionLibraryRefConsumer libraryRefConsumer) {
     while (true) {
       ProgressManager.checkCanceled();
 
@@ -455,6 +467,11 @@ public class DartAnalysisServerService implements Disposable {
             final int convertedReplacementOffset = getConvertedOffset(file, completionInfo.myOriginalReplacementOffset);
             final int convertedReplacementLength = getConvertedOffset(file, completionInfo.myOriginalReplacementLength);
             consumer.consumeCompletionSuggestion(convertedReplacementOffset, convertedReplacementLength, completion);
+          }
+
+          final Set<String> includedKinds = Sets.newHashSet(completionInfo.myIncludedSuggestionKinds);
+          for (final IncludedSuggestionSet includedSet : completionInfo.myIncludedSuggestionSets) {
+            libraryRefConsumer.consumeLibraryRef(includedSet, includedKinds);
           }
           return;
         }
@@ -720,6 +737,11 @@ public class DartAnalysisServerService implements Disposable {
 
   private void handleClosingLabelPreferenceChanged() {
     analysis_setSubscriptions();
+  }
+
+  @Nullable
+  public AvailableSuggestionSet getAvailableSuggestionSet(int id) {
+    return myServerData.getAvailableSuggestionSet(id);
   }
 
   @NotNull
@@ -1323,6 +1345,37 @@ public class DartAnalysisServerService implements Disposable {
   }
 
   @Nullable
+  public GetCompletionDetailsResult completion_getSuggestionDetails(@NotNull final VirtualFile file,
+                                                                    final int id,
+                                                                    final String label,
+                                                                    final int offset) {
+    final String filePath = FileUtil.toSystemDependentName(file.getPath());
+    final Ref<GetCompletionDetailsResult> resultRef = new Ref<>();
+
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return null;
+    }
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    server.completion_getSuggestionDetails(filePath, id, label, offset, new GetSuggestionDetailsConsumer() {
+      @Override
+      public void computedDetails(GetCompletionDetailsResult result) {
+        resultRef.set(result);
+        latch.countDown();
+      }
+
+      @Override
+      public void onError(RequestError requestError) {
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, GET_SUGGESTIONS_TIMEOUT);
+    return resultRef.get();
+  }
+
+  @Nullable
   public String completion_getSuggestions(@NotNull final VirtualFile file, final int _offset) {
     final String filePath = FileUtil.toSystemDependentName(file.getPath());
     final Ref<String> resultRef = new Ref<>();
@@ -1813,11 +1866,14 @@ public class DartAnalysisServerService implements Disposable {
         if (Registry.is("dart.projects.without.pubspec", false) && isAnalyzedFilesSubscriptionEnabled()) {
           startedServer.analysis_setGeneralSubscriptions(Collections.singletonList(GeneralAnalysisService.ANALYZED_FILES));
         }
+        startedServer.completion_setSubscriptions(ImmutableList.of(CompletionService.AVAILABLE_SUGGESTION_SETS));
 
         if (!myInitializationOnServerStartupDone) {
           myInitializationOnServerStartupDone = true;
 
           registerFileEditorManagerListener();
+
+
           registerDocumentListener();
           setDasLogger();
         }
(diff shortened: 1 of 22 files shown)


lambdabaa:
@alexander-doroshko Note that you'll need the latest version of analysis server to test the new behavior :)

alexander-doroshko:
Thanks! I'll review this PR on Monday.

alexander-doroshko:
I played with it and it looks great!

I noticed that managing imports could probably be better if the import statement is already there but the completed item is not yet accessible because of `show` or `hide` modifier. But that's a server-side issue. I guess it's WIP.

A few comments inline.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1322,6 +1344,37 @@ public void onError(RequestError error) {
...
+        latch.countDown();
+      }
+
+      @Override
+      public void onError(RequestError requestError) {
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, GET_SUGGESTIONS_TIMEOUT);
```
alexander-doroshko: This code runs in EDT, it blocks UI. We can't wait for so long here. I'd say, no longer than 100ms.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -455,8 +479,46 @@ public static LookupElementBuilder createLookupElement(@NotNull final Project pr
...
+    if (isNotYetImported) {
+      lookup = lookup.withInsertHandler((context, item) -> {
+        final DartAnalysisServerService das = DartAnalysisServerService.getInstance(project);
+        final GetCompletionDetailsResult result = das.completion_getSuggestionDetails(file, suggestionSetId, suggestion.getCompletion(),
+                                                                                      context.getStartOffset());
+        if (result == null) {
+          return;
+        }
+
+        context.getDocument().replaceString(context.getStartOffset(), context.getTailOffset(), result.completion);
```
alexander-doroshko: Why do we need this line? Isn't it effectively a noop?

alexander-doroshko:
> TODO: Either in this CL or a follow-up we need to update the client to respect relevance tags.

You already use it, don't you? Or what do you mean?

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1322,6 +1344,37 @@ public void onError(RequestError error) {
...
+    final String filePath = FileUtil.toSystemDependentName(file.getPath());
+    final Ref<GetCompletionDetailsResult> resultRef = new Ref<>();
+
+    final AnalysisServer server = myServer;
+    if (server == null) {
+      return null;
+    }
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    server.completion_getSuggestionDetails(filePath, id, label, offset, new GetSuggestionDetailsConsumer() {
```
alexander-doroshko: I guess we need to use `getOriginalOffset()`. Offsets in the IntelliJ's model are normalized and may be different from the ones in the Analysis Serve's model.

alexander-doroshko:
P.S. It would be great to add one or more (if needed) tests to `DartServerCompletionTest`.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
lambdabaa: Makes sense, done!

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
lambdabaa: Good catch!

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: > > TODO: Either in this CL or a follow-up we need to update the client to respect relevance tags.
> 
> You already use it, don't you? Or what do you mean?

That's client support for https://github.com/dart-lang/sdk/commit/b9bcfb4529905a2533200ad11547db67740fdc6c.

lambdabaa:
> > TODO: Either in this CL or a follow-up we need to update the client to respect relevance tags.
> 
> You already use it, don't you? Or what do you mean?

That's client support for https://github.com/dart-lang/sdk/commit/b9bcfb4529905a2533200ad11547db67740fdc6c (just a nice-to-have).

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Here the client asks server for the full replacement since an import may be needed in addition to the cursor-position completion insertion. This completion might not be the same as what we put in the LookupElement. Does that make sense?

lambdabaa:
> I noticed that managing imports could probably be better if the import statement is already there but the completed item is not yet accessible because of `show` or `hide` modifier. But that's a server-side issue. I guess it's WIP.

We did add server support for same-named symbols from different libraries using show, but there are probably cases that it doesn't cover. I will look into that separately with @scheglov, thanks!

lambdabaa:
> P.S. It would be great to add one or more (if needed) tests to `DartServerCompletionTest`.

ooo does that use a real analysis server or is it mocked out?

alexander-doroshko:
> ooo does that use a real analysis server or is it mocked out?

Real. Will the test pass in your environment?

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
alexander-doroshko: > This completion might not be the same as what we put in the LookupElement. Does that make sense?

If so then yes. Great scenario for a test!

scheglov:
> I played with it and it looks great!

  Thank you!
  I have been developing and fish-fooding it myself for a couple of weeks and found it convenient :-)

> I noticed that managing imports could probably be better if the import statement is already there but the completed item is not yet accessible because of `show` or `hide` modifier. But that's a server-side issue. I guess it's WIP.

  Yes, DAS can return any change, so we can enhance it later.

  Yes, it might be worth to update `show`.

  But `hide` is probably an explicit signal, that user does not (or did not) want to use this symbol because of a pre-existing conflicts. OTOH, if there isn't, we might do it anyway.

  We initially though that these enhancements might be done if users actually request them. But maybe I will find time to do it sooner.

lambdabaa:
> Real. Will the existing tests pass in your environment?

15/20 of them pass. There are three that are failing at

```
junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:55)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.Assert.assertNotNull(Assert.java:256)
	at junit.framework.Assert.assertNotNull(Assert.java:248)
	at junit.framework.TestCase.assertNotNull(TestCase.java:417)
	at com.jetbrains.dart.analysisServer.DartServerCompletionTest.selectLookup(DartServerCompletionTest.java:55)
```

selectLookup("identical", ...) appears always to be `null`.  The other two look to be legitimate tests we should update or disable until analysis server does better with the `show` behavior.

screen shot 2019-02-18 at 12 08 22 pm
screen shot 2019-02-18 at 12 08 50 pm

alexander-doroshko:
> Yes, DAS can return any change, so we can enhance it later.

I'm not a real user so I might be wrong with real user expectations. I just noticed that DAS adds one more import statement (without `show`/`hide`) while the existing one (with `show` or `hide`) also remains.

alexander-doroshko:
> 15/20 of them pass

Hm, I have all 20 passing against the latest nightly SDK build. Well, anyway, it's enough for you to add more tests! :)

lambdabaa:
> Hm, I have all 20 passing against the latest nightly SDK build. Well, anyway, it's enough for you to add more tests! :)

Yes, true :). I added a test that exercises the basic interaction. Let me know if you have any questions and thanks for your review help + suggestions.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
alexander-doroshko: I've updated the test a bit so that it shows `Process` but inserts `prefix0.Process`. Without this line it fails, so this line is now covered with tests.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -471,6 +533,39 @@ public static LookupElementBuilder createLookupElement(@NotNull final Project pr
...
 
+
+  @NotNull
+  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(@NotNull final AvailableSuggestion suggestion,
+                                                                                        @NotNull final String kind,
+                                                                                        @NotNull final AvailableSuggestionSet suggestionSet,
+                                                                                        final int relevance) {
+    Element element = suggestion.getElement();
+    return new CompletionSuggestion(
+      kind,
```
alexander-doroshko: I removed `kind` argument and passed `"UNKNOWN"` instead. We don't have `CompletionSuggestionKind`, we only have `ElementKind`.

Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerCompletionTest.java
```diff
@@ -176,4 +180,18 @@ public void testNoCompletionAfterDigit() {
     myFixture.completeBasic();
     assertNull(myFixture.getLookup());
   }
+
+  public void testNotYetImportedClass() throws InterruptedException {
+    // Wait until we've received symbols to cache from the dart:io library from analysis server.
+    while (true) {
```
alexander-doroshko: I removed `while(true)` and added `doHighlighting()` instead. While highlighting goes, available suggestions arrive. 
I removed unused `DartAnalysisServerService..getAvailableSuggestionSets()` and `DartServerData.getAvailableSuggestionSets()`.

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
```diff
@@ -31,6 +31,7 @@
 
   private final EventDispatcher<OutlineListener> myEventDispatcher = EventDispatcher.create(OutlineListener.class);
 
+  private final Map<Integer, AvailableSuggestionSet> myAvailableSuggestionSetMap = Collections.synchronizedMap(new THashMap<>());
```
alexander-doroshko: I also made sure to clear the map on server stop.

Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerCompletionTest.java
lambdabaa: Without these bits, how can the test know that the symbols from `dart:io` have been cached? If they haven't been sent in an available suggestions notification when we make the completions request, it won't show up in the completion suggestion response.

alexander-doroshko: Well, the test doesn't know it for sure. It's possible that the test will fail because of a race. For now, it waits for `DartAnalysisServerService.waitForAnalysisToComplete_TESTS_ONLY()` to return and this time is enough for getting completion data from the server. At least in my environment :). If it fails too often we'll think about a more reliable solution.

alexander-doroshko:
Pushed manually, thanks for the PR!

alexander-doroshko:
I noticed that completion of constructors disappeared. Just wanted to make sure this is expected.
Before:
image
After:
image
// cc @scheglov

scheglov:
This is a known limitation.
I have not decided whether we want constructors.

scheglov:
Brian and I discussed this, and decided that we want to add constructors.

PR #638, opened 2019-02-25
Dart plugin changes to support CompletionSuggestion.includedRelevanceTags

lambdabaa:
This is a follow up from https://github.com/JetBrains/intellij-plugins/commit/96fe224d4578583671570d65c4aae7fca0d0ba15. In this patch, I'm updating the client to read the includedSuggestionRelevanceTags and apply them on top of the base relevances of available suggestion sets. The corresponding protocol patch from @scheglov was https://github.com/dart-lang/sdk/commit/67c5923ccedf6e04af617f47ee734f574f13ad87#diff-441befac1720a80d45409f9dfe72ec80.

/cc @alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -53,6 +53,7 @@ import org.jetbrains.annotations.Nullable;
 import javax.swing.*;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import static com.intellij.patterns.PlatformPatterns.psiElement;
 import static com.intellij.patterns.PlatformPatterns.psiFile;
@@ -140,7 +141,7 @@ public class DartServerCompletionContributor extends CompletionContributor {
                  }
 
                  updatedResultSet.addElement(lookupElement);
-               }, (includedSet, includedKinds) -> {
+               }, (includedSet, includedKinds, includedRelevanceTags) -> {
                  final AvailableSuggestionSet suggestionSet = das.getAvailableSuggestionSet(includedSet.getId());
                  if (suggestionSet == null) {
                    return;
@@ -152,10 +153,10 @@ public class DartServerCompletionContributor extends CompletionContributor {
                      continue;
                    }
 
-                   CompletionSuggestion completionSuggestion =
-                     createCompletionSuggestionFromAvailableSuggestion(suggestion, suggestionSet, includedSet.getRelevance());
-                   LookupElementBuilder lookupElement =
-                     createLookupElement(project, completionSuggestion, suggestionSet.getId(), targetFile, true);
+                   CompletionSuggestion completionSuggestion = createCompletionSuggestionFromAvailableSuggestion(
+                     suggestion, suggestionSet, includedSet.getRelevance(), includedRelevanceTags);
+                   LookupElementBuilder lookupElement = createLookupElement(
+                     project, completionSuggestion, suggestionSet.getId(), targetFile, true, suggestionSet.getUri());
                    resultSet.addElement(lookupElement);
                  }
                });
@@ -338,12 +339,13 @@ public class DartServerCompletionContributor extends CompletionContributor {
 
   @NotNull
   public static LookupElementBuilder createLookupElement(@NotNull final Project project, @NotNull final CompletionSuggestion suggestion) {
-    return createLookupElement(project, suggestion, null, null, false);
+    return createLookupElement(project, suggestion, null, null, false, null);
   }
 
   @NotNull
   public static LookupElementBuilder createLookupElement(@NotNull final Project project, @NotNull final CompletionSuggestion suggestion,
-                                                         final Integer suggestionSetId, final VirtualFile file, boolean isNotYetImported) {
+                                                         final Integer suggestionSetId, final VirtualFile file, boolean isNotYetImported,
+                                                         String displayUri) {
     final Element element = suggestion.getElement();
     final Location location = element == null ? null : element.getLocation();
     final DartLookupObject lookupObject = new DartLookupObject(project, location, suggestion.getRelevance());
@@ -395,8 +397,8 @@ public class DartServerCompletionContributor extends CompletionContributor {
       }
 
       // If this is a class, try to show which package it's coming from.
-      if (element.getKind().equals(ElementKind.CLASS) && suggestion.getElementUri() != null) {
-        lookup = lookup.appendTailText(" (" + suggestion.getElementUri() + ")", true);
+      if (element.getKind().equals(ElementKind.CLASS) && displayUri != null) {
+        lookup = lookup.appendTailText(" (" + displayUri + ")", true);
       }
 
       // icon
@@ -541,16 +543,26 @@ public class DartServerCompletionContributor extends CompletionContributor {
 
 
   @NotNull
-  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(@NotNull final AvailableSuggestion suggestion,
-                                                                                        @NotNull final AvailableSuggestionSet suggestionSet,
-                                                                                        final int relevance) {
+  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(
+      @NotNull final AvailableSuggestion suggestion,
+      @NotNull final AvailableSuggestionSet suggestionSet,
+      final int suggestionSetRelevance,
+      Map<String, IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags) {
+    int relevance = suggestionSetRelevance;
+    for (String tag : suggestion.getRelevanceTags()) {
+      if (!includedSuggestionRelevanceTags.containsKey(tag)) {
+        continue;
+      }
+
+      relevance += includedSuggestionRelevanceTags.get(tag).getRelevanceBoost();
+    }
     Element element = suggestion.getElement();
+    List<String> suggestionTags = suggestion.getRelevanceTags();
     return new CompletionSuggestion(
       "UNKNOWN", // TODO should be CompletionSuggestionKind
       relevance,
       suggestion.getLabel(),
       null,
-      suggestionSet.getUri(),
       0,
       0,
       element.isDeprecated(),
@@ -567,8 +579,7 @@ public class DartServerCompletionContributor extends CompletionContributor {
       null,
       null,
       null,
-      null,
-      suggestionSet.getUri());
+      null);
   }
 
   private static Icon getBaseImage(Element element) {
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -397,7 +397,9 @@ public class DartServerCompletionContributor extends CompletionContributor {
       }
 
       // If this is a class, try to show which package it's coming from.
-      if (element.getKind().equals(ElementKind.CLASS) && displayUri != null) {
+      if (!element.getKind().equals(ElementKind.FUNCTION) &&
+          !element.getKind().equals(ElementKind.FUNCTION_INVOCATION) &&
+          displayUri != null) {
         lookup = lookup.appendTailText(" (" + displayUri + ")", true);
       }
 
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -285,10 +285,11 @@ public class DartAnalysisServerService implements Disposable {
                                    @NotNull final List<CompletionSuggestion> completions,
                                    @NotNull final List<IncludedSuggestionSet> includedSuggestionSets,
                                    @NotNull final List<String> includedSuggestionKinds,
+                                   @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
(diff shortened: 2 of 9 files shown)


Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -395,8 +397,8 @@ public static LookupElementBuilder createLookupElement(@NotNull final Project pr
       }
 
       // If this is a class, try to show which package it's coming from.
-      if (element.getKind().equals(ElementKind.CLASS) && suggestion.getElementUri() != null) {
-        lookup = lookup.appendTailText(" (" + suggestion.getElementUri() + ")", true);
+      if (element.getKind().equals(ElementKind.CLASS) && displayUri != null) {
```
scheglov: BTW, this means that when the user completes not a class, e.g. a typedef, or a function, we don't show the URI. It seems sad.

alexander-doroshko:
I'm sick at home today. Will take a look when I get better.

lambdabaa:
> I'm sick at home today. Will take a look when I get better.

Feel better Alex!

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Yes, good point. Looking through https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_ElementKind do you think there are other ElementKinds that we could include here that would make sense without changing the UI? We discussed offline, but there's an issue in the case of a function that the parameters (already in parentheses) could conflict with the package info.

scheglov: For everything that is not a function, we can show the URI.

And for functions probably display parameters.

lambdabaa: Sounds good, done!

lambdabaa:
@alexander-doroshko Gentle nudge on this :)

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
alexander-doroshko: @scheglov, @lambdabaa, should we also skip showing URI for types like `ElementKind.SETTER`, `ElementKind.METHOD`, `ElementKind.CONSTRUCTOR`, `ElementKind.CONSTRUCTOR_INVOCATION`, or any other `ElementKind`?

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -541,16 +545,26 @@ else if (shouldSetSelection) {
 
 
   @NotNull
-  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(@NotNull final AvailableSuggestion suggestion,
-                                                                                        @NotNull final AvailableSuggestionSet suggestionSet,
-                                                                                        final int relevance) {
+  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(
+      @NotNull final AvailableSuggestion suggestion,
+      @NotNull final AvailableSuggestionSet suggestionSet,
```
alexander-doroshko: IDE says that this argument became unused. Removed.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -541,16 +545,26 @@ else if (shouldSetSelection) {
...
+    int relevance = suggestionSetRelevance;
+    for (String tag : suggestion.getRelevanceTags()) {
+      if (!includedSuggestionRelevanceTags.containsKey(tag)) {
+        continue;
+      }
+
+      relevance += includedSuggestionRelevanceTags.get(tag).getRelevanceBoost();
+    }
     Element element = suggestion.getElement();
+    List<String> suggestionTags = suggestion.getRelevanceTags();
```
alexander-doroshko: Unused local var. Removed.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Good point. Maybe the right thing to do then is

```
if (displayUri != null &&
    element.getTypeParameters() == null &&
    element.getParameters() == null) {
  lookup = lookup.appendTailText(" (" + displayUri + ")", true);
}
```

since that's exactly when there won't be conflicts. Wdyt?

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -541,16 +545,26 @@ else if (shouldSetSelection) {
...
-  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(@NotNull final AvailableSuggestion suggestion,
-                                                                                        @NotNull final AvailableSuggestionSet suggestionSet,
-                                                                                        final int relevance) {
+  private static CompletionSuggestion createCompletionSuggestionFromAvailableSuggestion(
+      @NotNull final AvailableSuggestion suggestion,
+      @NotNull final AvailableSuggestionSet suggestionSet,
+      final int suggestionSetRelevance,
+      Map<String, IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags) {
+    int relevance = suggestionSetRelevance;
+    for (String tag : suggestion.getRelevanceTags()) {
```
alexander-doroshko: I got NPEs here both when running DartServerCompletionTest or when testing manually. Probably 'relevaneTags' field is optional in the protocol, so it appears to be null.
I surrounded this `for` loop with `if (relevanceTags != null) {...}`.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
alexander-doroshko: Sounds reasonable. Done as you suggest. Happy to hear @scheglov's comments.

alexander-doroshko:
Pushed, thanks!

Some follow-up comments:

- Just noticed that [Server API page](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_IncludedSuggestionRelevanceTag) doesn't contain some types descriptions like `AvailableSuggestion`,  `AvailableSuggestionSet`, or `IncludedSuggectionRelevanceTag`. But it mentions `AvailableSuggestionRelevanceTag` that doesn't exist AFAIU. Is this expected for some reason?

- Do you know how to write a new test in `DartServerCompletionTest` that would check relevanceTag-driven sorting? I guess it will be like existing `testSorting()` test, but I don't know how to write appropriate test file.

- I have only 2 tests from `DartServerCompletionTest` that fail locally against the latest nightly SDK. Both because of arguments placeholders. With this PR we now have `AvailableSuggestion.getDefaultArgumentListString()` and `AvailableSuggestion.getDefaultArgumentListTextRanges()`, so looks like we can fix the problem! If you have the fix in flight I'm ready to review. Otherwize I'm happy to fix it myself.

Cheers!

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
scheglov: Thank you, I think this is as good as we can do keeping the same approach. So, we can see that `json` top-level variant is from `dart:convert`. We cannot see that `jsonDecode` is also from `dart:convert`, but there is no space for it :-(

image

lambdabaa:
> * Just noticed that [Server API page](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_IncludedSuggestionRelevanceTag) doesn't contain some types descriptions like `AvailableSuggestion`,  `AvailableSuggestionSet`, or `IncludedSuggectionRelevanceTag`. But it mentions `AvailableSuggestionRelevanceTag` that doesn't exist AFAIU. Is this expected for some reason?
> *

It's because they were marked as experimental, but we should go ahead and set them to non-experimental now that we're shipping support in IntelliJ /cc @scheglov @bwilkerson



> * Do you know how to write a new test in `DartServerCompletionTest` that would check relevanceTag-driven sorting? I guess it will be like existing `testSorting()` test, but I don't know how to write an appropriate test data file.
> *

Good question, I will look into this.

> * I have only 2 tests from `DartServerCompletionTest` that fail locally against the latest nightly SDK. Both because of arguments placeholders. With this PR we now have `AvailableSuggestion.getDefaultArgumentListString()` and `AvailableSuggestion.getDefaultArgumentListTextRanges()`, so looks like we can fix the problem! If you have the fix in flight I'm ready to review. Otherwize I'm happy to fix it myself.

I can take a look at this as well and thanks for pointing it out :)

scheglov:
https://dart-review.googlesource.com/c/sdk/+/95661 will graduate these APIs from experiment.

lambdabaa:
> * Just noticed that [Server API page](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_IncludedSuggestionRelevanceTag) doesn't contain some types descriptions like `AvailableSuggestion`,  `AvailableSuggestionSet`, or `IncludedSuggectionRelevanceTag`. But it mentions `AvailableSuggestionRelevanceTag` that doesn't exist AFAIU. Is this expected for some reason?
> * Do you know how to write a new test in `DartServerCompletionTest` that would check relevanceTag-driven sorting? I guess it will be like existing `testSorting()` test, but I don't know how to write an appropriate test data file.
> * I have only 2 tests from `DartServerCompletionTest` that fail locally against the latest nightly SDK. Both because of arguments placeholders. With this PR we now have `AvailableSuggestion.getDefaultArgumentListString()` and `AvailableSuggestion.getDefaultArgumentListTextRanges()`, so looks like we can fix the problem! If you have the fix in flight I'm ready to review. Otherwize I'm happy to fix it myself.

After passing along these new fields from AvailableSuggestion I'm still seeing empty parameterNames which is causing the tests to fail. Should they be populated? /cc @scheglov

scheglov:
I'm not sure what could be a problem, but it might be a client issue.
As far I can see, we provide necessary data in DAS notification.

For this code:

```dart
void bar2(int a, double b) {}
```

DAS sends:

```json
{
  "event": "completion.availableSuggestions",
  "params": {
    "changedLibraries": [
      {
        "id": 1120,
        "uri": "file:///Users/scheglov/dart/test/bin/test2.dart",
        "items": [
          {
            "label": "bar2",
            "element": {
              "kind": "FUNCTION",
              "name": "bar2",
              "location": {
                "file": "/Users/scheglov/dart/test/bin/test2.dart",
                "offset": 5,
                "length": 0,
                "startLine": 1,
                "startColumn": 6
              },
              "flags": 0,
              "parameters": "(int a, double b)",
              "returnType": "void"
            },
            "defaultArgumentListString": "a, b",
            "defaultArgumentListTextRanges": [
              0,
              1,
              3,
              1
            ],
            "parameterNames": [
              "a",
              "b"
            ],
            "parameterTypes": [
              "int",
              "double"
            ],
            "requiredParameterCount": 2
          }
        ]
      }
    ],
    "removedLibraries": []
  }
}
```

alexander-doroshko:
> I have only 2 tests from DartServerCompletionTest that fail locally against the latest nightly SDK. Both because of arguments placeholders. With this PR we now have AvailableSuggestion.getDefaultArgumentListString() and AvailableSuggestion.getDefaultArgumentListTextRanges(), so looks like we can fix the problem! If you have the fix in flight I'm ready to review. Otherwize I'm happy to fix it myself.

I've just pushed the fix: https://github.com/JetBrains/intellij-plugins/commit/fd2bfd2264654d0aa94dab0b7decca2170a2b7d5. All completion tests now pass for me. Can you please review? I have just a few hours to cherry-pick it.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -541,16 +545,26 @@ else if (shouldSetSelection) {
...
+      if (!includedSuggestionRelevanceTags.containsKey(tag)) {
+        continue;
+      }
+
+      relevance += includedSuggestionRelevanceTags.get(tag).getRelevanceBoost();
```
DanTup: FYI - I was investigating an issue with relevance/sorting in VS Code (https://github.com/Dart-Code/Dart-Code/issues/1742) and noticed this in the comments for tags:

```
         * If an AvailableSuggestion has relevance tags that match more
	 * than one IncludedSuggestionRelevanceTag, the maximum
	 * relevance boost is used.
	 */
```

My code is currently adding them together (which I think I borrowed from here :-)) but it seems like that might be incorrect. I'm not sure who is best to confirm (@scheglov?).

scheglov: I don't remember if we made a decision to do otherwise, so probably we should follow the documentation recommendation.

lambdabaa: Thanks for reminding me about this @DanTup! We went back and forth on this a bit but I think we should use the maximum. I will go ahead and make a CL for the IntelliJ plugin to reflect that.

PR #639, opened 2019-02-25
Dart client changes to support completion.registerLibraryPaths

lambdabaa:
This change follows the Dart analysis server protocol changes in https://github.com/dart-lang/sdk/commit/d4b8e0696df14a7834b2f89659b7e3fc41fd3c91 and will allow other IntelliJ plugins (ie Flutter) to configure the libraries from which analysis server should provide code completions.

/cc @alexander-doroshko @scheglov @bwilkerson

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/utilities/RequestUtilities.java
+++ b/Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/utilities/RequestUtilities.java
@@ -18,6 +18,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import com.intellij.icons.AllIcons;
 import org.dartlang.analysis.server.protocol.*;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


alexander-doroshko:
Merged. Curious, what particular use case do you have in mind?

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/utilities/RequestUtilities.java
```diff
@@ -18,6 +18,7 @@
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import com.intellij.icons.AllIcons;
```
alexander-doroshko: Looks like a typo? I removed this line.

lambdabaa:
Thanks @alexander-doroshko! The idea is that depending on which part of the codebase a user is editing (widgets, tests, etc) different sets of completion suggestions may be more or less appropriate.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/utilities/RequestUtilities.java
lambdabaa: Whoops, thanks!

alexander-doroshko:
Why should this be a client-side logic?
Well, even if it should, nothing prevents client from filtering out less appropriate suggestions without introducing new server API. Am I wrong?

lambdabaa:
@alexander-doroshko It's a good question. What it comes down to is that we don't want server to have to index and send client cached/precomputed suggestions for all libraries, just the ones that client is specifically interested in.

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/a12949481d99a3b645979d44035039ac8c315cf5

PR #640, opened 2019-03-01
s/includedSuggestionKinds/includedElementKinds/

lambdabaa:
This change renames includedSuggestionKinds to includedElementKinds in the Dart plugin following a corresponding protocol and server change.


scheglov:
LGTM

PR #642, opened 2019-03-07
Use `toStringDeep()` for better display of variable in Dart debugger

jingbian:
We've added (and will add more) usages of a method called `toStringDeep()` to some classes, which expose more debugging information so that the IDE (like IntelliJ) can use it to gather more information and have a better display of the variables in debugger. It uses a new method instead of just `toString()` because we don't want to change the original behavior of all those objects. Also because sometimes the debug information is quite long, not suitable for logging etc.

Screenshot [here](https://photos.app.goo.gl/V53LkmLbzSeorapf6).

This can already be used in the [webdriver.dart](https://github.com/google/webdriver.dart) project after [this commit](https://github.com/google/webdriver.dart/commit/9a187cb44a288d7d0d4c64103dcbfc3e395440ed) has been submitted. Just debug the [chrome_json_wire_web_element_test](https://github.com/google/webdriver.dart/blob/master/test/chrome_json_wire_web_element_test.dart) and see.

Safety for failing case: The new method won't be touched if previous calculation of presentation works. And the new method will go back to the original computeDefaultPresentation method if it doesn't work out. Worst case, those variables without an already known presentation may be impacted. So in general it should be quite safe.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
@@ -164,12 +164,14 @@ public class DartVmServiceValue extends XNamedValue {
         public void received(final InstanceRef toStringInstanceRef) {
           if (toStringInstanceRef.getKind() == InstanceKind.String) {
             String content = toStringInstanceRef.getValueAsString();
-            String summary = content.split("\n")[0];
+            int firstLineBreak = content.indexOf('\n');
+            String summary = firstLineBreak < 0 ? content : content.substring(0, firstLineBreak);
             node.setPresentation(getIcon(), myInstanceRef.getClassRef().getName(), summary, true);
             if (toStringInstanceRef.getValueAsStringIsTruncated()) {
               addFullStringValueEvaluator(node, toStringInstanceRef);
             }
-            else {
+            else if (firstLineBreak >= 0){
+              // Multi-line content. Display a View link to reveal full content.
               node.setFullValueEvaluator(new ImmediateFullValueEvaluator("View", content));
             }
           }
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
@@ -172,7 +172,7 @@ public class DartVmServiceValue extends XNamedValue {
             }
             else if (firstLineBreak >= 0){
               // Multi-line content. Display a View link to reveal full content.
-              node.setFullValueEvaluator(new ImmediateFullValueEvaluator("View", content));
+              node.setFullValueEvaluator(new ImmediateFullValueEvaluator("...View", content));
             }
           }
           else {
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
@@ -149,8 +149,40 @@ public class DartVmServiceValue extends XNamedValue {
     if (computeRegExpPresentation(node)) return;
     if (computeMapPresentation(node)) return;
     if (computeListPresentation(node)) return;
-    computeDefaultPresentation(node);
+
+    // computeDefaultPresentation is called internally here when no result is got.
+    // The reason for this is that the async method used cannot be properly waited.
+    computeDebugPresentation(node);
+
     // todo handle other special kinds: Type, TypeParameter, Pattern, may be some others as well
+  }
+
+  private void computeDebugPresentation(final XValueNode node) {
+    myDebugProcess.getVmServiceWrapper()
+      .evaluateInTargetContext(myIsolateId, myInstanceRef.getId(), "toStringDeep()", new VmServiceConsumers.EvaluateConsumerWrapper() {
+        @Override
+        public void received(final InstanceRef toStringInstanceRef) {
+          if (toStringInstanceRef.getKind() == InstanceKind.String) {
+            String content = toStringInstanceRef.getValueAsString();
+            String summary = content.split("\n")[0];
+            node.setPresentation(getIcon(), myInstanceRef.getClassRef().getName(), summary, true);
+            if (toStringInstanceRef.getValueAsStringIsTruncated()) {
+              addFullStringValueEvaluator(node, toStringInstanceRef);
+            }
+            else {
+              node.setFullValueEvaluator(new ImmediateFullValueEvaluator("View", content));
+            }
+          }
+          else {
+            noGoodResult();
+          }
+        }
+
+        @Override
+        public void noGoodResult() {
+          computeDefaultPresentation(node);
+        }
+      });
   }
 
   private Icon getIcon() {


jwren:
Jing uploaded this a few hours ago and asked me to ping reviewers: @devoncarew @alexander-doroshko @lambdabaa

devoncarew:
I wonder if tostringdeep is the most descriptive name here; are there any other similar patterns we could crib off of?

jingbian:
Jacob Richman (@jacob314) mentioned to me this may be the way Flutter exposes some information. Hopefully I didn't get him wrong.
So I chose this name so that it may also work for Flutter.

alexander-doroshko:
Is `toStringDeep()` going to become a standard practice, documented and recommended for a wide audience? I mean, I don't mind in general. I'd like only to make sure that this feature is not for a single project or for a small group of developers.

P.S. I have a day off tomorrow and will get back to work on Monday.

jingbian:
We are at least adding this to pageloader and webdriver packages, which are the most hard to understand objects with current debugger (with only type name and a random WebDriver id). Hopefully also Flutter.

With this code added, it can become a general standard if anyone wants his object to be differently displayed, he can just add that.

devoncarew:
I wonder if it would help discoverability of this functionality if there was a `Describable` interface in some common package, that just defined the `String toStringDeep()` method? The debugger would still just look for a method with the correct name, it wouldn't care about things implementing the interface per se.

jingbian:
Agreed, though that can be a different topic.

I once thought about this, but moved away. Because unless it's in Dart SDK, there is no reason for other packages to reference this just for the method.

But if this works well for the planned packages, maybe we can think about this (add it to the Dart SDK?).

alexander-doroshko:
> pageloader and webdriver packages, which are the most hard to understand objects

Maybe just override standard `toString()` in such cases?

jacob314:
Moving `Diagnosticable` from `package:flutter` to a common package would be a good thing.

It would be ideal to avoid adding a new dependency to `package:flutter` so I would suggest adding it to `package:collection` which could plausibly be tweaked to use the class to add better debugging views of its collections.

jingbian:
> > pageloader and webdriver packages, which are the most hard to understand objects
> 
> Maybe just override standard `toString()` in such cases?

So there are some downsides compared to this PR.

1. It interferes with a common method, which some code may depend on. It's part of an object behavior.
2. `toString()` is commonly used in logging. It may be a problem if it's too long, or formatted in a custom way (like multi paragraph).
3. Instead of the original `toString()` way, we can design how `toDebugString()`. Like first line should be displayed directly while all the rest should only be displayed in a popup.

jingbian:
> Moving `Diagnosticable` from `package:flutter` to a common package would be a good thing.
> 
> It would be ideal to avoid adding a new dependency to `package:flutter` so I would suggest adding it to `package:collection` which could plausibly be tweaked to use the class to add better debugging views of its collections.

Agreed. Even though some implementers may not need to depend on that package when they only need simple information. Having a well implemented way to diagnose objects is a big addition!

alexander-doroshko:
> Instead of the original toString() way, we can design how toDebugString().

`toDebugString()`? Is it an alternative name for `toStringDeep()`? Will we have to support both?

jingbian:
> > Instead of the original toString() way, we can design how toDebugString().
> 
> `toDebugString()`? Is it an alternative name for `toStringDeep()`? Will we have to support both?

Sorry... That's a typo.

jingbian:
Since we've updated the pageloader package. Here is a new screenshot when you debug it: https://photos.app.goo.gl/zPtFokP2KseT4FdD9

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -149,10 +149,42 @@ public void computePresentation(@NotNull final XValueNode node, @NotNull final X
...
   }
 
+  private void computeDebugPresentation(final XValueNode node) {
+    myDebugProcess.getVmServiceWrapper()
+      .evaluateInTargetContext(myIsolateId, myInstanceRef.getId(), "toStringDeep()", new VmServiceConsumers.EvaluateConsumerWrapper() {
+        @Override
+        public void received(final InstanceRef toStringInstanceRef) {
+          if (toStringInstanceRef.getKind() == InstanceKind.String) {
+            String content = toStringInstanceRef.getValueAsString();
+            String summary = content.split("\n")[0];
```
alexander-doroshko: I usually try to avoid regexps in simple cases for performance reasons. Here looks like it can be replaced with `substring()` + `indexOf()`.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -149,10 +149,42 @@ public void computePresentation(@NotNull final XValueNode node, @NotNull final X
...
+            node.setPresentation(getIcon(), myInstanceRef.getClassRef().getName(), summary, true);
+            if (toStringInstanceRef.getValueAsStringIsTruncated()) {
+              addFullStringValueEvaluator(node, toStringInstanceRef);
+            }
+            else {
+              node.setFullValueEvaluator(new ImmediateFullValueEvaluator("View", content));
```
alexander-doroshko: Probably we don't need 'View' link if `content` is single-line?

alexander-doroshko:
> That's a typo.

Not so bad for a typo! :). BTW I noticed that some packages do already have `toDebugString()` in some classes.

So, @jingbian, @jwren, @devoncarew, @jacob314, @lambdabaa, can you confirm that `toStringDeep()` is the final name and trying to call it for every object in the debugger is the right thing to do? If you need more time for internal discussion, no rush here.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
jingbian: Fixed. Thanks!

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
jingbian: Good suggestion. Thanks!

jingbian:
Hi @alexander-doroshko , I've updated the code. Thanks for your help!

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/frame/DartVmServiceValue.java
alexander-doroshko: I mean, when `content` is just one line then it is equal to `summary` and `summary` is already shown, so in such case, we probably don't need `View` link at all.

jingbian: Yes, I get it. Take a look at the file after all three commits.

jingbian: The view link is only added when it's multi-line now.

alexander-doroshko: Correct! Sorry, I looked at a wrong commit.

jacob314:
> > That's a typo.
> 
> Not so bad for a typo! :). BTW I noticed that some packages do already have `toDebugString()` in some classes.
> 
> So, @jingbian, @jwren, @devoncarew, @jacob314, @lambdabaa, can you confirm that `toStringDeep()` is the final name and trying to call it for every object in the debugger is the right thing to do? If you need more time for internal discussion, no rush here.

`toStringDeep` is already used by most Flutter classes so I agree it is a good name to use for this case.

jwren:
I have no objection to the name `toStringDeep`.

alexander-doroshko:
Thanks, merged: https://github.com/JetBrains/intellij-plugins/commit/1be0c52fd24f32fac7ea0b2f3eac28e2d7db9c8b

jingbian:
> Thanks, merged: [1be0c52](https://github.com/JetBrains/intellij-plugins/commit/1be0c52fd24f32fac7ea0b2f3eac28e2d7db9c8b)

Sorry if I missed anything, but I cannot open the commit link. Just to confirm.

alexander-doroshko:
It works now, github repository is synced with a slight delay.

jingbian:
It works now! Thanks a lot!

PR #643, opened 2019-03-26
Add Dart plugin mechanism to listen to quick assist invocations

lambdabaa:
See https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003384140-Listening-for-whenever-a-quick-fix-is-applied-?page=1#community_comment_360000423640 for more context. We would like to quantify + understand when users apply quick fixes.

/cc @bwilkerson @alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -23,6 +23,7 @@ import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.EditorFactory;
 import com.intellij.openapi.editor.event.DocumentEvent;
 import com.intellij.openapi.editor.event.DocumentListener;
@@ -48,6 +49,7 @@ import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiFileSystemItem;
 import com.intellij.psi.search.FilenameIndex;
 import com.intellij.psi.search.SearchScope;
@@ -62,7 +64,10 @@ import com.jetbrains.lang.dart.DartFileListener;
 import com.jetbrains.lang.dart.DartFileType;
 import com.jetbrains.lang.dart.DartYamlFileTypeFactory;
 import com.jetbrains.lang.dart.assists.DartQuickAssistIntention;
+import com.jetbrains.lang.dart.assists.DartQuickAssistIntentionListener;
 import com.jetbrains.lang.dart.assists.QuickAssistSet;
+import com.jetbrains.lang.dart.fixes.DartQuickFix;
+import com.jetbrains.lang.dart.fixes.DartQuickFixListener;
 import com.jetbrains.lang.dart.ide.actions.DartPubActionBase;
 import com.jetbrains.lang.dart.ide.errorTreeView.DartFeedbackBuilder;
 import com.jetbrains.lang.dart.ide.errorTreeView.DartProblemsView;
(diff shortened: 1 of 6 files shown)


alexander-doroshko:
I'm afraid this won't always work. `IntentionManager` is an app-level service but `DartAnalysisServerService` is a project-level one. We need to register intentions only once per app (controlled by `ourIntentionsRegistered`). So with this change, only the first opened Dart project will have not empty `myQuickAssistIntentions`.

Also, see the comment that you've deleted: that 'moronic' code is there on purpose.

lambdabaa:
@alexander-doroshko Okay so I went ahead and switched the analysis server list referencing the intentions (and the registration methods) to be static. This fixes the issue, right? Also curious why the i variable wouldn't work for telling IntentionManager that the intentions are different? I reverted that change from the loop to calling it 20 times directly anyhow. Thanks!

alexander-doroshko:
We generally try to avoid static data.
While this may work, there's a risk that listeners that passed to `addQuickAssistIntentionListener()` will contain references to `Project` or anything else not app-level. This is especially dangerous when creating listeners as anonymous classes or lambdas: it might me not quite clear if such listeners contain references to something project-specific or not.

What about the following suggestion: 
- In `DartAnalysisServerService` have not static `myXListeners`, `addXListener()`/`removeXListener()` and `fireXEvent()` (names simplified)
- On intention invocation call `DartAnalysisServerService.getInstance(project).fireXEvent()`.

Side note: in the forum you wrote about quick fixes which are tied to errors. But this PR works with intentions, which do not have any relation to errors in code. Curious, what are you actually going to do in your listeners?

Nit: no need in having both PsiFile and Project as parameters, PsiFile.getProject() is to the rescue (unfortunately quite a few Platform API methods do have both).

As for strange intentions registration: Platform checks that the same intention class is not accidentally registered several times. Normally, this is not expected and might be a bug. But Dart case is special. To make Platform happy we register 20 *different classes*.

lambdabaa:
> We generally try to avoid static data.
> While this may work, there's a risk that listeners that passed to `addQuickAssistIntentionListener()` will contain references to `Project` or anything else not app-level. This is especially dangerous when creating listeners as anonymous classes or lambdas: it might me not quite clear if such listeners contain references to something project-specific or not.

Makes sense and thank you for explaining. I have taken your suggestion, PTAL :).

> Side note: in the forum you wrote about quick fixes which are tied to errors. But this PR works with intentions, which do not have any relation to errors in code. Curious, what are you actually going to do in your listeners?

We're interested in another plugin in listening for when quick fixes are applied and associating them with errors which we'll listen to (also through DartAnalysisServerService) in the downstream plugin.

> Nit: no need in having both PsiFile and Project as parameters, PsiFile.getProject() is to the rescue (unfortunately quite a few Platform API methods do have both).

Done! Thanks for the tip :+1: 



> As for strange intentions registration: Platform checks that the same intention class is not accidentally registered several times. Normally, this is not expected and might be a bug. But Dart case is special. To make Platform happy we register 20 _different classes_.

Oh I got it so it's not different instances, it's different classes that the API wants.

alexander-doroshko:
> listening for when quick fixes are applied and associating them with errors

Then why this PR is not about quick fixes? :)
Intention actions (aka quick assists) - that's a different entity, which is not related to errors by any means.

lambdabaa:
Sorry I misspoke. Quick assists is correct. I am mistakenly saying fix
rather than assist!

bwilkerson:
I think we actually need data for both.

The data I want to collect is, when a user looks for a fix for a diagnostic, which, if any, of the available suggestions do they select. (I'm almost more interested in the case where a user wanted a fix but didn't find one that sounded right than I am in the case where one was selected.)

That said, many of the "fixes" for lint rules are actually implemented as assists so that they are available to users whether or not the lint has been enabled. So when there is a lint violation we need to know whether they looked for an assist as well. (As far as I know, the same UI affordance is used for both, so I'm not sure why there would be a different API, but that's probably beside the point.)

lambdabaa:
> I think we actually need data for both.

Okay, I updated this patch so there are now two hooks so that we can pick up quick fix or quick assist applications. Apologies for my confusion.

alexander-doroshko:
@lambdabaa Sorry, something is wrong with the PR in its current state. There are 2 different `DartQuickFixListener.java` files, and also a bunch of compilation errors.

alexander-doroshko:
@bwilkerson
> I'm almost more interested in the case where a user wanted a fix but didn't find one that sounded right than I am in the case where one was selected.

If I'm not mistaken, this PR doesn't help to accomplish this.
Maybe you'd want to work on the Dart plugin part and the downstream part in parallel, just to make sure that the API we're going to add to the Dart plugin fits your needs.

lambdabaa:
> @lambdabaa Sorry, something is wrong with the PR in its current state. There are 2 different DartQuickFixListener.java files, and also a bunch of compilation errors.

Sorry about that! I think things got confused in one of the updates. Should be fixed now.

lambdabaa:
> I'm almost more interested in the case where a user wanted a fix but didn't find one that sounded right than I am in the case where one was selected.

> If I'm not mistaken, this PR doesn't help to accomplish this.

That's correct. This is purely aiming to provide a subscription API for invoked quick fixes and intention actions.

alexander-doroshko:
I'm afraid it's still messed up. But I tried to fix it myself. Please have a look: https://github.com/JetBrains/intellij-plugins/commit/b2c77da6dca41f64a4044d47598dfc79fe9c562c

If you decide that you don't need this API please let me know, I'll remove it from the plugin.

PR #644, opened 2019-03-27
Add action that calls the Dart VM Service reloadSources

fkettelhoit:
Adds an action to the Debugger window that exposes the `reloadSources` RPC of the Dart VM Service (similar to Flutter's Hot Reload, but for any Dart program):

https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#reloadsources

If the debugger is attached to a Dart isolate, the action saves all files and reloads the sources into the currently running isolate. As an example, I used the following Dart program:

```
main(List<String> arguments) {
  _ping();
}

void _ping() {
  Future.delayed(Duration(milliseconds: 4000)).then((_) {
    _printTime();
    _ping();
  });
}

void _printTime() {
  print('ping? ${DateTime.now()}');
}
```

The `Reload Sources (Dart)` action can then be used to change the output of `_printTime()` while the isolate is still running. The change will take effect once the next future is completed and `_printTime()` is called:

Dart Reload Sources Action

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -302,5 +302,6 @@
     </group>
     <action id="Dart.pub.cache.repair" class="com.jetbrains.lang.dart.ide.actions.DartPubCacheRepairAction" text="Pub: Repair Cache..."
             description="Run 'pub cache repair'"/>
+    <action id="Dart.reload.sources" class="com.jetbrains.lang.dart.ide.runner.actions.DartReloadSourcesAction"/>
   </actions>
 </idea-plugin>
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -208,6 +208,8 @@ open.observatory.action.text=Open Observatory
 open.observatory.action.description=Open Observatory: a Profiler for Dart apps
 dart.pop.frame.action.text=Drop Frame (Dart)
 dart.pop.frame.action.description=Pop the current frame off the stack
+dart.reload.sources.action.text=Reload Sources (Dart)
+dart.reload.sources.action.description=Save files and reload sources of current isolate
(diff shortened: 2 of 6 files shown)


jwren:
Thank you for the contribution!  Let me talk to the Dart VM team and JetBrains folks to make sure that this VM protocol isn't about to be deprecated and that this is the best place to contribute the action in the IntelliJ Debugger UI.

jwren:
I have chatted with developers and folks here on the Dart team, while the Dart VM does have this protocol, it is provided to support the flutter_tools stack.  The consequence here is that there are implicit expectations around this API that would be broken. Given that we'd own this going forward and the possible infrastructure improvements, I have to unfortunately push back on this PR.

Thank you for the contribution-- developer tools are important and impact their users.  The YouTrack bug tracking system, https://youtrack.jetbrains.com/issues?q=%23dart%20%23Unresolved, has a bunch of ideas and opportunities to contribute in other ways to the Dart Plugin.

- Jaime

alexander-doroshko:
@jwren, thanks for the update!
@fkettelhoit, thank you once again for the contribution! Sounds like we need to wait for the Dart VM to provide a more official way to implement the feature. So closing the PR.

fkettelhoit:
@jwren That's disappointing and somewhat unexpected, given that `reloadSources` is part of the public VM service RPCs and nothing in the documentation suggests any implicit expectations regarding how it is to be used. How are developers expected to know which of these "public" RPCs can be relied on and which ones could change in the future?

Also, is there any documentation or discussion about these "implicit expectations around this API" that is publicly accessible? Was your discussion with the Dart VM team public so that I could at least read up on the reasons for the decision?

I get that Dart is still primarily used inside Google and that they have no obligations to external users of Dart. But if the documented parts of the official VM service are not an "official way to implement this feature", then I don't know how you expect external developers to contribute in any meaningful way. Adding a link to the bug tracker with "opportunities to contribute" right after rejecting the PR with a minimum of information is somewhat ironic. Lesson learned, I guess...

PR #646, opened 2019-04-01
Make Dart IntelliJ plugin recognize includedSuggestionKinds if includedElementKinds is missing from CompletionSuggestion

lambdabaa:
Fixes https://github.com/flutter/flutter-intellij/issues/3338.

This change makes it so that the Dart plugin will respect an "includedSuggestionKinds" field on `CompletionSuggestion` in place of "includedElementKinds" if the latter is missing. The reason for this change is that Flutter cut a version of the Dart SDK with a WIP version of the new code completion spec, and we only added support for the latest version in the Dart plugin. Adding this fallback will make it so that code completion continues to work for Flutter SDK 1.2.1 stable users.

If it's possible to publish a new Dart plugin ASAP after this is landed, it would be greatly appreciated as code completion is mostly unavailable for a significant number of Flutter users.


jwren:
LGTM

jwren:
@alexander-doroshko Would it be possible to have this patched back into the appropriate 191* branch and re-released?

PR #649, opened 2019-04-16
Update the Dart plugin to the Dart Analysis Server protocol version 1.26.0

jwren:
@alexander-doroshko @bwilkerson @devoncarew @lambdabaa

Recent updates to the Dart Analysis Server protocol have only been to add a specific API or additional parameter, building up technical debt.  This PR attempts to pay it down by pulling in all generated types, the AnalysisServer.java file, and updates the *Consumer.java files appropriately to version 1.26.0.

https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_analysis.getReachableSources - note that some APIs are not shown in the linked html as they are `experimental`.

Changes to DartAnalysisServerService.java and RemoteAnalysisServerImpl.java were updated appropriately.


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -133,7 +133,7 @@
 
   // Do not wait for server response under lock. Do not take read/write action under lock.
   private final Object myLock = new Object();
-  @Nullable private AnalysisServer myServer;
+  @Nullable private RemoteAnalysisServerImpl myServer;
```
devoncarew: I'm curious why references to `AnalysisServer` were changed to `RemoteAnalysisServerImpl`? Shouldn't we use the more generic interface? Will this have implications for clients who use this as API (the Flutter plugin)?

jwren: The reason changed it was do to the removal of `generateUniqueId()` and 2 `sendRequestToServer()` methods.  Instead of adding these to the generated AnalysisServer I as I copied them over, I changed this variable to be more specific it its type.  I don't have a strong preference if folks do think this should remain generic.

It is private and not referenced by the Flutter plugin so it won't effect plugins downstream.

devoncarew: gotcha, thanks for the explaination

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1546,15 +1546,16 @@ public void onError(final RequestError error) {
...
+                                            final int offset) {
     final String filePath = FileUtil.toSystemDependentName(file.getPath());
     final Ref<SourceFileEdit> resultRef = new Ref<>();
 
     final AnalysisServer server = myServer;
     if (server == null || StringUtil.compareVersionNumbers(mySdkVersion, "1.25") < 0) return null;
 
     final CountDownLatch latch = new CountDownLatch(1);
-    server.edit_importElements(filePath, importedElements, new ImportElementsConsumer() {
+    server.edit_importElements(filePath, importedElements, offset, new ImportElementsConsumer() {
```
alexander-doroshko: I've wrapped the offset into `getOriginalOffset()` to make sure it works on Windows.

Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/AnalysisError.java
```diff
@@ -122,9 +121,10 @@ public static AnalysisError fromJson(JsonObject jsonObject) {
     Location location = Location.fromJson(jsonObject.get("location").getAsJsonObject());
     String message = jsonObject.get("message").getAsString();
     String correction = jsonObject.get("correction") == null ? null : jsonObject.get("correction").getAsString();
-    String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
+    String code = jsonObject.get("code").getAsString();
```
alexander-doroshko: I've reverted this line back to
```
String code = jsonObject.get("code") == null ? null : jsonObject.get("code").getAsString();
```
Otherwize the Dart plugin turns into a brick for older SDKs (and we are still compatible with SDK 1.12+)

Dart/thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/Outline.java
```diff
@@ -96,9 +103,11 @@ public static Outline fromJson(Outline parent, JsonObject outlineObject) {
     Element element = Element.fromJson(elementObject);
     int offset = outlineObject.get("offset").getAsInt();
     int length = outlineObject.get("length").getAsInt();
+    int codeOffset = outlineObject.get("codeOffset").getAsInt();
+    int codeLength = outlineObject.get("codeLength").getAsInt();
```
alexander-doroshko: Aren't `codeOffset` and `codeLenght` optional? I had to change these 2 lines to
```
    int codeOffset = outlineObject.get("codeOffset") == null ? offset : outlineObject.get("codeOffset").getAsInt();
    int codeLength = outlineObject.get("codeLength") == null ? length : outlineObject.get("codeLength").getAsInt();
```
in order not to crash with older SDKs.

jwren:
Thank you @alexander-doroshko!

PR #650, opened 2019-04-17
Provide package info for constructors in the type area

lambdabaa:
We were seeing an issue where multiple libraries re-export a constructor which makes it difficult to know from where you're importing a constructor. This patch makes it so that we use the space in type text if there are parameters that would conflict (rather than dropping this information altogether).

B0H4ArypjAu

/cc @alexander-doroshko @scheglov @devoncarew @bwilkerson

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -401,10 +401,9 @@ public class DartServerCompletionContributor extends CompletionContributor {
       }
 
       // If this is a class or similar global symbol, try to show which package it's coming from.
-      if (displayUri != null &&
-          element.getTypeParameters() == null &&
-          element.getParameters() == null) {
-        lookup = lookup.appendTailText(" (" + displayUri + ")", true);
+      if (displayUri != null) {
+        String packageInfo = "(" + displayUri + ")";
+        lookup = lookup.withTypeText(StringUtils.isEmpty(returnType) ? packageInfo : returnType + " " + packageInfo , true);
       }
 
       // icon


scheglov:
Would it be better to be consistent and always show the library URI in the right column?

bwilkerson:
That was my first thought as well. Consistency in a UI presentation generally makes it easier for the user to find what they're looking for.

lambdabaa:
I was thinking that of a case where there might be both a return type and a
display uri. But perhaps that cannot happen?

bwilkerson:
Yes, it's possible to have both a return type and a display URI. But I don't understand what that has to do with consistency. It seems to me that we should decide what information we're displaying right aligned, and _only_ display that information in that location.

I don't know whether there are conventions in IntelliJ that we should be honoring, but that should be a factor in the bigger decision.

lambdabaa:
> Yes, it's possible to have both a return type and a display URI. But I don't understand what that has to do with consistency. It seems to me that we should decide what information we're displaying right aligned, and _only_ display that information in that location.

> I don't know whether there are conventions in IntelliJ that we should be honoring, but that should be a factor in the bigger decision.

Right so in Java and currently in Dart we show type info in the right hand column. Java doesn't have this particular issue because constructor parameters are never shown (just the class name as below).

Screen Shot 2019-04-17 at 5 40 48 PM

However, since the `new` keyword is optional in Dart there is an ambiguity between class names and constructor invocations within code completion. Basically it's only because we're showing the optional parameters in Dart that a user knows what's a class vs. a constructor invocation. Although it's not perfectly consistent it is an improvement over the status quo where users don't know where the constructor suggestions are coming from at all.

scheglov:
While I agree that displaying library somewhere is an improvement, I don't agree that this is how we should decide that placing libraries inconsistently is OK.

alexander-doroshko:
So looks like we have 3 starting points:
- We want users to see all the useful information about an item: type parameters `<T>`, parameters `(int i)`, return type `Map<K, V>` and URI `package:angular/angular.dart`.
  - Parameters string is sometimes extremely long
  - The second place goes to URI. Sometimes it may be quite long as well
  - Type parameters are usually not so long but there might be pathological cases
- We want consistency in Dart completion items presentation.
- We want Dart completion items presentation to be similar to Java's and others.

Just some thoughts:
- I think being similar to Java is a bit less important, but consistency and completeness are. 
- Shortening:
  - It's definitely fine to shorten too long parameters. Maybe to save space it might be fine to sometimes shorten it to `(...)`?
  - It's probably acceptable to shorten type parameters, maybe down to `<>` if needed (WDYT?)
  - URI might be shortened down to file name, not sure how acceptable it is.
  - Return type with parameters type might be shortened to `Foo<...>` but again, I'm not sure.
- If we keep thinking about shortening, it might be hard to find optimal logic. Not obvious at all for me.
- Right column: return type and/or URI? Hard question. Thoughts welcome. Screenshot for thoughts: what would you improve and where would you add URI?
image

devoncarew:
some thoughts:
- I like the shortening ideas above
- when we show package info for a symbol, we should be consistent about where we show it
- I do like using the right-hand side for the package info (for the completion decoration, I think that's using `withTypeText()`)
- we may not need to be consistent in having all symbol types have the same presentation. Java does have a different presentation for classes vs methods
- for us, that may mean that constructors look like `Foo(param1, param2)      Foo` and types like `Foo               package:bar`? Or, we use the right hand column to show the best info about the containing element; for some things that may by the package name, and for others the class name.

lambdabaa:
@alexander-doroshko @bwilkerson @scheglov @devoncarew 

Here's an iteration that puts both the type and package info right-aligned. The type is to the left of the package so that it will always be visible (even if there is a long package name). Either package or return type can be missing and the other will show by itself.

Screenshot from 2019-04-18 11-57-20

scheglov:
Looks better, we might land this and observe how it works for us / users.

alexander-doroshko:
I've pushed the PR. It may appear on GitHub with a bigger delay this time (infrastructure maintenance).

PR #651, opened 2019-04-24
Make RequestListener and ResponseListener pass Strings instead of JsonObjects

lambdabaa:
This change switches the Dart plugin's `RequestListener` and `ResponseListener` objects to use primitive `String` instead of `JsonObject`. The rationale is so that downstream plugins can make their own choices about JSON parsing and representation instead of needing to use gson at a specific version.

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/com/google/dart/server/RequestListener.java
+++ b/Dart/thirdPartySrc/analysisServer/com/google/dart/server/RequestListener.java
@@ -1,11 +1,9 @@
 // Copyright 2000-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 package com.google.dart.server;
 
-import com.google.gson.JsonObject;
-
 /**
  * Listener for arbitrary requests sent from the analysis server.
  */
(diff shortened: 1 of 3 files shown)


lambdabaa:
@alexander-doroshko: if this can be landed and cherry-picked onto a 2019.1 version soon, we would very much appreciate it!

alexander-doroshko:
Pushed

lambdabaa:
> Pushed

Hey @alexander-doroshko -- would you be able to cherry pick this onto a 191.x plugin branch/release? I am not seeing it at the 191.7479 tag.

alexander-doroshko:
Some commits may be messed up, check file content, this change is there in the 191.7479 branch. https://github.com/JetBrains/intellij-plugins/blob/191.7479/Dart/thirdPartySrc/analysisServer/com/google/dart/server/RequestListener.java

lambdabaa:
> Some commits may be messed up, check file content, this change is there in the 191.7479 branch. https://github.com/JetBrains/intellij-plugins/blob/191.7479/Dart/thirdPartySrc/analysisServer/com/google/dart/server/RequestListener.java

Oh great, thank you!

PR #652, opened 2019-05-07
Fix issue where DartContributedReference.myRefRange starts at negative numbers

lambdabaa:
The Dart plugin throws a lot of exceptions like `String index out of range: -593` from the `element.getText().substring()` call in the `DartContributedReference` constructor. I could be totally off, but the caller here https://github.com/JetBrains/intellij-plugins/blob/a2032f89fc97f0e997e0440238f815655329276e/Dart/src/com/jetbrains/lang/dart/resolve/DartResolver.java#L148 doesn't look like we'd need further to subtract the element's start offset?

/cc @alexander-doroshko @jwren @devoncarew

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/psi/DartContributedReference.java
+++ b/Dart/src/com/jetbrains/lang/dart/psi/DartContributedReference.java
@@ -28,9 +28,7 @@ class DartContributedReference implements PsiPolyVariantReference {
                                   @NotNull final DartServerData.DartNavigationRegion navigationRegion) {
     myElement = element;
     myNavigationRegion = navigationRegion;
-    final int startOffset =
-      InjectedLanguageManager.getInstance(element.getProject()).injectedToHost(element, element.getTextRange().getStartOffset());
-    myRefRange = TextRange.from(navigationRegion.getOffset() - startOffset, navigationRegion.getLength());
+    myRefRange = TextRange.from(navigationRegion.getOffset(), navigationRegion.getLength());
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
@lambdabaa, we definitely need to subtract element's start offset.
- see `getRangeInElement()` javadoc
- see `DartServerResolverTest.testRefsInDocComments()`, it fails if this PR is applied
- try this code. Reference inside doc comment won't work if this PR is applied, also `StringIndexOutOfBoundsException` is thrown as you edit the file.
```
main() {

}

/// [Foo] is awesome
class Foo{}
```

The problem is trickier. See file history. Originally we had
```
myRefRange = TextRange.from(navigationRegion.getOffset() - element.getTextRange().getStartOffset(), navigationRegion.getLength());
```
But with this code, we had dozens of OutOfBoundsException for Angular projects because of code injection in HTML. The fix was to convert injected offsets using `injectedToHost()` method. It doesn't change offset if there's no injection and does convert if there is. Unfortunately, no test added for this conversion. And unfortunately, I don't remember the use case that was fixed using this `injectedToHost()`. And it might appear that this fix is not complete.

Do you know how to reproduce the exception? I'd be happy to see the file text and server navigation data for this file.

devoncarew:
@lambdabaa, do we have a repro here? Otherwise, Alex can add code to try and catch and report the issue.

alexander-doroshko:
I've added diagnostics: https://github.com/JetBrains/intellij-plugins/commit/d6bad61b13abc3991a72f46d3ac1ded9f102da9a. I think this PR may be closed. If you get an idea about STR please let me know.

PR #654, opened 2019-05-16
Provide getter for DartQuickFix#getSourceChange

lambdabaa:
This will allow DartQuickFixListener implementations to read the source change off of this object when beforeQuickFixInvoked is triggered. This is a follow-up from #643. Would you be able to cherry pick this onto 2019.1 @alexander-doroshko?

/cc @jwren


lambdabaa:
Actually I think this is not needed since we have access to the cursor position, nevermind!

PR #655, opened 2019-05-18
WEB-38855 - Angular @Attribute decorator not recognized

lppedd:
As per YouTrack [issue](https://youtrack.jetbrains.com/issue/WEB-38855), here is my attempt at fixing the problem.  
It's my first commit here so I hope to have done everything correctly.

Being that the `@Attribute` directive is applicable to constructor parameters, both shorthand field declarations and simple arguments, I decided to create a new `Angular2Element` named `Angular2DirectiveCtorParameter` and extend `Angular2SourceDirectiveDecoratedCtorParameter`, which is a general object to represent decorated constructor's parameters.

I'm really open to suggestion, as maybe something could have been done in a easier way!
Also, there are points missing:
- I'm still missing the Stubs part, as I had difficulties understanding how to correctly implement them.
However, that was because I was trying to work on `Angular2DirectiveProperty `metadata. Now I can handle a separate object.
- Maybe improvable behavior:
When having a definition such as
```
constructor(@Attribute('type') type: ButtonType = 'default') {}
```
the `type` parameter cannot navigate to its usage, because it isn't a class property.
```
<my-button
  [disabled]="!formGroup.valid
  type="dashed"
>
  Save
</my-button>
```


AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
```diff
@@ -327,11 +328,23 @@ public PsiElement getDeclaration() {
...
+        final Collection<PsiElement> propertiesPsiElements =
+              StreamEx.of(directive.getInputs())
+                      .filter(property -> {
+                        final boolean isOneTimeBinding = isOneTimeBindingProperty(property);
+                        return myInfo.name.equals(property.getName()) && isOneTimeBinding;
+                      })
+                      .map(Angular2Element::getNavigableElement)
+                      .toList();
+
+        final Collection<PsiElement> ctorParametersPsiElements =
```
piotrtomiak: You could use `StreamEx.concat()` and avoid second toList() call.

AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
```diff
@@ -571,6 +585,18 @@ private Angular2AttributeDescriptor createOneTimeBinding(@NotNull Angular2Direct
              : null;
     }
 
+    @NotNull
+    private Angular2AttributeDescriptor createCtorOneTimeBinding(@NotNull Angular2DirectiveCtorParameter info) {
```
piotrtomiak: Maybe let's stick to "attribute" naming, so `createAttributeBinding`

AngularJS/src/org/angular2/entities/Angular2CtorBoundParameters.java
```diff
@@ -0,0 +1,25 @@
+package org.angular2.entities;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * A simple object which hold constructor's bound parameters.
+ */
+public class Angular2CtorBoundParameters {
```
piotrtomiak: I would try to avoid this object. Inputs and outputs can similarly be defined in constructor params and there is no special object to hold them.

AngularJS/src/org/angular2/entities/Angular2DirectiveCtorParameter.java
```diff
@@ -0,0 +1,18 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import com.intellij.lang.javascript.psi.JSType;
+
+/**
+ * Represents a constructor parameter (property/argument).
+ * This kind of element is always one-time binding, so there is no need to check.
+ */
+public interface Angular2DirectiveCtorParameter extends Angular2Element {
```
piotrtomiak: Please name the class `Angular2DirectiveAttribute`

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -114,6 +122,16 @@ public boolean isTemplate() {
     );
   }
 
+  @NotNull
+  private Angular2CtorBoundParameters getCachedCtorParameters() {
```
piotrtomiak: Please incorporate into `getProperties()` method to avoid traversing class properties twice.

lppedd:
@piotrtomiak thanks for the comments, I'll fix them! Regarding the `*Attribute` naming, I decided to go for `*CtorParameter` because there might be the necessity to bind something else other than `@Attribute` decorated parameters. But maybe it's unlikely to be so, what do you think?

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -143,6 +161,51 @@ public boolean isTemplate() {
                 Collections.unmodifiableCollection(outputs.values()));
   }
 
+  @NotNull
+  private Angular2CtorBoundParameters getCtorParameters() {
```
piotrtomiak: I am not sure, but I think you should traverse the class hierarchy as the `@Attribute` could be specified in parent class.

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -143,6 +161,51 @@ public boolean isTemplate() {
...
+        continue;
+      }
+
+      final String paramName = ctorParam.getName();
+      final Optional<String> bindingName =
+            Arrays.stream(attributeList.getDecorators())
+                  .filter(d -> Angular2DecoratorUtil.ATTRIBUTE_DEC.equals(d.getDecoratorName()))
+                  .findFirst()
+                  .map(Angular2SourceDirective::getStringParamValue)
+                  .map(n -> StringUtil.notNullize(n, paramName));
```
piotrtomiak: Instead of having if below you could:
`.ifPresent(key -> attributes.putIfAbsent(key, new Angular2SourceDirectiveDecoratedCtorParameter(key, ctorParam))`

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -143,6 +161,51 @@ public boolean isTemplate() {
...
+   * 			The class constructor
+   */
+  @NotNull
+  private Angular2CtorBoundParameters processCtorParameters(final JSFunction ctor) {
+    final Map<String, Angular2DirectiveCtorParameter> attributes = new LinkedHashMap<>(8);
+
+    for (final JSParameter ctorParam : ctor.getParameterVariables()) {
+      final JSAttributeList attributeList = ctorParam.getAttributeList();
+
+      if (Objects.isNull(attributeList)) {
```
piotrtomiak: Maybe `StreamEx.ofNullable(ctorParam.getAttributeList())
.flatArray(getDecorators())`

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -143,6 +161,51 @@ public boolean isTemplate() {
                 Collections.unmodifiableCollection(outputs.values()));
   }
 
+  @NotNull
+  private Angular2CtorBoundParameters getCtorParameters() {
+    // The @Attribute decorated parameters can be found (and are valid) only
+    // in the class constructor, thus we have to get hold of it and process each parameter
+    final JSFunction ctor = getTypeScriptClass().getConstructor();
```
piotrtomiak: There can be multiple constructors in TS class, which one is valid?

AngularJS/src/org/angular2/entities/source/Angular2SourceDirectiveDecoratedCtorParameter.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @NotNull
+  @Override
+  public PsiElement getNavigableElement() {
+    return myParameter.getNavigationElement();
+  }
+
+  @NotNull
+  public Collection<ES6Decorator> getDecorators() {
```
piotrtomiak: I think this one is not needed.

piotrtomiak:
@lppedd My approach is that at this point it's clearer to use `*Attribute` naming and if it needs generalization in the future it will be a good moment to do it than, because we will have more information :)

Overall it's a solid change and I am really happy to have a community contribution to the project!

As far as the metadata support is concerned, you can generate metadata using Angular CLI lib project. I think for `@Attribute('foo')` you should have decorator arguments visible in the metadata.  Note that you need to increase number in `MetadataJsonStubBuilder#getStubVersion` when you change stubs implementation.

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
lppedd: But Angular needs to know how to inject the constructor parameter, or it will emit a `NullInjectorError`. I did some experiments and the only constructor which seems to matter is the sub-class one, if there is a hierarchy.

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
lppedd: If there is only one constructor, I'll use that one. If there are multiple, I'll pick the only one which has `isOverloadImplementation == true`

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
piotrtomiak: Ok, great.

AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
lppedd: Yes that's better, thanks! I used `inputsStream.append(attributesStream).toList()`.

lppedd:
@piotrtomiak 
For reference, this is the `metatadata.json` portion which shows an `@Attribute` decorated parameter, the fourth one.
```
{
   "parameterDecorators":[
      null,
      null,
      null,
      [
         {
            "__symbolic":"call",
            "expression":{
               "__symbolic":"reference",
               "module":"@angular/core",
               "name":"Attribute",
               "line":49,
               "character":51
            },
            "arguments":[
               "name"
            ]
         }
      ],
      null
   ],
   "parameters":[
      {
         ...
      },
      {
         ...
      },
      {
         ...
      },
      {
         "__symbolic":"reference",
         "name":"string"
      },
      {
         ...
      }
   ]
}
```
That means I need to look inside `"parameterDecorators"`, and then, using the same index, extract the parameter name from `"parameters"`.
That means I'll have to add
```
@NonNls protected static final String SYMBOL_PARAMETER_DECORATORS = "parameterDecorators";
@NonNls protected static final String SYMBOL_PARAMETERS = "parameters";
```

piotrtomiak:
@lppedd I think that matching by index is the only way in case of metadata constructors and that it's the intended way of doing things with constructor parameters.

I want to avoid, if possible too much of a hierarchy in metadata, so attributes should be direct children of an `Angular2MetadataDirectiveStubBase`. You can prefix attribute names for instance with `___attr.`. The attribute metadata should contain the index to match it with the source when needed.

lppedd:
@piotrtomiak sorry if I didn't write back, had some personal problems. Just wanted to tell you the code you reviewed above is fixed now, and I'm studying the stub part.

piotrtomiak:
@lppedd No problem, I hope everything is fine! Let me know if you have any questions :)

lppedd:
@piotrtomiak thanks! 
A method I don't understand is `getCachedValueWithClassDependencies`, which is used to return properties (`Input `and `Output`). The `dependencies` array parameter is still obscure to me. What should be passed to it?

piotrtomiak:
> The dependencies array parameter is still obscure to me. What should be passed to it?
Cached values are being stored until they are garbage collected or when their dependencies change. The dependency can be a `PsiElement`, or a `ModificationTracker` and the list should include all of the sources of information, so that it's refreshed when the information source change. In case of `@Attribute` I think you need only the TypeScript class and the Metadata object of the directive, because those are the only sources of your information. In case of `Input` and `Output` you need to collect all the parent classes as well, as the values come from the whole hierarchy of a class.

lppedd:
@piotrtomiak okay, got it! Thanks : )
Another question regarding the `___attr.` prefix. This seems the same as what you already do with `___dec.` prefixed ones.
I've seen for each of the `___dec.` prefixed, an `Angular2MetadataStringStub` is created.
Then, they're read using the `findMember() `method
```
public MetadataElementStub findMember(@Nullable String name) {
  return membersMap.getValue().get(name);
}
```
which in turns asks `getChildrenStubs()`.
The thing is, you're using `___dec.` with fixed names, such as `___dec.inputs` or `___dec.outputs`. Meanwhile, with `___attr.`, the name would be dynamic, so I would have to iterate over `getChildrenStubs()` or save them somewhere, instead of accessing via a key as I was doing with a `Map` above.

Does this make sense or I'm completely off road?

piotrtomiak:
@lppedd You're right, and the reason is that decorators can contain complex expressions, so some of them need to be properly stubbed. I think the map is fine in your case, however, you need to store index and name, as some `@Attribute` can have different names assigned.

lppedd:
@piotrtomiak thank you! Yeah, in the experiment I did I was storing a `Map<String, Integer>` as `alias -> index`. I had to add a couple methods to read and write `<String, Integer>` maps though, as there were only `<String, String>` ones.
Just to complete this matter, if a user has exposed multiple `@Attribute`s with the same alias
```
constructor(
    @Attribute('type') private readonly type: ButtonType = 'default',
    @Attribute('type') private readonly typeTwo: ButtonType = 'default',
) { ... }
```
it will be possible to navigate only to the first one, obviously.

piotrtomiak:
@lppedd - sure, that's ok.

lppedd:
@piotrtomiak I pushed what I had done the previous days.
The methods `loadAttributesMapping` and `buildAttributesMapping` in `Angular2MetadataDirectiveStubBase` might be too complex. Let me know if I have to simplify them.

> I think you need only the TypeScript class and the Metadata object of the directive

Hope I got this part right. The "metadata object of the directive" should be the object I'm operating over (or better, in).

AngularJS/src/org/angular2/entities/metadata/psi/Angular2MetadataDirectiveBase.java
```diff
@@ -72,6 +76,12 @@ public Angular2DirectiveSelector getSelector() {
     return getCachedProperties().second;
   }
 
+  @NotNull
+  @Override
+  public Collection<? extends Angular2DirectiveAttribute> getAttributes() {
+    return getCachedValueWithClassDependencies(this::getAttributes);
```
piotrtomiak: You can use simpler `getCachedClassBasedValue`, because your results do not depend on any additional things, just metadata object and class object.

piotrtomiak: `return getCachedClassBasedValue(this::buildAttributes)`

AngularJS/src/org/angular2/entities/metadata/psi/Angular2MetadataDirectiveBase.java
```diff
@@ -96,6 +106,71 @@ public Angular2DirectiveSelector getSelector() {
...
+  }
+
+  /**
+   * Constructs {@link Angular2DirectiveAttribute} from stubs.
+   *
+   * @param tsClass
+   *       The TypeScript class from which to build metadata objects, if needed
+   */
+  @NotNull
+  private Collection<Angular2DirectiveAttribute> buildAttributes(@NotNull final TypeScriptClass tsClass) {
```
piotrtomiak: tsClass can be `null`, please annotate with `@Nullable`

AngularJS/src/org/angular2/entities/metadata/psi/Angular2MetadataDirectiveBase.java
```diff
@@ -96,6 +106,71 @@ public Angular2DirectiveSelector getSelector() {
...
+   *
+   * @param tsClass
+   *       The TypeScript class from which to build the metadata objects
+   */
+  @NotNull
+  private Result<Collection<Angular2DirectiveAttribute>> getAttributes(@NotNull final TypeScriptClass tsClass) {
+    return Result.create(buildAttributes(tsClass), tsClass, this);
+  }
+
+  /**
```
piotrtomiak: Please don't add comments on private methods.

AngularJS/src/org/angular2/entities/metadata/stubs/Angular2MetadataDirectiveStubBase.java
```diff
@@ -98,6 +133,60 @@ protected FlagsStructure getFlagsStructure() {
...
+   *
+   * @param source
+   *       The JSON metadata document
+   */
+  @NotNull
+  private Map<String, Integer> loadAttributesMapping(@NotNull final JsonObject source) {
+    return StreamEx.ofNullable(getPropertyValue(source.findProperty(MEMBERS), JsonObject.class))
+                   .map(toPropertyValue(CONSTRUCTOR, JsonArray.class))
+                   .nonNull()
+                   .map(JsonArray::getValueList)
```
piotrtomiak: use `.flatCollection`

AngularJS/src/org/angular2/entities/metadata/stubs/Angular2MetadataDirectiveStubBase.java
```diff
@@ -98,6 +133,60 @@ protected FlagsStructure getFlagsStructure() {
...
+                   .flatMap(StreamEx::of)
+                   .map(toCast(JsonObject.class))
```
piotrtomiak: use `select(JSonObject.class)`

AngularJS/src/org/angular2/entities/metadata/stubs/Angular2MetadataDirectiveStubBase.java
```diff
@@ -98,6 +133,60 @@ protected FlagsStructure getFlagsStructure() {
...
+    final Predicate<JsonObject> isAttributeDecorator = object -> {
+      final JsonObject expr = getPropertyValue(object.findProperty(EXPRESSION), JsonObject.class);
+      final String decoratorName = expr != null
+            ? readStringPropertyValue(expr.findProperty(REFERENCE_NAME))
+            : null;
+      return ATTRIBUTE_DEC.equals(decoratorName);
+    };
+
+    return EntryStream.of(paramDecorators.getValueList())
+                      .mapValues(toCast(JsonArray.class))
```
piotrtomiak: `selectValues(JsonArray)`

lppedd: @piotrtomiak thanks! `select*` is a lot cleaner.

AngularJS/src/org/angular2/entities/metadata/stubs/Angular2MetadataDirectiveStubBase.java
```diff
@@ -115,4 +204,14 @@ private void loadAdditionalBindingMappings(@NotNull Map<String, String> mappings
...
     }
   }
+
+  private static <T extends JsonValue> Function<JsonObject, T> toPropertyValue(
+        @NotNull final String property,
+        @NotNull final Class<T> clazz) {
+    return o -> getPropertyValue(o.findProperty(property), clazz);
+  }
+
+  private static <T extends JsonValue> Function<Object, T> toCast(@NotNull final Class<T> clazz) {
```
piotrtomiak: I think this can be removed if using `select`

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -122,6 +129,16 @@ public boolean isRegularDirective() {
     );
   }
 
+  @NotNull
+  private Collection<Angular2DirectiveAttribute> getCachedAttributeParameters() {
```
piotrtomiak: This is simple enough, so that it can be merged into `getAttributes`

AngularJS/src/org/angular2/entities/metadata/stubs/Angular2MetadataDirectiveStubBase.java
lppedd: @piotrtomiak yes, absolutely!

AngularJS/src/org/angular2/entities/source/Angular2SourceDirective.java
```diff
@@ -151,6 +168,53 @@ public boolean isRegularDirective() {
...
+
+  /**
+   * Analyze the class constructor's parameters to find {@code @Attribute} bound ones.
+   *
+   * @param ctor
+   * 			The class constructor
+   */
+  @NotNull
+  private Collection<Angular2DirectiveAttribute> processCtorParameters(final JSFunction ctor) {
+    final Map<String, Angular2DirectiveAttribute> attributes = new LinkedHashMap<>(8);
```
piotrtomiak: Can you make it one long Stream call ? :)

piotrtomiak:
@lppedd Thank you for your work! This is a really great contribution! I've added several minor remarks. Please also remove all of JavaDoc comments; we try to have self-documenting code and avoid comments almost at all costs :) Your code looks pretty clean and I don't think any comments are needed :) If you feel otherwise, let's discuss.

lppedd:
@piotrtomiak No no, it's okay to remove them if we have self-explaining code. Sometimes I abuse JavaDoc so thanks for reminding me.
Sorry for the `map(toCast...`, I didn't even notice the `select` method.

piotrtomiak:
> Sometimes I abuse JavaDoc so thanks for reminding me.

@lppedd You're welcome :D

> Sorry for the `map(toCast...`, I didn't even notice the `select` method.

No problem, `StreamEx` has so many cool things I am still finding some new ones my self ;) Like `ofNullable` e.g.

lppedd:
@piotrtomiak fixed and pushed!

piotrtomiak:
@lppedd Thank you very much for your contribution! 

I have allowed myself to modify your contribution a bit - I have squashed your commits, corrected formatting and made our internal inspections happy. 

Your changes has been manually merged into `master` branch as dfc38543ef5614c95393742e30db9c266be3f996. 

In separate commits I've updated attributes building logic to match my today's performance fixes, added some tests and added code completion for self-selecting attributes. I will be cherry-picking all the changes to `191`.

lppedd:
@piotrtomiak awesome! Thank you very much!
> corrected formatting and made our internal inspections happy.

For future contributions is it possible to have the formatting and inspection settings shared via VCS?

> I have allowed myself to modify your contribution a bit

I've looked at your edits and I'll use them for reference in the future.

> Just to complete this matter, if a user has exposed multiple @Attributes with the same alias

I will try out of curiosity to wire in the multi-target popup you also implemented for WEB-37928. Might be useful and I'll also study the codebase a bit more, if it's not a problem.

piotrtomiak:
> For future contributions is it possible to have the formatting and inspection settings shared via VCS?

I think it should be possible, I will try to get the settings synchronized.

> I will try out of curiosity to wire in the multi-target popup you also implemented for WEB-37928. Might be useful and I'll also study the codebase a bit more, if it's not a problem.

I think this should already work, as the support was generic and based on XML attribute descriptors. If there is something wrong with navigation the problem might be in `Angular2XmlElementSourcesResolver`. 

> Ah btw, should I target by default the master branch?

It would be best, but I think this time it wasn't possible as we haven't had latest master build released yet, so I guess it would be hard for you to compile from `master` - not sure though. Anyway, there is a release from `master` now.

lppedd:
@piotrtomiak 
> I think it should be possible, I will try to get the settings synchronized.

Thanks :)

> it wasn't possible as we haven't had latest master build released yet

Yeah, I tried working/building over `master` yesterday but it seems it is based, as you wrote, on the latest code pushed to your `intellij-community` and internal repos, which are not `LATEST-EAP` (still `191.*`).

> I think this should already work
I'll take a look, thanks again!

piotrtomiak:
> Yeah, I tried working/building over master yesterday but it seems it is based, as you wrote, on the latest code pushed to your intellij-community and internal repos, which are not LATEST-EAP (still 191.*).

I've just tried and basically you need to match checked out branch with the release branch. In case of Angular plugin it's mostly OK, because in 96% it's in sync between `master` and `191`. However, you might want to build from the latest EAP branch `192.4205` to ease cherry-picking for me :) Additionally, there are no more features being pushed to `191` - @Attribute support was the last one I think, so it's also better to stay as close to the latest version as possible.

lppedd:
@piotrtomiak 
> However, you might want to build from the latest EAP branch 192.4205

Okay, I'll just clone the `intellij-community` repository and build over the latest released branch.
Until now I used the `org.jetbrains.intellij` Gradle plugin, which point to `https://www.jetbrains.com/intellij-repository/releases/`, which has `191.7479` as latest available.

Thanks for your help! : )

piotrtomiak:
> Okay, I'll just clone the intellij-community repository and build over the latest released branch.

@lppedd I don't think it'll help, as JavaScript support is proprietary and required by Angular. I think it might be better to stick with latest release branches.

PR #656, opened 2019-05-20
Increase timeouts for autocompletion to make large projects work.

DaveShuckerow:
Large users of Dart are encountering timeouts on requesting DAS for suggestions.

This change fixes their issues.

cc @jwren


jwren:
@lambdabaa we don't have any hard data on this do we?

@DaveShuckerow LGTM, especially for slow timeouts such as 3ms and APIs where the UI doesn't lock

lambdabaa:
LGTM, thanks @DaveShuckerow! The data that we do have says that 70% of completion requests finish in <=100ms.

alexander-doroshko:
I'm going to test each touched line in real to make sure that it won't cause freezes or bad user experience. This will take a bit of time.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -97,24 +97,24 @@
   private static final long UPDATE_FILES_TIMEOUT = 300;
 
   private static final long CHECK_CANCELLED_PERIOD = 10;
-  private static final long SEND_REQUEST_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
-  private static final long EDIT_FORMAT_TIMEOUT = TimeUnit.SECONDS.toMillis(3);
+  private static final long SEND_REQUEST_TIMEOUT = TimeUnit.SECONDS.toMillis(10);
```
devoncarew: As a drive by comment, we may want to normalize all these to `TimeUnit.MILLISECONDS.toMillis()` in order to make it easier to compare between entries? Some here use seconds and some milliseconds, and its easy to mistake the two when scanning down the list.

DaveShuckerow: This seems like a very good idea.

We definitely don't want, for instance, GET_SUGGESTIONS_TIMEOUT = 100 Seconds.

DaveShuckerow:
> I'm going to test each touched line in real to make sure that it won't cause freezes or bad user experience. This will take a bit of time.

Hi @alexander-doroshko, if you'll have to test this by hand, I think it might make more sense to shrink the scope of the changes based on the timeout problem we're seeing.  I believe we can get about the same improvement with increases to just the GET_ASSISTS_TIMEOUT and the GET_FIXES_TIMEOUT.

I'll test on my end and shrink this change if it works out.

alexander-doroshko:
Sounds great!

DaveShuckerow:
Done.  Alexander, please take a look when you get a chance.  Thanks!

alexander-doroshko:
I took a closer look at the problem and did some experiments.
- First of all, I want to clarify the problem we are trying to solve. Originally I thought that by 'suggestions' you mean code completion but it appears that I was wrong and you mean Quick Fixes and Intention Actions (aka Quick Assists). Are you sure you experienced problems caused by timeouts for *both*?
- Speaking about Quick Fixes. The method name is `askForFixesAndWaitABitIfReceivedQuickly()` and its doc says that in case of slow server it still doesn't ignore the late result but anyway notifies the consumer. And I made sure that it works fine in such 'async' mode and you get Quick Fixes in the editor as soon as server returns them. So I want to make sure once again: are you sure you have problems with Quick Fixes caused by timeout?
- Speaking about Intention Actions (aka Quick Assists). The method is `edit_getAssists()` and it doesn't work in async mode. If the server replies later than the timeout happens, the returned result is ignored. So I believe you could have this problem in large projects and that's the problem you are trying to solve by increasing the timeout. I played with this and noticed that `edit_getAssists()` may be called both in the EDT (aka AWT, aka Event Dispatch Thread) and in pooled threads. Longer timeout is not a problem in pooled threads: regardless of the timeout the method immediately exits on any user action (like typing or caret movement), thanks to the `awaitForLatchCheckingCanceled()` method and the underlying idea documented in the `ProcessCanceledException` class. But unfortunately, the trick with the `ProcessCanceledException` doesn't work when `edit_getAssists()` is called in EDT. In the latter case, IDE freezes completely until either timeout ends or a response from the server arrives. We can't afford user-visible UI freeze, that's why timeout was relatively short: 100ms. I have an idea. What about keeping timeout short when the method is called in EDT and making it longer (up to 1 second) - when in a pooled thread? Do you have a real project to test it in a real life? To check EDT use `ApplicationManager.getApplication().isDispatchThread()`. AFAIU, `edit_getAssists()` is called in EDT when user explicitly presses Alt+Enter, but in a pooled thread on typing or caret movement.

Short changes sometimes cause long discussions, sorry :)

DaveShuckerow:
Hi Alexander, no worries: it will take time to get this right, and even more time to fix it if we do a release and don't get the desired result.  We want to measure the problem three times and only cut one release.

The problem we're seeing is specifically in the alt+enter case, which I believe is called in EDT mode.  The reproduction case I'm using is to remove an import, then go to one of the unresolved names, hit alt+enter, and see if the suggestion to re-import the removed import comes up.  I'm looking at times when analysis is running on the project, so I'll try restarting the analysis server after the fix shows.

alexander-doroshko:
Ok, thanks for the scenario. 'Autocompletion' word in the PR title still confuses me a bit :).
Both Quick Fixes and Intention Actions (aka Quick Assists) show up in the Alt+Enter popup. But the scenario you described is about Quick Fixes - unlike Quick Assists they are tied to errors.

Unfortunately, I'm not sure we can improve anything with Quick Fixes. I strongly believe that we must not block UI. It means that to get Quick Fixes user may have to press Alt+Enter once again after getting nothing on the first attempt. As soon as a response from the server arrives user will see Quick Fixes. I don't see what we can improve without freezing IDE.

DaveShuckerow:
Thanks for the feedback, Alexander.

Do you have any objections to making the fix and assist timeouts instance variables, with default values identical to what they currently are?  This would allow us to consume the API in a Google-internal plugin like so:

```java
DartAnalysisServerService analysisService = DartPlugin.getInstance().getAnalysisService(project);
analysisService.setQuickFixesTimeout(TimeUnit.MILLISECONDS.toMillis(500));
```

This would allow us to tweak the behavior for our internal users who are regularly seeing fix requests timeout on them.  Some of these users get timeouts often enough that they sometimes copy-paste or hand-type their import statements.  

I want to get something to those users in the short-term that makes this easier.

alexander-doroshko:
> Do you have any objections to making the fix and assist timeouts instance variables, with default values identical to what they currently are?

Well, I think we might land this but I still hope we could find a better solution.

Today I've [increased timeouts](https://github.com/JetBrains/intellij-plugins/commit/13e6baea7e121ce6acfb9341037efb6f8ed3174e) when the server is called not from the EDT. I'm not sure this will help but why not give it a try?

Ideally, I'd be happy if Google internal devs used the same IDE environment as all other developers. I'm sure they don't like freezing UI as well as everyone else. And yes, they do need available Quick Fixes as well as everyone else.

We may want to experiment with something like a progress like 'Calculating Quick Fixes...'. Just a raw thought that came to my mind 1 second ago, not a call to action. I'm brainstorming about other possible and impossible solutions... Making the server faster... Sorry, probably not possible for huge projects.

alexander-doroshko:
In version 2020.1, `QuickFix.isAvailable()` and `QuickAssist.isAvailable()` will always be called in a background thread, even on explicit `Alt+Enter`. If `isAvailable()` takes too much time, user will see a cancelable progress. So, in 2020.1, we will always have long enough timeout to wait for the Analysis Server response. Feel free to try the feature by running IntelliJ IDEA + Dart plugin from sources (master branch).

PR #658, opened 2019-05-23
Dart hover change

jwren:
Dart Plugin hover changes: no "propagated" types, this notion was removed with the migration to Dart 2.  Also, the library name formatting is changing from being labeled as the "Containing library: " to simply being the first item.  A recent change in the Dart Analysis Server will also change the library name from what is after the 'library' keyword, to the importing URI, see https://dart-review.googlesource.com/c/sdk/+/10348


alexander-doroshko:
Thanks, lgtm. Please update `DartDocUtilTest` accordingly and I'll merge this.

I'm going to keep WEB-39001 open because as I understand it is about something else.

jwren:
Thanks Alex, updated `DartDocUtilTest`

alexander-doroshko:
Thanks, pushed.

PR #659, opened 2019-05-29
Max rather than sum for included suggestion relevance boost

lambdabaa:
From the [API spec](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html):

> If an AvailableSuggestion has relevance tags that match more than one
> IncludedSuggestionRelevanceTag, the maximum relevance boost is used.

/cc @alexander-doroshko @scheglov @DanTup

### DIFF
--- a/AngularJS/resources/messages/Angular2Bundle.properties
+++ b/AngularJS/resources/messages/Angular2Bundle.properties
@@ -13,7 +13,6 @@ angular.action.new-project.label-additional-parameters=Additional parameters:
 angular.action.new-project.label-defaults=Use the &default project setup
 angular.action.new-project.label-content-root-folder=Content root folder
 angular.action.new-project.label-project-name=Project
-angular.action.new-project.label-schematics-collection=Schematics collection:
 angular.action.new-project.name=Angular CLI
 angular.action.new-project.presentable-package-name=Angular &CLI:
 angular.action.new-project.wrong-folder-name={0} name ''{1}'' is not valid. {0} name must start with a letter, and must contain only alphanumeric characters or dashes. When adding a dash the segment after the dash must also start with a letter.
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -549,7 +549,7 @@ private static CompletionSuggestion createCompletionSuggestionFromAvailableSugge
       for (String tag : relevanceTags) {
         IncludedSuggestionRelevanceTag relevanceTag = includedSuggestionRelevanceTags.get(tag);
         if (relevanceTag != null) {
-          relevance += includedSuggestionRelevanceTags.get(tag).getRelevanceBoost();
+          relevance = Math.max(relevance, relevanceTag.getRelevanceBoost());
```
scheglov: Hm... "the maximum relevance *boost* is used".
Which means that the boost is *added* to the base relevance.

DanTup: This was my understanding too. It should be `base relevance + (maximum of tag boosts)`.

lambdabaa: Cool makes sense, updated!

alexander-doroshko: @scheglov, is it safe to start with zero assuming that there always be at least one non-negative relevance boost?

@lambdabaa, there are some not related files in the commit :)

lambdabaa: Thanks for pointing that out. `git commit --amend` not actually the same as `git rebase -i HEAD~2`...

lambdabaa: > @scheglov, is it safe to start with zero assuming that there always be at least one non-negative relevance boost?

There may be no relevance boosts but then 0 boost is what we want.

lambdabaa:
Moved to #660.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
scheglov: Yes, there are no negative boosts.

PR #660, opened 2019-05-30
Max rather than sum for included suggestion relevance boost

lambdabaa:
Continued from #659 since I made a git whoopsie, sorry!

From the [API spec](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html):

> If an AvailableSuggestion has relevance tags that match more than one
> IncludedSuggestionRelevanceTag, the maximum relevance boost is used.

/cc @alexander-doroshko @scheglov @DanTup


alexander-doroshko:
Thanks, pushed.

lambdabaa:
@alexander-doroshko Thanks! Would you be able to publish this change in a 2019.1.1 plugin?

alexander-doroshko:
Yes, I've cherry-picked it to the branch 191. We may release any time. But I guess this is a very minor fix, so we may wait a bit and collect more cherry-picks?

PR #661, opened 2019-05-30
Update the DartServerDocUtilTest tests with the latest changes from the Dart SDK

jwren:
Changes in the Dart Analysis Server: https://dart-review.googlesource.com/c/sdk/+/103701 and https://dart-review.googlesource.com/c/sdk/+/103961

The library name is no longer produced for tests to avoid conflicts with different formatting between Windows to other OSes.

### DIFF
--- a/Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerDocUtilTest.java
+++ b/Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerDocUtilTest.java
@@ -17,11 +17,12 @@ public class DartServerDocUtilTest extends CodeInsightFixtureTestCase {
   }
 
   public void testAbstractClassSig() {
-    doTest("<code>abstract class Foo extends Bar<br><br></code>", "abstract class <caret>Foo extends Bar { }\nclass Bar { }");
+    doTest("<code><b>test.dart</b><br>abstract class Foo extends Bar<br><br></code>",
+           "abstract class <caret>Foo extends Bar { }\nclass Bar { }");
   }
 
   public void testClassMultilineDoc1() {
-    doTest("<code>class A<br><br></code>\n" +
+    doTest("<code><b>test.dart</b><br>class A<br><br></code>\n" +
            "<p>doc1\n" +
            "doc2\n" +
            " doc3</p>\n" +
@@ -46,7 +47,7 @@ public class DartServerDocUtilTest extends CodeInsightFixtureTestCase {
   }
 
   public void testClassMultilineDoc2() {
-    doTest("<code>abstract class A<br><br></code>\n<p>doc1\n" +
+    doTest("<code><b>test.dart</b><br>abstract class A<br><br></code>\n<p>doc1\n" +
            "doc2\n" +
            " doc3\n" +
            "doc4\n" +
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Hm, is that what we want to see in docs?
image

jwren:
Yes, recent changes from myself and others in the analysis server have this changed a bit to:

```
C:/dart_pro...main.dart
void set x(int x)

Containing class: A
```

as the updated tests in this PR show.

The `C:/..` is the importing URI, other examples include `dart:*` and `package:*`, in this case it is a local reference, which would be imported directly as a relative file.

Additional changes coming include making methods and constructors with lots of arguments, formatted like the following instead of just wrapping:

```
foo(int x,
     int y, 
     int z);
```

devoncarew:
I would think we'd want to show the url you'd import it from here. So instead of `file:///C:/dart_projects/dart_web_app/web/main.dart`, something like `web/main.dart`, or `package/foo/bar.dart` for package references. Showing a file: url with a driver letter seems like we're leaking implementation details through.

alexander-doroshko:
Could you clarify the latest server changes regarding `HoverInformation.containingLibraryName`?
It's [current description](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_HoverInformation) is 
`The URI of the containing library, examples here include "dart:core", "package:.." and file uris represented by the path on disk, "/..". The data is omitted if the element is declared inside an HTML file.`

Looks like the property name (`containingLibraryName`) doesn't represent the property meaning anymore (`library URI`). Is it intentional?

If so, then we should first decide what we want to show in our Quick Doc popup. I believe the long bold `file:...` URI is not so interesting in this popup.

jwren:
Apparently my response to Devon agreeing with the two of you didn't post Friday.

Yes, I agree, the long `file:///C:/...` is to much and not the intention.  Working on the analysis server to change this, that is these will read 'web/main.dart'.  The goal here is to have these hover look and feel generally better.  In Java, the text shown is the name of the library, i.e. `java.lang`.  I am trying to get this field to provide the same functionality.  In the case of some local file in a library, a user imports it as the relative import.

> import `foo.dart`;

Landing this file as an incremental step would be fine and show history.  If it doesn't land, I'll rebase on this PR after the DAS change has landed in its' master branch.

alexander-doroshko:
Great! So with the server update, we won't need to have `isUnitTestMode()` in `DartDocUtil`, right? If so, I'd prefer to wait for the server change.

And just to clarify, it is intentional that `containingLibraryName` is not giving containing library name anymore but gives URI instead? Would you consider adding a different field to `HoverInformation` instead?

jwren:
Correct, I believe that we won't need the `isUnitTestMode()` any more.

Your next questions, correct.  The "URI" is how the containing library name is referenced - by other importing dart sources.  The goal of the `containingLibraryName` is to indicate to the user which library the source is referencing, the library name isn't sufficient for all cases and isn't a clear as simply providing the URI.  VS Code (until this change) actually went through the exercise in the client to produce the URI to import the element, from the library name.  Brian and I discussed changing the API, but given how subtle it is, opted to leaving the name as-is to reduce breaking changes.

Some additional context, note the "if you choose to name it" verbage, now even when libraries aren't named, users will have a clear understanding of which `import *` correlates to some class or field:
https://dart.dev/guides/language/effective-dart/style#do-name-libraries-and-source-files-using-lowercase_with_underscores

alexander-doroshko:
Ok, thanks for the detailed answer!
So, let's wait for the server changes that would allow this PR to get rid of `isUnitTestMode()` and then land this.

jwren:
PTAL.  Landed the change in the Dart SDK (https://dart-review.googlesource.com/c/sdk/+/104660), and reflected the changes back in this PR.

PR #665, opened 2019-06-01
Use completion.existingImports to filter completion suggestions

lambdabaa:
This CL processes `completion.existingImports` notifications and uses them to build and maintain a map from file onto import URI onto names exported from the import. Then, at completion time, `CompletionSuggestion.libraryFile` is used to lookup a map of all imported libraries and their declared names. Then only libraries which are already imported are allowed to suggest it if some library has already imported the name. This prevents the scenario where I've already imported 'package:foo/foo.dart' which provides `Foo` and not-yet-imported 'package:bar/bar' suggests `Foo` causing you to import a new package unnecessarily. 

Related to https://github.com/flutter/flutter-intellij/issues/3415.

/cc @alexander-doroshko @scheglov @jwren @devoncarew

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -204,8 +204,8 @@ public class DartAnalysisServerService implements Disposable {
     }
 
     @Override
-    public void computedExistingImports(@NotNull String file, @NotNull Map<String, Set<String>> uriToNames) {
-      myServerData.computedExistingImports(file, uriToNames);
+    public void computedExistingImports(@NotNull String file, @NotNull Map<String, Map<String, Set<String>>> existingImports) {
+      myServerData.computedExistingImports(file, existingImports);
     }
 
     @Override
@@ -814,7 +814,7 @@ public class DartAnalysisServerService implements Disposable {
   }
 
   @Nullable
-  public Map<String, Set<String>> getExistingImports(String file) {
+  public Map<String, Map<String, Set<String>>> getExistingImports(String file) {
     return myServerData.getExistingImports(file);
   }
 
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -302,7 +302,7 @@ public class DartAnalysisServerService implements Disposable {
                                    @NotNull final List<String> includedElementKinds,
                                    @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
                                    final boolean isLast,
-                                   String libraryFile) {
+                                   @Nullable String libraryFile) {
       synchronized (myCompletionInfos) {
         myCompletionInfos.add(new CompletionInfo(completionId, replacementOffset, replacementLength, completions, includedSuggestionSets,
                                                  includedElementKinds, includedSuggestionRelevanceTags, isLast, libraryFile));
@@ -2253,7 +2253,7 @@ public class DartAnalysisServerService implements Disposable {
     void consumeLibraryRef(@NotNull IncludedSuggestionSet includedSet,
                            @NotNull Set<String> includedKinds,
                            @NotNull Map<String, IncludedSuggestionRelevanceTag> includedRelevanceTags,
-                           String libraryFile);
+                           @Nullable String libraryFile);
   }
 
   private static class CompletionInfo {
@@ -2271,7 +2271,7 @@ public class DartAnalysisServerService implements Disposable {
     @NotNull private final List<String> myIncludedElementKinds;
     @NotNull private final List<IncludedSuggestionRelevanceTag> myIncludedSuggestionRelevanceTags;
     private final boolean isLast;
-    private final String myLibraryFile;
+    @Nullable private final String myLibraryFile;
 
     CompletionInfo(@NotNull final String completionId,
                    int replacementOffset,
@@ -2281,7 +2281,7 @@ public class DartAnalysisServerService implements Disposable {
                    @NotNull final List<String> includedElementKinds,
                    @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
                    boolean isLast,
-                   String libraryFile) {
+                   @Nullable String libraryFile) {
       this.myCompletionId = completionId;
       this.myOriginalReplacementOffset = replacementOffset;
       this.myOriginalReplacementLength = originalReplacementLength;
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -204,6 +204,11 @@ public class DartAnalysisServerService implements Disposable {
     }
 
     @Override
+    public void computedExistingImports(@NotNull String file, @NotNull Map<String, Set<String>> uriToNames) {
+      myServerData.computedExistingImports(file, uriToNames);
+    }
+
+    @Override
     public void computedErrors(@NotNull final String filePathSD, @NotNull final List<AnalysisError> errors) {
       final String fileName = PathUtil.getFileName(filePathSD);
 
@@ -296,10 +301,11 @@ public class DartAnalysisServerService implements Disposable {
                                    @NotNull final List<IncludedSuggestionSet> includedSuggestionSets,
                                    @NotNull final List<String> includedElementKinds,
                                    @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
-                                   final boolean isLast) {
+                                   final boolean isLast,
+                                   String libraryFile) {
       synchronized (myCompletionInfos) {
         myCompletionInfos.add(new CompletionInfo(completionId, replacementOffset, replacementLength, completions, includedSuggestionSets,
-                                                 includedElementKinds, includedSuggestionRelevanceTags, isLast));
+                                                 includedElementKinds, includedSuggestionRelevanceTags, isLast, libraryFile));
         myCompletionInfos.notifyAll();
       }
     }
@@ -490,7 +496,7 @@ public class DartAnalysisServerService implements Disposable {
             includedRelevanceTags.put(includedRelevanceTag.getTag(), includedRelevanceTag);
           }
           for (final IncludedSuggestionSet includedSet : completionInfo.myIncludedSuggestionSets) {
-            libraryRefConsumer.consumeLibraryRef(includedSet, includedKinds, includedRelevanceTags);
+            libraryRefConsumer.consumeLibraryRef(includedSet, includedKinds, includedRelevanceTags, completionInfo.myLibraryFile);
           }
           return;
         }
@@ -805,6 +811,11 @@ public class DartAnalysisServerService implements Disposable {
   @Nullable
   public AvailableSuggestionSet getAvailableSuggestionSet(int id) {
     return myServerData.getAvailableSuggestionSet(id);
+  }
+
+  @Nullable
+  public Map<String, Set<String>> getExistingImports(String file) {
+    return myServerData.getExistingImports(file);
   }
 
   @NotNull
@@ -2241,7 +2252,8 @@ public class DartAnalysisServerService implements Disposable {
   public interface CompletionLibraryRefConsumer {
     void consumeLibraryRef(@NotNull IncludedSuggestionSet includedSet,
                            @NotNull Set<String> includedKinds,
-                           @NotNull Map<String, IncludedSuggestionRelevanceTag> includedRelevanceTags);
+                           @NotNull Map<String, IncludedSuggestionRelevanceTag> includedRelevanceTags,
+                           String libraryFile);
   }
 
   private static class CompletionInfo {
@@ -2259,6 +2271,7 @@ public class DartAnalysisServerService implements Disposable {
     @NotNull private final List<String> myIncludedElementKinds;
     @NotNull private final List<IncludedSuggestionRelevanceTag> myIncludedSuggestionRelevanceTags;
     private final boolean isLast;
+    private final String myLibraryFile;
 
     CompletionInfo(@NotNull final String completionId,
                    int replacementOffset,
@@ -2267,7 +2280,8 @@ public class DartAnalysisServerService implements Disposable {
                    @NotNull final List<IncludedSuggestionSet> includedSuggestionSets,
                    @NotNull final List<String> includedElementKinds,
                    @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
-                   boolean isLast) {
+                   boolean isLast,
+                   String libraryFile) {
       this.myCompletionId = completionId;
       this.myOriginalReplacementOffset = replacementOffset;
       this.myOriginalReplacementLength = originalReplacementLength;
@@ -2276,6 +2290,7 @@ public class DartAnalysisServerService implements Disposable {
       this.myIncludedElementKinds = includedElementKinds;
       this.myIncludedSuggestionRelevanceTags = includedSuggestionRelevanceTags;
       this.isLast = isLast;
+      this.myLibraryFile = libraryFile;
     }
   }
 
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -153,12 +153,14 @@ public class DartServerCompletionContributor extends CompletionContributor {
                    }
 
                    Set<String> importedLibraries = new HashSet<>();
-                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
-                     String importedLibraryUri = entry.getKey();
-                     Map<String, Set<String>> importedLibrary = entry.getValue();
-                     Set<String> names = importedLibrary.get(suggestion.getDeclaringLibraryUri());
-                     if (names != null && names.contains(suggestion.getLabel())) {
-                       importedLibraries.add(importedLibraryUri);
+                   if (existingImports != null) {
+                     for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
+                       String importedLibraryUri = entry.getKey();
+                       Map<String, Set<String>> importedLibrary = entry.getValue();
+                       Set<String> names = importedLibrary.get(suggestion.getDeclaringLibraryUri());
+                       if (names != null && names.contains(suggestion.getLabel())) {
+                         importedLibraries.add(importedLibraryUri);
+                       }
                      }
                    }
 
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -145,21 +145,25 @@ public class DartServerCompletionContributor extends CompletionContributor {
                    return;
                  }
 
-                 final Map<String, Set<String>> uriToNames = das.getExistingImports(libraryFile);
+                 final Map<String, Map<String, Set<String>>> existingImports = das.getExistingImports(libraryFile);
                  for (AvailableSuggestion suggestion : suggestionSet.getItems()) {
                    final String kind = suggestion.getElement().getKind();
                    if (!includedKinds.contains(kind)) {
                      continue;
                    }
 
-                   Set<String> declaringLibraries = new HashSet<>();
-                   for (String libraryUri : uriToNames.keySet()) {
-                     if (uriToNames.get(libraryUri).contains(suggestion.getLabel())) {
-                       declaringLibraries.add(libraryUri);
+                   Set<String> importedLibraries = new HashSet<>();
+                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
+                     String importedLibraryUri = entry.getKey();
+                     Map<String, Set<String>> importedLibrary = entry.getValue();
+                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
+                     if (importedLibrary.entrySet().stream().anyMatch(
+                         importEntry -> doesImportSuggestion(importEntry, suggestionSet.getUri(), suggestion.getLabel()))) {
+                       importedLibraries.add(importedLibraryUri);
                      }
                    }
 
-                   if (!declaringLibraries.isEmpty() && !declaringLibraries.contains(suggestionSet.getUri())) {
+                   if (!importedLibraries.isEmpty() && !importedLibraries.contains(suggestionSet.getUri())) {
                      // If some library exports this label but the current suggestion set does not, we should filter.
                      return;
                    }
@@ -175,6 +179,11 @@ public class DartServerCompletionContributor extends CompletionContributor {
                });
              }
            });
+  }
+
+  // Checks whether the uri and name of a completion suggestion match those of an existing import.
+  private boolean doesImportSuggestion(Map.Entry<String, Set<String>> importedElements, String uri, String label) {
+    return importedElements.getKey().equals(uri) && importedElements.getValue().contains(label);
   }
 
   private static boolean isRightAfterBadIdentifier(@NotNull CharSequence text, int offset) {
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -156,9 +156,9 @@ public class DartServerCompletionContributor extends CompletionContributor {
                    for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
                      String importedLibraryUri = entry.getKey();
                      Map<String, Set<String>> importedLibrary = entry.getValue();
-                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
-                     if (importedLibrary.entrySet().stream().anyMatch(
-                         importEntry -> doesImportSuggestion(importEntry, suggestionSet.getUri(), suggestion.getLabel()))) {
+                     // TODO: Use individual suggestion uris instead of suggestion set uri.
+                     Set<String> names = importedLibrary.get(suggestionSet.getUri());
+                     if (names != null && names.contains(suggestion.getLabel())) {
                        importedLibraries.add(importedLibraryUri);
                      }
                    }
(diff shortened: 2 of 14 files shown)


Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public void process(JsonObject response) throws Exception {
+    JsonObject paramsObject = response.get("params").getAsJsonObject();
+    String file = paramsObject.get("file").getAsString();
+    JsonObject existingImports = paramsObject.get("imports").getAsJsonObject();  // ExistingImports
+    Map<String, Set<String>> uriToNames = buildUriNamesMap(existingImports);
+    getListener().computedExistingImports(file, uriToNames);
+  }
+
+  private Map<String, Set<String>> buildUriNamesMap(JsonObject existingImports) {
```
scheglov: I think this is wrong.

This method should return something like `Map<importedLibraryUri, Map<declaringLibraryUri, declaredName>>`. What I see as a single `Map<declaringLibraryUri, declaredName>`, so we know *what* is imported into the target library, but not *how*. This is important for re-exports. For example both "material.dart" and "cupertino.dart" re-export `Widget`, and we need to know which suggestion set to use for `Widget`.

alexander-doroshko:
Please add a test to the PR.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+      JsonObject existingImport = item.getAsJsonObject();  // ExistingImport
+      existingImport.get("elements").getAsJsonArray().forEach(element -> {
+        int index = element.getAsInt();
+        String uri = strings.get(uris.get(index));
+        if (!uriToNames.containsKey(uri)) {
+          uriToNames.put(uri, new HashSet<>());
+        }
+
+        String name = strings.get(names.get(index));
+        uriToNames.get(uri).add(name);
```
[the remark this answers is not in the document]
lambdabaa: Sounds good, done!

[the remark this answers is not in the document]
lambdabaa: Will do, thanks!

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private Map<String, Set<String>> buildUriNamesMap(JsonObject existingImports) {
+    JsonObject elements = existingImports.get("elements").getAsJsonObject();  // ImportedElementSet
+    List<String> strings = new ArrayList<>();
+    elements.getAsJsonArray("strings").forEach(item -> strings.add(item.getAsString()));
+    List<Integer> uris = new ArrayList<>();
+    elements.getAsJsonArray("uris").forEach(item -> uris.add(item.getAsInt()));
+    List<Integer> names = new ArrayList<>();
+    elements.getAsJsonArray("names").forEach(item -> names.add(item.getAsInt()));
```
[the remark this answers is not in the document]
lambdabaa: Done.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
lambdabaa: Thanks @scheglov! I got confused and missed that one library can re-export multiple names.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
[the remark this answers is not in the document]
lambdabaa: Then we would need to use hashmaps instead of lists for constant lookup though, right?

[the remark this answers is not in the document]
scheglov: No, but yes, in some sense :-)
`List<String> uris` maps `int` index to `String` value.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -141,18 +139,34 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
                  for (AvailableSuggestion suggestion : suggestionSet.getItems()) {
                    final String kind = suggestion.getElement().getKind();
                    if (!includedKinds.contains(kind)) {
                      continue;
                    }
 
+                   Set<String> declaringLibraries = new HashSet<>();
+                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
+                     String importedLibraryUri = entry.getKey();
+                     Map<String, Set<String>> importedLibrary = entry.getValue();
```
scheglov: Probably better name it `importedElements`.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -141,18 +139,34 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
+                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
+                     if (importedLibrary.values().stream().anyMatch(names -> names.contains(suggestion.getLabel()))) {
+                       declaringLibraries.add(importedLibraryUri);
```
scheglov: `declaringLibraries` -> `importedLibraries` like "libraries which import provide the suggestion".

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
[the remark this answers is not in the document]
lambdabaa: But then if we get a String from the uris or names lookup we have to use it as an index into strings?

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -141,18 +139,34 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
                    if (!includedKinds.contains(kind)) {
                      continue;
                    }
 
+                   Set<String> declaringLibraries = new HashSet<>();
+                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
+                     String importedLibraryUri = entry.getKey();
+                     Map<String, Set<String>> importedLibrary = entry.getValue();
+                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
+                     if (importedLibrary.values().stream().anyMatch(names -> names.contains(suggestion.getLabel()))) {
```
scheglov: This is not precise.
We need to use the URI and the name of the suggestion, something like `importedLibrary[suggestion.declaringLibraryUri][suggestion.name]`. However it seems that we don't have URI in `AvailableSuggestion`. So, we might have to extend `AvailableSuggestionSet` to provide URIs.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/NotificationCompletionExistingImportsProcessor.java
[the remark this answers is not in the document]
lambdabaa: Okay, discussed offline and fixed :)

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Done.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Done.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Good catch! We do have uri on AvailableSuggestionSet (not on AvailableSuggestion) so I used that.

lambdabaa:
@alexander-doroshko I wrote a unit test but wasn't able to run it since my intellij-community is broken after updating. If you can check it as part of merging, I would appreciate the help!

alexander-doroshko:
> intellij-community is broken after updating

What kind of problems do you have? First of all, make sure that you use git pull/commit/push right from the IDE, not from the command line. IDE integration works transparently with multiple repositories and also, much fewer chances to commit red code or not-related files or to forget to commit related files, etc. There are no reasons not to use IntelliJ's Git integration.

As for the test, I guess you need to use `configureByFiles()` as you need a multi-file project setup.

alexander-doroshko:
@scheglov thank you for the review! Does the PR LGTY in its current state?

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -141,18 +139,35 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
                    if (!includedKinds.contains(kind)) {
                      continue;
                    }
 
+                   Set<String> importedLibraries = new HashSet<>();
+                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
+                     String importedLibraryUri = entry.getKey();
+                     Map<String, Set<String>> importedLibrary = entry.getValue();
+                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
+                     if (importedLibrary.entrySet().stream().anyMatch(
```
scheglov: I think this is very inefficient.
We don't use the fact that this is a map, and just iterate over entries, checking the key.
We should rewrite it to something like `importedElements[suggestion.uri].contains(suggestion.label)`.

This is especially scary because we do this for every suggestion.

And I need to add `uri` to `AvailableSuggestion`.

lambdabaa: Whoops, thanks @scheglov! Updated.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -156,9 +156,9 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
                    for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
                      String importedLibraryUri = entry.getKey();
                      Map<String, Set<String>> importedLibrary = entry.getValue();
-                     // Checks whether any of the libraries exported by this import declares the same name as this suggestion.
-                     if (importedLibrary.entrySet().stream().anyMatch(
-                         importEntry -> doesImportSuggestion(importEntry, suggestionSet.getUri(), suggestion.getLabel()))) {
+                     // TODO: Use individual suggestion uris instead of suggestion set uri.
```
scheglov: DAS does provide `AvailableSuggestion.declaringLibraryUri` with https://dart-review.googlesource.com/c/sdk/+/104808
Is there something that prevents us from using it in this PR?

lambdabaa: Nope, I just pushed this before you landed that change :). Updating now!

alexander-doroshko:
@lambdabaa, compilation broken. And please check that the test works.

alexander-doroshko:
@lambdabaa, compilation broken. And please check that the test works.

lambdabaa:
> @lambdabaa, compilation broken. And please check that the test works.

Sorry, fixed the compilation issue!

alexander-doroshko:
@lambdabaa I'm getting NPEs with SDK 2.3 or older. We still support SDK 1.12 as a minimal version.

lambdabaa:
> @lambdabaa I'm getting NPEs with SDK 2.3 or older. We still support SDK 1.12 as a minimal version.

@alexander-doroshko Thanks for the heads up! I went ahead and updated the logic that reads AvailableSuggestion and CompletionSuggestion to not assume the existence of the new declaringLibraryUri and libraryFile fields.

lambdabaa:
> And please check that the test works.

Something strange is happening in the test that I am not fully understanding. This works outside of the context of a test, but the ExistingImports notification is different in test mode (missing the symbols re-exported from ExistingImportLibrary.dart). Any ideas about how to fix it @alexander-doroshko?

alexander-doroshko:
> ExistingImports notification is different in test mode

Probably because there's only one file in the actual test project (see `doTest()` implementation).
See my [first comment](https://github.com/JetBrains/intellij-plugins/pull/665#issuecomment-498599480) about the test:

> As for the test, I guess you need to use configureByFiles() as you need a multi-file project setup.

lambdabaa:
> > ExistingImports notification is different in test mode
> 
> Probably because there's only one file in the actual test project (see `doTest()` implementation).
> See my [first comment](https://github.com/JetBrains/intellij-plugins/pull/665#issuecomment-498599480) about the test:
> 
> > As for the test, I guess you need to use configureByFiles() as you need a multi-file project setup.

Ah that was it, thanks @alexander-doroshko!

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -141,18 +139,34 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
 
+                 final Map<String, Map<String, Set<String>>> existingImports = das.getExistingImports(libraryFile);
                  for (AvailableSuggestion suggestion : suggestionSet.getItems()) {
                    final String kind = suggestion.getElement().getKind();
                    if (!includedKinds.contains(kind)) {
                      continue;
                    }
 
+                   Set<String> importedLibraries = new HashSet<>();
+                   for (Map.Entry<String, Map<String, Set<String>>> entry : existingImports.entrySet()) {
```
alexander-doroshko: @lambdabaa, just to check the environment: do you get a warning here?

image

And IntelliJ IDEA is completely right. You get NPEs with Dart SDK 1.23.1 or older.

alexander-doroshko:
I tested the feature with today's nightly SDK. Completing in an empty file: I noticed that we've lost some classes from `dart:async` from the completion list, for example, `SynchronousStreamController`. I guess it's because `dart:core` [exports 2 classes](https://github.com/dart-lang/sdk/blob/1231244edeeee39dd1d6aeec744159d6be3c0b2f/sdk/lib/core/core.dart#L173) from `dart:async`.
Will changing `return` to `continue` in [DartServerCompletionContributor](https://github.com/JetBrains/intellij-plugins/pull/665/files#diff-b788dc0bc2f2b9f218821dc4d4802bfaR167) be a correct fix?

BTW, it sounds like a good scenario for a one more test!

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -296,10 +301,11 @@ public void computedCompletion(@NotNull final String completionId,
                                    @NotNull final List<IncludedSuggestionSet> includedSuggestionSets,
                                    @NotNull final List<String> includedElementKinds,
                                    @NotNull final List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
-                                   final boolean isLast) {
+                                   final boolean isLast,
+                                   String libraryFile) {
```
alexander-doroshko: We try to use `@Nullable`/`@NotNull` where possible for arguments and return types. Could you add `@Nullable` here and everywhere where this value is propagated?

alexander-doroshko:
`DartServerCompletionTest#testFunctionAfterShow` falis for me with this PR applied (against today's nightly SDK). Console contains exception:
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.lambda$buildUriNamesMap$4(NotificationCompletionExistingImportsProcessor.java:39)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.buildUriNamesMap(NotificationCompletionExistingImportsProcessor.java:37)
	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.process(NotificationCompletionExistingImportsProcessor.java:22)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processNotification(RemoteAnalysisServerImpl.java:675)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processResponse(RemoteAnalysisServerImpl.java:707)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.access$600(RemoteAnalysisServerImpl.java:48)
	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl$ServerResponseReaderThread.run(RemoteAnalysisServerImpl.java:1127)
```
Do you get the same?

Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerCompletionTest.java
```diff
@@ -188,4 +188,15 @@ public void testNoCompletionAfterDigit() {
   public void testNotYetImportedClass() {
     doTest("Process");
   }
+
+  public void testExistingImports() {
```
alexander-doroshko: This test doesn't fail for me with the new code commented out or against older SDKs. Apparently, it doesn't test the feature.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
lambdabaa: Okay I think I hit all of them!

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
lambdabaa: Yes, I see it now. I went ahead and added a safety check for null existingImports. Thanks!

Dart/testSrc/com/jetbrains/dart/analysisServer/DartServerCompletionTest.java
lambdabaa: Good point. We really need to be checking that one (and only one) library offers the completion (the already-imported one). I went ahead and updated the test.

lambdabaa:
Potentially blocked on https://github.com/dart-lang/sdk/issues/37242

lambdabaa:
> `DartServerCompletionTest#testFunctionAfterShow` falis for me with this PR applied (against today's nightly SDK). Console contains exception:
> 
> ```
> java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
> 	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
> 	at java.util.ArrayList.get(ArrayList.java:433)
> 	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.lambda$buildUriNamesMap$4(NotificationCompletionExistingImportsProcessor.java:39)
> 	at java.lang.Iterable.forEach(Iterable.java:75)
> 	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.buildUriNamesMap(NotificationCompletionExistingImportsProcessor.java:37)
> 	at com.google.dart.server.internal.remote.processor.NotificationCompletionExistingImportsProcessor.process(NotificationCompletionExistingImportsProcessor.java:22)
> 	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processNotification(RemoteAnalysisServerImpl.java:675)
> 	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.processResponse(RemoteAnalysisServerImpl.java:707)
> 	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl.access$600(RemoteAnalysisServerImpl.java:48)
> 	at com.google.dart.server.internal.remote.RemoteAnalysisServerImpl$ServerResponseReaderThread.run(RemoteAnalysisServerImpl.java:1127)
> ```
> 
> Do you get the same?

Fixed!

lambdabaa:
> Will changing `return` to `continue` in [DartServerCompletionContributor](https://github.com/JetBrains/intellij-plugins/pull/665/files#diff-b788dc0bc2f2b9f218821dc4d4802bfaR167) be a correct fix?

Yes that's correct, done!

lambdabaa:
@alexander-doroshko Gentle ping :)

alexander-doroshko:
Sorry for delay, I had a few days off. The PR is merged now, thanks! The commit will appear here shortly.

maRci002:
@lambdabaa Is there any option to disable this feature in the IDE?
I get strange completion ordering so I would like to fall back to old completion, in VS Code I am able to control it with "Auto Import Completions"

alexander-doroshko:
@maRci002, there's no option to disable the feature. I think we'd rather fix strange completion ordering than add an option. Can you share the details of the problem?

maRci002:
@alexander-doroshko thanks for quick answer.

My config:
- Android Studio 3.5 Build #AI-191.8026.42.35.5791312, built on August 9, 2019
- JRE: 1.8.0_202-release-1483-b03 amd64
- JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
- Windows 10 10.0
- Flutter plugin: v38.2.3
- Dart plugin: v191.8423
- IntelliJ version: according to [developer.android](https://developer.android.com/studio/releases#idea-2019-1) IntelliJ IDEA 2019.1

My problem is that when a constructor (let's say `new ListView.builder(...)`) waits for specific type of arguments then the IDE should consider type matching while using autocomplete feature.

Here's a simplified example of this issue, the padding argument shoud implemet EdgeInsetsGeometry abstract class but autocomplete shows me random methods / keywords...:
Android Studio autocomplete wrong

Autocomplete should know there are some classes which implement `EdgeInsetsGeometry `abstract class like: `class EdgeInsets extends EdgeInsetsGeometry` and should offer them as first results.

So the expected behaviour should be like this (this is old behaviour when `completion.existingImports` wasn't used):
Android Studio autocomplete

Moreover the problem exists in simple situations for example argument should be a `String`
Android Studio autocomplete wrong4

So to sum up, I think this feature significantly slows down development time because I don't want to manually check type hierarchy every single time. 3 months ago I made a [VS Code issue](https://github.com/Dart-Code/Dart-Code/issues/1742) and @DanTup figured out the problem occours when `completion.existingImports` event is used to search completion results, however in VS Code there is an option to ignore this feature via `"dart.autoImportCompletions": false` in settings.json.

I think I should have created a new Issue on [youtrack](https://youtrack.jetbrains.com/issues/WEB?q=Subsystem:%20Dart) but there wasn't option to create new account, and I am using Android Studio, but I think the problem occours in every Intellij based IDE which using the newest Dart plugin. (confirm?)

I really appreciate that you try to satisfy the [community's request](https://github.com/dart-lang/sdk/issues/25820), so keep it up 💪. I also hope you can fix this problem or at least give an option to disable this feature while working on fix.

DanTup:
Possibly we should move https://github.com/Dart-Code/Dart-Code/issues/1742 to the SDK repo. I do think the ranking is worse with this enabled, though I think it's a known trade-off (see https://github.com/Dart-Code/Dart-Code/issues/1742#issuecomment-496975928).

That said, VS Code ignores our sorting as soon as the user starts typing anyway (https://github.com/microsoft/vscode/issues/79516), so it's possible that even if fixed in the server it won't be perfect in VS Code because the chances of there being other items VS Code thinks are "more relevant" are increased with the unimported items 😞

maRci002:
I manually downgraded Dart plugin to [191.8026.36](https://plugins.jetbrains.com/plugin/6351-dart/update/65924), this is the last update which doesn't include this `completion.existingImports` feature, but the problem still exists. Maybe Android Studio mess up completion order so I should downgrade it to 3.4.2 which is based on ([IntelliJ IDEA 2018.3.4](https://developer.android.com/studio/releases#3-4-0)) and using Dart plugin 183.*.

maRci002:
I can confirm
- Android Studio 3.4.2 Build #AI-183.6156.11.34.5692245, built on June 27, 2019
- Flutter plugin: v38.2.1
- Dart plugin: v183.6270
Works fine for me.

So for example if I write `StreamController sc =` then I can manually `import 'dart:async';` or manually trigger `Quick Fix -> Import library 'dart:async'` and here is the magic: if I trigger autocomplete after assignment then IDE shows me `StreamController `and `StreamController.broadcast` contructors 😍
Android Studio autocomplete OLD

However in newer Android Studio 3.5 the manual import isn't required, but if I trigger autocomplete after assignment then the IDE shows me random stuffs even if I explicilty say I want to search after constructors with `new` keyword which makes me going crazy 
Android Studio autocomplete NEW

alexander-doroshko:
> a new Issue on youtrack but there wasn't option to create new account

There definitely is. You can log in using JetBrains account, as well as GitHub, Google account and more. But at the moment there's no need for YouTrack issue, I think we first need to continue the discussion with the Analysis Server team.

alexander-doroshko:
@scheglov, are there new ideas about improving type-aware completion sorting? Is there anything we are not doing right on the Dart plugin end?

maRci002:
> 1. DAS does not set `relevanceTags` for available declarations of constructors. https://dart-review.googlesource.com/c/sdk/+/114563 will fix this.

I would like to highlight 

> So for example if I write `StreamController sc` = and if I trigger autocomplete after assignment then IDE shows me StreamController and StreamController.broadcast contructors 😍

this is not true if I write `StreamController<int> sc = ` then autocomplete fails, however `StreamController<dynamic> sc = ` works fine.

scheglov:
With https://dart-review.googlesource.com/c/sdk/+/114969

image

image

maRci002:
Thanks @scheglov your work so far, today I was able to test this feature on flutter dev channel.

- Flutter (Channel dev, v1.10.5)
- Dart version 2.6.0 (build 2.6.0-dev.1.0 cb80ea7ba9)
- Android Studio 3.5
  - Flutter plugin: v39.0.3
  - Dart plugin: v191.8423

If I create a new empty dart file like you did a `test.dart`, then it works fine.
empty file

However I created a new project `flutter create my_test` with some library global variable and wrapped the `Column` widget with `Container/AspectRatio/Opacity` for illustration (ne need to create new project, just fill `test.dart` with some classes / global variables / some code). So if I trigger completion then relevance for global variables/classes are heigh.
global 1

When I trigger autocomplete in a local method then global variables/classes relevance is still boosted as well as local variables (`_counter / widget`)/methods (`toString()`). I like the way autocomplete shows `EdgeInsets.all( )` constructor for `padding` but shows only this one named constructor. Since EdgeInsetsGeometry is an abstract class maybe it shouldn't be boosted until there is no support for [anonymous classes/object literals](https://github.com/dart-lang/sdk/issues/3028).
local 1

PR #666, opened 2019-06-04
Invoke RequestListener.onRequest for every request, not just those that use LocalConsumer

lambdabaa:
We're seeing that `RequestListener.onRequest` is only being invoked for the `analysis.*` methods of analysis server. We should issue this at the beginning of the request across all requests though (e.g. code completion, fixes, etc.)


lambdabaa:
@alexander-doroshko If you wouldn't mind publishing a 2019.1.1 after this lands, we would very much appreciate it. Thanks!

PR #668, opened 2019-06-17
Dart hover improvement- indent parameters in a similar manner to the Java support

jwren:


alexander-doroshko:
Good idea!
Could you also handle nested generics? Looks like simple splitting on comma is not enough. Perhaps we need to write a custom splitter that counts angle brackets balance.
```
foo(Map<String, Map<String, String>> x) {}
bar(Map<String, Map<String, String>> x, y) {}
```
P.S. Now I don't care how many commits there are in a PR (now I have to take the whole PR as a single patch), so you don't need to force-push.

jwren:
Yes, will do.  I'll ping again after I do so.

jwren:
Take a look.  Nested generics now supported, test added.

alexander-doroshko:
Thanks, merged!

PR #670, opened 2019-06-20
Dart hover improvement in formatSignature, don't split around commas when in a Function parameter

jwren:


jwren:
Sorry for the multiple commits, the PR is rebased on top of master as I have access to it.  Any idea when it will be fixed?

Screenshot from 2019-06-20 09-59-37

alexander-doroshko:
I expect mirroring to be fixed soon. This PR will get auto-closed when fixed. Merged the 3rd commit from this PR.

PR #671, opened 2019-06-24
Implement inline rename refactoring for local Dart declarations

lambdabaa:
This change adds a new `DartVariableInplaceRenamer` which will take precedence over the dialog-based rename flow in the case of an `ElementKind.LOCAL_VARIABLE`. This is also how the [Python variable rename support](https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/inspections/quickfix/PyRenameElementQuickFix.java) works from what I can tell.

ezgif-1-75014536c0f5

/cc @alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartVariableInplaceRenamer.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartVariableInplaceRenamer.java
@@ -4,8 +4,6 @@ package com.jetbrains.lang.dart.ide.refactoring;
 import com.intellij.find.findUsages.FindUsagesHandler;
 import com.intellij.find.findUsages.FindUsagesOptions;
 import com.intellij.openapi.editor.Editor;
-import com.intellij.openapi.util.Pair;
-import com.intellij.openapi.util.TextRange;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiNamedElement;
 import com.intellij.psi.PsiReference;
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
Thanks, I'll review tomorrow. For now 2 blind comments:
- Make sure you use `editor.getSettings().isVariableInplaceRenameEnabled()` to switch between in-place or dialog-based rename
- How do we want to handle potential usages? With dialog-based rename user can select which potential usages to rename and which - not. Example:
```
main(a, b, Foo c) {
  a.toRename;
  b.toRename;
  c.toRename;
}

class Foo {
  var toRename;
}
```

lambdabaa:
> * Make sure you use `editor.getSettings().isVariableInplaceRenameEnabled()` to switch between in-place or dialog-based rename

Done!

> * How do we want to handle potential usages? With dialog-based rename user can select which potential usages to rename and which - not. Example:

As in Python, this treatment will only be applied in the case of local variables where it doesn't make sense for the user to choose some references to not update.

alexander-doroshko:
Thanks!
I think I have an idea how to make this even simpler. It might be quite easy to make standard `VariableInplaceRenamer` *just work* for Dart (just return `LocalSearchScope` for local vars in `DartPsiCompositeElementImpl#getUseScope`). I'll experiment with this a bit tomorrow. Stay tuned.

Dart/src/com/jetbrains/lang/dart/ide/refactoring/DartVariableInplaceRenamer.java
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }
+
+    return referenceProcessor.getResults();
+  }
+
+  @Override
+  protected PsiElement checkLocalScope() {
+    PsiElement element = myElementToRename;
+    while (element != null && element.getNode().getElementType().toString() != "FUNCTION_BODY") {
```
alexander-doroshko: Just FYI. 
Comparing with `DartTokenTypes.FUNCTION_BODY` (with `==`) is preferred to string comparison (which usually shouldn't be done with `==` or `!=` anyway :) ). 
But in this case `PsiTreeUtil.getParentOfType(myElementToRename, DartFunctionBody.class)` would be even better (because simpler).
No actions needed, just FYI.

lambdabaa: Thanks!

alexander-doroshko:
I'm going to close this PR as I've pushed the change that does pretty much the same but a bit simpler. GitHub synchronization is still not fixed but should get fixed soon. I hope you'll see all the recent commits shortly.

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/adf2c3a785a96005a43414a26fc1d293c1f7532c#diff-35900d987289a83af101d18a9f7c3501

PR #672, opened 2019-06-25
Implement WEB-39577- folding of DartAssertStatements

jwren:
https://youtrack.jetbrains.com/issue/WEB-39577

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
+++ b/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
@@ -56,7 +56,8 @@ public class DartFoldingBuilder extends CustomFoldingBuilder implements DumbAwar
         DartStringLiteralExpression.class,
         DartSetOrMapLiteralExpression.class,
         DartNewExpression.class,
-        DartCallExpression.class});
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
Just curious, is it just to behave similarly to foldings produced by Analysis Server? Do people write lots of multi-line asserts where they could benefit from folding?

jwren:
Yes, a user reached out after having to review & read through an assert that was 20 lines.

devoncarew:
Multi-line asserts are a common pattern in the Flutter framework code:

https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/scaffold.dart#L1553

alexander-doroshko:
@jwren @devoncarew What about folding only multiline asserts but keep single-line untouched? This is how we handle map/set literals, for example.

devoncarew:
👍 sounds reasonable to me

jwren:
SGTM, I'll take a look today.

jwren:
Nevermind, you took care of it already with https://github.com/JetBrains/intellij-plugins/commit/7da3d957f7b0b58970ad2d38ae1618410e58d7cf 

Thanks Alex

PR #673, opened 2019-07-01
adjust the size of the 'copy' button in the dart remote debug config

devoncarew:
- adjust the size of the 'copy' button in the dart remote debug config

Screen Shot 2019-07-01 at 9 19 46 AM

@alexander-doroshko


alexander-doroshko:
Is it to fix the layout on HiDPI monitor? Do you have a screenshot showing the problem without this change?

devoncarew:
Here's without that change:

Screen Shot 2019-07-01 at 11 35 21 AM

I recently made a similar change in the flutter plugin, and do see this constant in use in the intellij community source base.

alexander-doroshko:
Thanks, will merge.

PR #674, opened 2019-07-01
address a todo related to dart error annotations

devoncarew:
- address a todo related to dart error annotations
- use error codes to differentiate some error types, rather than the error message

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/annotator/DartAnnotator.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/annotator/DartAnnotator.java
@@ -263,7 +263,7 @@ public class DartAnnotator implements Annotator {
 
     final String severity = error.getSeverity();
     final String message = error.getMessage();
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Maybe fallback to old logic if there's no error code? (We still support SDK 1.12+)

devoncarew:
I can't find a changelog with the analysis server spec, but assume that `AnalysisError` has always had a `code` field?

From some brief investigation, it's had that field at least since when a portion of the spec was split off into a separate file (https://github.com/dart-lang/sdk/commit/2cbcd963274fc77f599731e089b6ec9c55e087c8) - around the sdk 1.24 timeframe.

alexander-doroshko:
Probably since SDK 1.16. See comments in https://github.com/JetBrains/intellij-plugins/pull/399.

alexander-doroshko:
Pushed with fallback logic for older SDKs. Also, I updated a test because this commit has changed highlighting for unused symbols in a `show` list of an import statement.

PR #675, opened 2019-07-09
WEB-14127- add the “Go To Test” feature for Dart

jwren:
@alexander-doroshko  This extension point should either have a language attribute attached in the XML, or an additional required method to override for the file type so the Java provider isn't called for Dart, and vise versa.

### DIFF
--- a/Dart/resources/META-INF/plugin.xml
+++ b/Dart/resources/META-INF/plugin.xml
@@ -103,6 +103,7 @@
                                     implementationClass="com.jetbrains.lang.dart.ide.marker.DartServerOverrideMarkerProvider"/>
 
     <codeInsight.gotoSuper language="Dart" implementationClass="com.jetbrains.lang.dart.ide.actions.DartServerGotoSuperHandler"/>
+    <testFinder implementation="com.jetbrains.lang.dart.ide.testIntegration.DartTestFinder"/>
     <definitionsScopedSearch implementation="com.jetbrains.lang.dart.ide.actions.DartInheritorsSearcher"/>
 
     <codeInsight.overrideMethod language="Dart"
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
@jwren LGTM. I'm ready to merge it as is tomorrow.

Alternatively, I think it's a good idea to make a few more steps and respect not only immediate children of `lib` and `test` folders. I looked through some public repositories (like https://github.com/dart-lang/sdk/) and noticed that it's quite common to have pairs like `lib/src/dart/element/builder.dart <-> test/dart/element/builder_test.dart`. I'd suggest that we use `FilenameIndex` and search for files that match name pattern (`x.dart <-> x_test.dart`), ignoring relative path to `lib` or `test`. 

Also, maybe support [all possible source file locations](https://dart.dev/tools/pub/package-layout): `benchmark`, `bin`, `lib`, `tool` and `web`?

P.S. Idea for test improvement: it could be a functional test rather than unittest if it is transformed to call `new DartTestFinder().findTestsForClass()` and `new DartTestFinder().findClassesForTest()`. `TestFinderHelper` may help.

jwren:
Ping.  Any changes you want before landing? I was hoping to have this land before continuing to add the above mentioned support.

alexander-doroshko:
No, sorry, I'm not waiting for changes. I just thought that if we decide to go with `FilenameIndex`-based solution then we may skip this PR as there probably will be little common code. I think I'll have time for it this week unless you beat me to it :).

alexander-doroshko:
@jwren I've pushed `FilenameIndex`-based implementation, please review: https://github.com/JetBrains/intellij-plugins/commit/b3097f865b993fa243eabd6f71d050565714ac42

jwren:
LGTM, thanks @alexander-doroshko

PR #676, opened 2019-07-09
Additional Dart parameter info tests, and two related bug fixes

jwren:


alexander-doroshko:
Thanks for the improvements in this area! Merged.
Just in case you want to iterate further in this area, the biggest problem I know is about a different order of passed named parameters.
```
foo() {
  bar(1, 2, y4: 4, y2: 2);
}
bar(int x1, int x2, {int y1 = 239, int y2 = 239, int y3 = 239, int y4 = 239}) {}
```
Kotlin has similar feature, we may look at their solution for inspiration.

jwren:
Yes, I will go follow up.  I believe I put in a failing test to this PR to this effect.

PR #678, opened 2019-07-18
Completion of constructors should insert parentheses

lambdabaa:
See https://github.com/dart-lang/sdk/issues/37563. We were using a different test (see https://github.com/JetBrains/intellij-plugins/pull/678/files#diff-b788dc0bc2f2b9f218821dc4d4802bfaR443) for whether `handleFunctionInvocationInsertion` should be applied in normal vs client-cached suggestions that made it so that constructors didn't trigger it.

/cc @alexander-doroshko  @bwilkerson @scheglov

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -471,10 +471,7 @@ public class DartServerCompletionContributor extends CompletionContributor {
         }
 
         if (element != null &&
-            (element.getKind().equals(ElementKind.FUNCTION) ||
-             element.getKind().equals(ElementKind.FUNCTION_INVOCATION) ||
-             element.getKind().equals(ElementKind.CONSTRUCTOR) ||
-             element.getKind().equals(ElementKind.CONSTRUCTOR_INVOCATION)) &&
+            (element.getKind().equals(ElementKind.CONSTRUCTOR)|| element.getKind().equals(ElementKind.FUNCTION)) &&
             suggestion.getParameterNames() != null) {
           handleFunctionInvocationInsertion(context, item, suggestion);
         }
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -470,9 +470,8 @@ public class DartServerCompletionContributor extends CompletionContributor {
           return;
         }
 
-        ElementKind elementKind = element.getKind();
         if (element != null &&
-            (elementKind.equals(ElementKind.FUNCTION_INVOCATION) || elementKind.equals(ElementKind.CONSTRUCTOR_INVOCATION)) &&
+            (element.getKind().equals(ElementKind.FUNCTION_INVOCATION) || element.getKind().equals(ElementKind.CONSTRUCTOR_INVOCATION)) &&
             suggestion.getParameterNames() != null) {
           handleFunctionInvocationInsertion(context, item, suggestion);
         }
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -471,7 +471,10 @@ public class DartServerCompletionContributor extends CompletionContributor {
         }
 
         if (element != null &&
-            (element.getKind().equals(ElementKind.FUNCTION_INVOCATION) || element.getKind().equals(ElementKind.CONSTRUCTOR_INVOCATION)) &&
+            (element.getKind().equals(ElementKind.FUNCTION) ||
+             element.getKind().equals(ElementKind.FUNCTION_INVOCATION) ||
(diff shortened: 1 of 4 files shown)


alexander-doroshko:
@lambdabaa, I'm sorry, do you have a code sample that works better with this change?

As far as I can see, `suggestion.getKind()` is always `"UNKNOWN"` at this point, see https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java#L576

lambdabaa:
> As far as I can see, `suggestion.getKind()` is always `"UNKNOWN"` at this point, see https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java#L576

@alexander-doroshko You're totally right, I think what is really needed is simply an update to check for both of the invocation kinds (function and constructor).

alexander-doroshko:
@lambdabaa Hm, how do you manage to submit uncompilable PRs?

lambdabaa:
> @lambdabaa Hm, how do you manage to submit uncompilable PRs?

I'm not sure I understand the question? The patch that I see on GitHub compiles and runs for me.

bfifwVHuE1U

alexander-doroshko:
`element.getKind()` returns `String`, doesn't it?

image

lambdabaa:
You make a very good point and yet

Screenshot from 2019-07-23 10-12-06

lambdabaa:
I fixed the issue, although I am still a bit confused about my development environment. Sorry for the headache.

alexander-doroshko:
Notice yellow editor tab background of the `DartServerCompletionContributor.java` file on your screenshot. And its icon: Java file instead of class. And the IDE that is started from sources: no Dart plugin. 
Looks like you haven't added `Dart-community` module to the `intellij-community` project. See step 4 in [README](https://github.com/JetBrains/intellij-plugins/tree/master/Dart).

alexander-doroshko:
You don't have code completion and code highlighting in `DartServerCompletionContributor.java`. Suspicious, isn't it? :) And you must have been warned about editing out-of-project files. :)

lambdabaa:
> You don't have code completion and code highlighting in `DartServerCompletionContributor.java`. Suspicious, isn't it? :) And you must have been warned about editing out-of-project files. :)

Ahhh every time I throw away my local changes to update to master these settings get blown away...

devoncarew:
I use `git stash` for this. If we frequently have issues merging back in the `git stash pop`, we could write a small script that would get the latest and re-apply targeted diffs to the iml files.

lambdabaa:
> I use `git stash` for this. If we frequently have issues merging back in the `git stash pop`, we could write a small script that would get the latest and re-apply targeted diffs to the iml files.

My stashes always have conflicts :(. I would definitely appreciate a script that addressed this issue...

alexander-doroshko:
One more way (using the IDE git integration):
- right-click the changelist(s) in the Version Control tool window and select Shelve Changes
- main menu -> VCS -> Git -> Pull, select upstream and its master branch (repeat for all 4 repositories)
- unshelve changes (using context menu or drag-n-drop) in the Version Control tool window (Shelf tab)

lambdabaa:
@alexander-doroshko I think this is ready to go -- wdyt?

alexander-doroshko:
@lambdabaa What SDK did you use to test? Do you have a code sample that works better with this change? 

I'm afraid it still doesn't work for me. Even worse, it stops working for global functions. And it breaks 3 tests.

This is the current state of this PR that I tried:
image

lambdabaa:
> I'm afraid it still doesn't work for me. Even worse, it stops working for global functions. And it breaks 3 tests.

Sorry I think I didn't have the latest version with CONSTRUCTOR and FUNCTION element types included up. I pushed this change and also have screenshots of the newly working functionality.

Screenshot from 2019-07-25 09-30-43
Screenshot from 2019-07-25 09-31-11

alexander-doroshko:
Much better!
As far as I understand, `FUNCTION_INVOCATION` and `CONSTRUCTOR_INVOCATION` can't happen here, but that's minor.

One step left: pease add a test to `DartServerCompletionTest`.

alexander-doroshko:
Does the added test pass for you? I don't think so :)

Nit: make sure to format all the edited lines in the source code. You may do it manually (Cmd+Alt+L for selection or for the whole file). Also, you may have 'Reformat Code before commit' checked in the commit dialog (or under gear in the Version Control tool window if Commit Dialog is disabled).

alexander-doroshko:
Something goes not great with this PR. To save our time and get future PRs quickly reviewed and merged, here's a checklist before submitting a PR:
- code must compile
- edited lines of code should be formatted by the IDE, imports in touched files should be optimized (no worries, formatter optimizes imports by default)
- no new warnings should be introduced in the edited files (unless there are strong reasons)
- the fix must be thoroughly tested manually, including all corner cases that you can imagine (like different SDK versions, different usage scenarios, etc, everything that makes sense for the particular fix)
- PR must not break existing tests that pass without this PR. If some existing test is wrong, it should be updated to stay passing.
- If possible and makes sense, PR should include a test for the problem it fixes. It could be a new test or (sometimes better) an existing test updated to include a new use case. Make sure that the test fails without the PR.

lambdabaa:
> Does the added test pass for you? I don't think so :)

Sorry about that, it's passing now!

alexander-doroshko:
Merged

PR #680, opened 2019-07-27
[Vue] Add "Expand v-model" intention

panstromek:
vmodelexpandfixed

First time trying to contribute to IntelliJ plugin, so I chose a little intention I want to use sometimes. Let me know how should I polish this ;), it's probably not aligned with all conventions.

I also don't know if 192 is the right target branch - it seems that it is the most active branch for Vue plugin, so I forked that instead of master.
 
I already signed Contributors agreement.

### DIFF
--- a/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
+++ b/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
@@ -11,7 +11,7 @@ import com.intellij.psi.xml.XmlElementType
 import org.jetbrains.vuejs.codeInsight.tags.VueElementDescriptor
 
 class VueExpandVModelIntention : JavaScriptIntention() {
-    override fun getFamilyName(): String = "Expand v-model into :value and @input"
+    override fun getFamilyName(): String = "Expand v-model into :value and @input/@change"
     override fun getText(): String = this.familyName
     private val validModifiers = setOf("lazy", "number", "trim")
 
@@ -44,12 +44,19 @@ class VueExpandVModelIntention : JavaScriptIntention() {
         val isComponent = parentTag.descriptor is VueElementDescriptor
         var assignedValue = if (isComponent) "\$event" else "\$event.target.value"
         var eventName = "@input"
+        var nextNeedsParens = false
         getModifiers(directiveText).forEach {
-            when {
-                it === "trim" -> assignedValue = "$assignedValue.trim()"
-                it == "number" -> assignedValue =
-                        "isNaN(parseFloat($assignedValue)) ? $assignedValue : parseFloat($assignedValue)"
-                it == "lazy" -> eventName = "@change"
+            if (nextNeedsParens) {
+                assignedValue = "($assignedValue)"
+                nextNeedsParens = false
+            }
+            when (it) {
+                "trim" -> assignedValue = "$assignedValue.trim()"
+                "number" -> {
+                    assignedValue = "isNaN(parseFloat($assignedValue)) ? $assignedValue : parseFloat($assignedValue)"
+                    nextNeedsParens = true
+                }
+                "lazy" -> eventName = "@change"
             }
         }
         val modelVariableName = parentAttr.value
--- a/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
+++ b/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
@@ -8,63 +8,60 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
 import com.intellij.psi.xml.XmlAttribute
 import com.intellij.psi.xml.XmlElementType
+import org.jetbrains.vuejs.codeInsight.attributes.VueAttributeNameParser
+import org.jetbrains.vuejs.codeInsight.attributes.VueAttributeNameParser.*
+import org.jetbrains.vuejs.codeInsight.attributes.VueAttributeNameParser.VueAttributeKind.DIRECTIVE
 import org.jetbrains.vuejs.codeInsight.tags.VueElementDescriptor
+import org.jetbrains.vuejs.index.isVueContext
 
 class VueExpandVModelIntention : JavaScriptIntention() {
-    override fun getFamilyName(): String = "Expand v-model into :value and @input/@change"
+    override fun getFamilyName(): String = "Expand v-model"
     override fun getText(): String = this.familyName
     private val validModifiers = setOf("lazy", "number", "trim")
 
     override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
-        return org.jetbrains.vuejs.index.isVueContext(element) &&
-                element.node.elementType == XmlElementType.XML_NAME &&
+        return element.node.elementType == XmlElementType.XML_NAME &&
                 element.parent?.node?.elementType == XmlElementType.XML_ATTRIBUTE &&
-                isValid(element)
+                element.parent.isValid &&
+                (element.parent as XmlAttribute).parent.descriptor is VueElementDescriptor &&
+                isValidVModel(element.parent as XmlAttribute) &&
+                isVueContext(element)
     }
 
-    private fun isValid(element: PsiElement): Boolean {
-        val text = element.text
-        if (!text.startsWith("v-model")) {
-            return false
-        }
-        val modifiers = getModifiers(text)
-        return validModifiers.containsAll(modifiers)
-    }
-
-    private fun getModifiers(text: String): List<String> {
-        val split = text.split('.')
-        return split.subList(1, split.size)
+    private fun isValidVModel(attribute: XmlAttribute): Boolean {
+        val info = VueAttributeNameParser.parse((attribute.name), attribute.parent)
+        return info.kind === DIRECTIVE &&
+                (info as? VueDirectiveInfo)?.directiveKind == VueDirectiveKind.MODEL &&
+                validModifiers.containsAll(info.modifiers)
     }
 
     override fun invoke(project: Project, editor: Editor?, psiElement: PsiElement) {
         editor ?: return
         val parent: PsiElement = psiElement.parent
-        val parentAttr = parent as XmlAttribute
-        val parentTag = parentAttr.parent
-        val directiveText = parent.name
-        val isComponent = parentTag.descriptor is VueElementDescriptor
-        var assignedValue = if (isComponent) "\$event" else "\$event.target.value"
-        var eventName = "@input"
-        var nextNeedsParens = false
-        getModifiers(directiveText).forEach {
-            if (nextNeedsParens) {
-                assignedValue = "($assignedValue)"
-                nextNeedsParens = false
-            }
-            when (it) {
-                "trim" -> assignedValue = "$assignedValue.trim()"
-                "number" -> {
-                    assignedValue = "isNaN(parseFloat($assignedValue)) ? $assignedValue : parseFloat($assignedValue)"
-                    nextNeedsParens = true
-                }
-                "lazy" -> eventName = "@change"
-            }
+        val modelAttribute = parent as XmlAttribute
+        val componentTag = modelAttribute.parent
+        val componentDescriptor = componentTag.descriptor as? VueElementDescriptor ?: return
+
+        val model = componentDescriptor.getModel()
+        var event = model?.event ?: "input"
+        val prop = model?.prop ?: "value"
+        val info = VueAttributeNameParser.parse(parent.name, parent.parent)
+
+        val modifiers = info.modifiers
+        var eventValue = "\$event"
+        if (modifiers.contains("trim")) {
+            eventValue = "typeof $eventValue === 'string' ? $eventValue.trim() : $eventValue"
         }
-        val modelVariableName = parentAttr.value
+        if (modifiers.contains("number")) {
+            eventValue = "_n($eventValue)"
+        }
+        if (modifiers.contains("lazy")) {
+            event = "change"
+        }
         CommandProcessor.getInstance().executeCommand(project, {
             WriteAction.run<RuntimeException> {
-                parentAttr.name = ":value"
-                parentTag.setAttribute(eventName, "$modelVariableName = $assignedValue")
+                modelAttribute.name = ":$prop"
+                componentTag.setAttribute("@$event", "${modelAttribute.value} = $eventValue")
             }
         }, "Expand V-Model", "VueExpandVModel")
     }
--- a/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
+++ b/vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
@@ -1,0 +1,63 @@
+package org.jetbrains.vuejs.intentions.expandVModel
+
+import com.intellij.lang.javascript.intentions.JavaScriptIntention
+import com.intellij.openapi.application.WriteAction
+import com.intellij.openapi.command.CommandProcessor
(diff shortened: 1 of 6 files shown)


vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+    private val validModifiers = setOf("lazy", "number", "trim")
+
+    override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
+        return element.node.elementType == XmlElementType.XML_NAME &&
+                element.parent?.node?.elementType == XmlElementType.XML_ATTRIBUTE &&
+                isValid(element)
+    }
+
+    private fun isValid(element: PsiElement): Boolean {
+        val text = element.text
```
piotrtomiak: I think you could use `VueAttributeNameParser` here to extract information about the attribute. It parses modifiers and arguments. Please use `XmlAttribute#getName` instead of `getText` to get attribute name

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+        editor ?: return
+        val parent: PsiElement = psiElement.parent
+        val parentAttr = parent as XmlAttribute
+        val parentTag = parentAttr.parent
+        val directiveText = parent.name
+        val isComponent = parentTag.descriptor is VueElementDescriptor
+        var assignedValue = if (isComponent) "\$event" else "\$event.target.value"
+        var eventName = "@input"
+        var nextNeedsParens = false
+        getModifiers(directiveText).forEach {
```
piotrtomiak: Please use `AttributeNameParser` for extracting modifiers. Does the order of modifiers matter on `v-model`?

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+
+class VueExpandVModelIntention : JavaScriptIntention() {
+    override fun getFamilyName(): String = "Expand v-model into :value and @input/@change"
+    override fun getText(): String = this.familyName
+    private val validModifiers = setOf("lazy", "number", "trim")
+
+    override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
+        return element.node.elementType == XmlElementType.XML_NAME &&
+                element.parent?.node?.elementType == XmlElementType.XML_ATTRIBUTE &&
+                isValid(element)
```
piotrtomiak: Please add a check whether you are in a Vue context - `org.jetbrains.vuejs.index.isVueContext(PsiElement)`

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    override fun invoke(project: Project, editor: Editor?, psiElement: PsiElement) {
+        editor ?: return
+        val parent: PsiElement = psiElement.parent
+        val parentAttr = parent as XmlAttribute
+        val parentTag = parentAttr.parent
+        val directiveText = parent.name
+        val isComponent = parentTag.descriptor is VueElementDescriptor
+        var assignedValue = if (isComponent) "\$event" else "\$event.target.value"
+        var eventName = "@input"
```
piotrtomiak: I think the event name should be taken from the component's `model` property:
```
(parentTag.descriptor as VueElementDescriptor).getModel().event
```
You need to add the `getModel()` function similarly to `getProps()`, but I guess you have to take into account only the first source.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,70 @@
...
+                    assignedValue = "isNaN(parseFloat($assignedValue)) ? $assignedValue : parseFloat($assignedValue)"
+                    nextNeedsParens = true
+                }
+                "lazy" -> eventName = "@change"
+            }
+        }
+        val modelVariableName = parentAttr.value
+        CommandProcessor.getInstance().executeCommand(project, {
+            WriteAction.run<RuntimeException> {
+                parentAttr.name = ":value"
```
piotrtomiak: Same for `value` name, please take it from the component's `model` property.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: I thought about that, too. Will do ;)

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: yes, order does matter (even though in this case not too much)

panstromek:
@piotrtomiak Thanks for review, I will look at it ;)

One more problem I have is that project fails to compile because of missing or private methods in JS plugin - I suppose there is some newer version of it internally in JetBrains.. I had to comment out or inline that code to make it build and run (it was unrelated to this PR). Can I do something about it?

image

I had similar problems before. Dennis Ushakov told be that tests are disabled because there are some unpublished changes - but that was few months ago already ( here: https://youtrack.jetbrains.com/issue/WEB-37226 ).

I didn't even try to run tests this time, I'll try to add some and run.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: ` I guess you have to take into account only the first source.`  - what do you mean by that? There can be only one v-model on a component - unless there's a syntax error. In that case, User will still want to transform the model that is under the caret, which is not necessarily the one returned from the descriptor..

piotrtomiak: I meant the implementation for `getModel` function in `VueElementDescriptor`; it has `sources` field, which usually has only one element, but sometimes more. This ambiguity is present, because in same rare cases we are not sure which of the candidates is really used in the template.

piotrtomiak:
@panstromek This compilation problem is present, because you are on `192` branch, which is already ahead of the latest published build - 2019.2 . It seems that until the first 2019.2.1 EAP is published you should actually checkout `192.5728` branch. You can create PR for that branch as well. I am sorry for the inconvenience.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -16,7 +16,8 @@ class VueExpandVModelIntention : JavaScriptIntention() {
     private val validModifiers = setOf("lazy", "number", "trim")
 
     override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
-        return element.node.elementType == XmlElementType.XML_NAME &&
+        return org.jetbrains.vuejs.index.isVueContext(element) &&
```
piotrtomiak: The vue context check might be an expensive operation as it needs to hit the cache and possibly filesystem. The `isAvailable` method is run on every PsiElement in the file, so it is better to move the context either just before `isValid` call or maybe better after. I think it would be good to add an import for `isVueContext`.

panstromek:
I see..  I thought the same thing but I thought that it will be cheap because it is cached. I will reorder it ;) thx

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: .. Actually - after some investigation I found out it doesn't - and code generated by vue-compiler is a bit different for some cases, too. I will need to update the code.

panstromek: (order does matter only for events)

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: I reordered it as last

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: done ;)

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: I misunderstood the meaning of VueElementDescriptor at first. I implemented the method. Used first source but still left defaults (`@input`, `:value`). Those are 90% times the correct ones.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: Refactored it to not care about order.

panstromek:
@piotrtomiak I addressed your points (hopefully) 😉

After some investigation I decided to make this intention only available for Vue components for now, because Vue actually generates pretty complicated code for some native HTML elements. In those cases it doesn't really seem useful to expand it and I also didn't want to go into too complicated stuff for the first PR.

About the tests - I can't figure out how to run them. Gradle task `test` doesn't run any. Also all test files in the project miss syntax highlighting, which makes them almost impossible to write. Can you point me to how should I set it up?

btw. what is the convention to resolving review? Does author or reviewer mark it as resolved?

alexander-doroshko:
@panstromek thanks for the update! At the moment, Piotr is on vacation, please expect a slight delay with this PR.

panstromek:
@alexander-doroshko yup, he told me. No hurry ;) thx

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,68 @@
...
+                element.parent?.node?.elementType == XmlElementType.XML_ATTRIBUTE &&
+                element.parent.isValid &&
+                (element.parent as XmlAttribute).parent.descriptor is VueElementDescriptor &&
+                isValidVModel(element.parent as XmlAttribute) &&
+                isVueContext(element)
+    }
+
+    private fun isValidVModel(attribute: XmlAttribute): Boolean {
+        val info = VueAttributeNameParser.parse((attribute.name), attribute.parent)
+        return info.kind === DIRECTIVE &&
```
piotrtomiak: Minor: if you do the cast below, no need to check the kind.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,68 @@
...
+
+class VueExpandVModelIntention : JavaScriptIntention() {
+    override fun getFamilyName(): String = "Expand v-model"
+    override fun getText(): String = this.familyName
+    private val validModifiers = setOf("lazy", "number", "trim")
+
+    override fun isAvailable(project: Project, editor: Editor?, element: PsiElement): Boolean {
+        return element.node.elementType == XmlElementType.XML_NAME &&
+                element.parent?.node?.elementType == XmlElementType.XML_ATTRIBUTE &&
+                element.parent.isValid &&
```
piotrtomiak: I think you should add a check for the component descriptor here as well.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
```diff
@@ -0,0 +1,68 @@
...
+    private fun isValidVModel(attribute: XmlAttribute): Boolean {
+        val info = VueAttributeNameParser.parse((attribute.name), attribute.parent)
+        return info.kind === DIRECTIVE &&
+                (info as? VueDirectiveInfo)?.directiveKind == VueDirectiveKind.MODEL &&
+                validModifiers.containsAll(info.modifiers)
+    }
+
+    override fun invoke(project: Project, editor: Editor?, psiElement: PsiElement) {
+        editor ?: return
+        val parent: PsiElement = psiElement.parent
```
piotrtomiak: Minor: maybe one variable for attribute: `val attribute = psiElement.parent as? XmlAttribute ?: return` ?

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: My reasoning was that "kind" check is faster than cast, so better check that first.. but I can drop that if you think so..

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: Is it really needed? I don't use it anywhere.. How should I check it, in case?

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
panstromek: will do

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
piotrtomiak: You look for it in the `invoke` logic:
```
 val componentDescriptor = componentTag.descriptor as? VueElementDescriptor ?: return
```
And if the descriptor is missing nothing happens, such behavior might be confusing for users.

vuejs/src/org/jetbrains/vuejs/intentions/expandVModel/VueExpandVModelIntention.kt
piotrtomiak: That's an interesting question :) I think that if `info` is already loaded into local variable (i.e. registry) the `instanceof` check will be faster than accessing that variables field and comparing it to another variables field.

piotrtomiak:
Closing the PR due to lack of activity — feel free to create a new PR with the suggested fixes.

panstromek:
Yes, sorry. After the last round of reviews, the build broke again, and I didn't have enough enthusiasm to spend another half a day trying to fix it.

piotrtomiak:
@panstromek Understood. I've made a lot of improvements to the gradle setup. I'll see if I can rebase this change onto master and merge it with the proposed changes. Sounds ok for you?

panstromek:
Ok, that'd be cool. I guess it's not very much needed feeture, though, so probably not worth spending too much time on it if it's broken

piotrtomiak:
I've applied your patch, made some very minor changes and merged it. Good work, thanks :)

panstromek:
Nice ;) thanks

panstromek:
Can you point me to the commit? I can't find it..

piotrtomiak:
@panstromek - it takes some time for the commit to be published to GitHub, as this repository is a mirror of a directory in our internal repository.

PR #681, opened 2019-07-31
Follow up on Dart Parameter Info action, https://github.com/JetBrains/intellij-plugins/pull/676

jwren:
@alexander-doroshko 

I did not update the tests as both my Linux box and Mac book are having problems running any of the tests (I'll continue to iterate).  I thought it was close enough in implementation to get your thoughts.  I wasn't able to convinced myself that `functionDescription` will always be the same as `functionDescription2`, so I left the distinction (again, can't run tests.)

```
foo({String str1= '1', String str2= '2'}) { }

main() {
  foo(str2: '', str1: ''); // caret in each parameter, Parameter Info works even though str2 is before str1 :)
}
```

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/info/DartFunctionDescription.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/info/DartFunctionDescription.java
@@ -38,6 +38,11 @@ public class DartFunctionDescription {
     return myParameters;
   }
 
+  @NotNull
+  public DartOptionalParameterDescription[] getOptionalParameters() {
+    return myOptionalParameters;
+  }
+
   public String getParametersListPresentableText() {
     final StringBuilder result = new StringBuilder();
     for (DartParameterDescription parameterDescription : myParameters) {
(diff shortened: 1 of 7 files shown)


alexander-doroshko:
Ok, I'll have a look tomorrow.

Speaking of tests, deleting `intellij-community/test-system` and `intellij-community/test-config` folders sometimes helps.

jwren:
Deleting `intellij-community/config` did it. I'll update this PR with passing and updated tests.

Thanks @alexander-doroshko

jwren:
Rebased, updated and forced pushed.  The latest commit on this PR removes the Dart parameter info test number 25- the ParameterInfoContexts are mocked in the tests, meaning this new support is not testable with these tests.

alexander-doroshko:
Cool! Merged.
BTW what's the problem with writing tests for it?

P.S. (minor corner cases) maybe show nothing with bold in cases like this
image
or this (incorrect call) 
image

Current parameter may be highlighted incorrectly in case of incorrect code:
image

jwren:
Thanks for landing and testing Alex.  This bug has bothered me a long time, I am happy to have it fixed.  I don't think that we need to spend any more time on it for these corner cases.

As for the tests, to get the named parameters to work, the `DartFunctionDescription` is retrieved from the `UpdateParameterInfoContext` (change in `DartParameterInfoHandler`), the problem for the tests is that the `UpdateParameterInfoContext` is a mocked object that does not have a `DartFunctionDescription` reference.

PR #682, opened 2019-08-03
Allow contributing literal completion suggestions like integral values

lambdabaa:
Now that ML completion is landing in Dart, analysis server can suggest literals like string and numeric values. Here is an example of this behavior in Dart's VS Code support:

Screen Shot 2019-08-01 at 7 49 52 PM

Removing this check which ensured historically that we would only complete tokens with lexemes that begin with alphabetic characters will allow this to work in Dart IntelliJ as well.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -70,12 +70,6 @@ public class DartServerCompletionContributor extends CompletionContributor {
                                            @NotNull final CompletionResultSet originalResultSet) {
                final PsiFile originalFile = parameters.getOriginalFile();
                final Project project = originalFile.getProject();
-
-               if (originalResultSet.getPrefixMatcher().getPrefix().isEmpty() &&
-                   isRightAfterBadIdentifier(parameters.getEditor().getDocument().getImmutableCharSequence(), parameters.getOffset())) {
-                 return;
-               }
-
                final CompletionSorter sorter = createSorter(parameters, originalResultSet.getPrefixMatcher());
                final String uriPrefix = getPrefixIfCompletingUri(parameters);
                final CompletionResultSet resultSet = uriPrefix != null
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
`isRightAfterBadIdentifier()` check was there on purpose. It workarounds the Analysis Server issue: https://github.com/dart-lang/sdk/issues/34672. I've found it by following a link in the commit message that added `isRightAfterBadIdentifier()`. And there's a test for it, which fails with this PR applied.

alexander-doroshko:
BTW, would integer literals completion work with this PR applied? Sorry, I haven 't tested myself. But my guess is that with this PR applied, if you select `10` from the completion list, you'll end up with `110` in code. Am I right?

lambdabaa:
> BTW, would integer literals completion work with this PR applied? Sorry, I haven 't tested myself. But my guess is that with this PR applied, if you select `10` from the completion list, you'll end up with `110` in code. Am I right?

Yup, this is the behavior I see locally with the patch applied...

Before:

Screenshot from 2019-08-06 11-22-18

After:

Screenshot from 2019-08-06 11-22-46

lambdabaa:
> `isRightAfterBadIdentifier()` check was there on purpose. It workarounds the Analysis Server issue: [dart-lang/sdk#34672](https://github.com/dart-lang/sdk/issues/34672). I've found it by following a link in the commit message that added `isRightAfterBadIdentifier()`. And there's a test for it, which fails with this PR applied.

I think this issue is still valid. None of the literals are suggested by analysis server if I simply type rather than triggering completion via ctrl+space before beginning to type the number. And a bunch of other invalid stuff shows up.

alexander-doroshko:
BTW is there a way for me to test the feature?
I tried to emulate the feature by adding a fake `CompletionSuggestion` to the server response. And here's what I get:
complete_int
I guess we need to update prefix calculation to get this working.

At the same time, until dart-lang/sdk#34672 is fixed we can't remove a workaround for it.

lambdabaa:
> BTW is there a way for me to test the feature?

@alexander-doroshko Yup! If you download a bleeding edge build of analysis server from http://gsdview.appspot.com/dart-archive/channels/be/raw/latest/sdk/ you only need to update your Dart SDK path and set the Registry value `dart.server.additional.arguments` to `--completion-model=/path/to/sdk/model/lexeme`.

lambdabaa:
I'm not sure what difference there is between your synthetic `CompletionSuggestion` and the real one but I don't see the same issue.

alexander-doroshko:
Probably in your experiment you've invoked completion right after `MAX_BUFFER_SIZE = ` and typed `1` only when the completion list was already visible. It's a different scenario.
You'll get unexpected completion behavior if you invoke completion *after* partially typed integer value.

PR #684, opened 2019-08-06
WEB-40331 Dart 'test' and 'group' argument folding for package:test support

jwren:


lookfirst:
I ❤️ you @jwren

alexander-doroshko:
LGTM.
One thought: do we want test/group name to stay visible even when the implementation is folded? For example, we could fold only the second argument of a test()/group() call. Or include test/group name in placeholder text.

alexander-doroshko:
@jwren, WDYT about folding the 2nd argument only thus keeping the name visible (+fold only multiline implementations)?

jwren:
See line 395 in `DartFoldingBuilder.java`- this is already implemented.  Line 399 has the logic for only multi-line implementations.

LMK if I need to rebase the PR to master to make it easier to land.

alexander-doroshko:
Oops, my bad, I didn't notice the update. Sorry for such a delay. Merged.

PR #687, opened 2019-08-06
Insert parens in case of (CONSTRUCTOR|FUNCTION)_INVOCATION

lambdabaa:
This is a follow-up from #678 addressing https://github.com/dart-lang/sdk/issues/37593. Flutter widget invocations do indeed have a `CONSTRUCTOR_INVOCATION` so we need to make sure to call `handleFunctionInvocationInsertion` for this case.

/cc @alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -471,7 +471,10 @@ public class DartServerCompletionContributor extends CompletionContributor {
         }
 
         if (element != null &&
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Hm, the example from dart-lang/sdk#37593 works fine for me without this PR. I see that `ElementKind.CONSTRUCTOR` is returned in that case.
If I'm not mistaken, this is the code that generates `ElementKind` for `AvailableSuggestion.getElement().getKind()`: https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/lib/src/domains/completion/available_suggestions.dart#L131. And it doesn't have `xxx_INVOCATION` items. We may ask Konstantin if unsure.

PR #688, opened 2019-08-10
failureaccess-1.0.1.jar is available on https url

muminc:
failureaccess-1.0.1.jar is available on https://repo1.maven.org, change to https url


C-Otto:
Duplicate of #757

chashnikov:
I've fixed the problem by upgrading Error Prone to 2.3.4 where the problem with missing dependency on failureaccess artifact is fixed. It has a different [missing dependency](https://github.com/google/error-prone/issues/1445) though, but I included it manually.

PR #695, opened 2019-08-28
Automatically detect osmorc by bnd.bnd file

tyborgplusplus:
Actually OSGI facet will be automatically detected by Manifest.MF file. 
This pull request aims to detect it by bnd.bnd file.


trespasserw:
Merged, thanks!

PR #696, opened 2019-09-06
WEB-40944 Angular - fix Make public intention

lppedd:
This PR aim to fix the issue described in [WEB-40944](https://youtrack.jetbrains.com/issue/WEB-40944).  
It also fixes the un-availability of the intention inside Angular Templates for constructur properties.

There are still a couple of things which don't work as I would expect, e.g. 
 - `isAvailable` returns false for
  `{{ testField{caret} }}` or `{{ testFunction{caret}() }}`.
-  caret position after applying the intention from Template isn't always perfect
   `@Decorator t{caret}estFunction(): void {}`

### DIFF
--- a/AngularJS/src/org/angular2/inspections/AngularInaccessibleComponentMemberInAotModeInspection.java
+++ b/AngularJS/src/org/angular2/inspections/AngularInaccessibleComponentMemberInAotModeInspection.java
@@ -190,8 +190,12 @@ public class AngularInaccessibleComponentMemberInAotModeInspection extends Local
 
     @Nullable
     private PsiElement locateMemberToEdit(@NotNull PsiElement element) {
-      if (element.getParent() instanceof JSReferenceExpression) {
-        element = element.getParent();
+      element = element instanceof PsiWhiteSpace
+            ? element.getPrevSibling()
+            : element.getParent();
+
+      if (!(element instanceof JSReferenceExpression)) {
+        element = element.getPrevSibling();
       }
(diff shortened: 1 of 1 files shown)


lppedd:
The only thing is that I have troubles running tests, as some testing classes aren't there. e.g.
`com.intellij.lang.javascript.JSChangeSignatureTestBase`

piotrtomiak:
@lppedd Thanks for the contribution! I think your changes require tests to be run, so you need to be able to run them. I have a task on my plate to publish test projects for Angular and Vue.js - https://youtrack.jetbrains.com/issue/WEB-37959 and I will have a look at it ASAP, hopefully this week. Let's defer review till then.

> I was able to configure DCEVM for hotswapping (obviously under Oracle JDK)
Cool, I think the Core team is experimenting with DCEVM for JetBrains runtime, so it's possible we will have that feature by default when using standard setup.

lppedd:
@piotrtomiak that's fine for me. Let me know, thanks!

piotrtomiak:
@lppedd Getting tests to run smoothly turned out to be more complicated than I thought, so I wasn't able to finish it for 2019.3. I've applied your changes, did some cleanup and added tests. Your change has been pushed to master and should arrive on github soon :) Thank you very much for the contribution and sorry for taking so long to apply it!

lppedd:
@piotrtomiak no problem, and thanks for the update.
I'll see when I'll be able to complete the other inspection PR, but currently I'm full of Angular stuff to do :(

PR #698, opened 2019-09-13
Use the extension method icon in Dart code completions, that is used in the Structure view.

jwren:
Any chance @alexander-doroshko that we could get an extension method icon?, i.e. the class `/nodes/class.svg` icon but with an `E` instead of a `C`.


alexander-doroshko:
> class.svg icon but with an E instead of a C

It's already there, it's called `Enum` and we use it for enums :)

jwren:
Of course, and we happen to have Enums in Dart...

The anonymous class icon isn't great though... other ideas, perhaps a lower case `e`?  Is there precedence set by another language here?

alexander-doroshko:
We may ask for `(e)` or `(ex)` icon.
Speaking of existing ones, we may consider `AllIcons.Nodes.Include`:
image
Though, `AllIcons.Nodes.AnonymousClass` also doesn't look that bad to me.

jacob314:
Using a `+` to indicate an extension looks reasonable to me.

alexander-doroshko:
Ok, I've changed to `(+)` icon.

jwren:
Thanks Alex, PR LGTM

PR #699, opened 2019-09-17
WEB-41425 Angular - implement duplicate selector and property binding inspection

lppedd:
This PR implements an inspection that reports duplicate selectors and property bindings.
While they might work, it's better to have a visual error/warning.

Example
```
<my-comp
  foo-bar        // The selector foo-bar is already specified
  foo-bar        // The selector foo-bar is already specified
  foo="bar"      // The property binding for foo is already specified
  [foo]="'baz'"  // The property binding for foo is already specified
></my-comp>
```

The relevant issue is https://youtrack.jetbrains.com/issue/WEB-41425

### DIFF
--- a/AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
+++ b/AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
@@ -1,0 +1,164 @@
+// Copyright 2000-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package org.angular2.inspections;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.angular2.codeInsight.attributes.Angular2AttributeDescriptor;
+import org.angular2.entities.Angular2Directive;
+import org.angular2.lang.Angular2Bundle;
+import org.angular2.lang.html.parser.Angular2AttributeType;
+import org.angular2.lang.selector.Angular2DirectiveSimpleSelector;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.lang.javascript.psi.JSElement;
+import com.intellij.psi.xml.XmlAttribute;
+import com.intellij.psi.xml.XmlElement;
+import com.intellij.psi.xml.XmlTag;
+
+import one.util.streamex.StreamEx;
+
+public class AngularDuplicateAttributeInspection extends AngularHtmlLikeTemplateLocalInspectionTool {
+   @Override
+   protected void visitXmlTag(
+         @NotNull final ProblemsHolder problemsHolder,
+         @NotNull final XmlTag xmlTag) {
+      StreamEx.of(xmlTag.getAttributes())
+              .mapToEntry(XmlAttribute::getDescriptor, Function.identity())
+              .selectKeys(Angular2AttributeDescriptor.class)
+              .filterKeys(AngularDuplicateAttributeInspection::isInspectable)
+              .flatMapKeyValue(AngularDuplicateAttributeInspection::toDuplicateAttribute)
+              .collect(Collectors.groupingBy(duplicateAttribute -> duplicateAttribute.targetName))
+              .values()
+              .forEach(duplicateAttributes -> {
+                 if (duplicateAttributes.size() > 1) {
+                    registerProblems(problemsHolder, duplicateAttributes);
+                 }
+              });
+   }
+
+   private static boolean isInspectable(@NotNull final Angular2AttributeDescriptor descriptor) {
+      final Angular2AttributeType type = descriptor.getInfo().type;
+      return type == Angular2AttributeType.REGULAR
+            || type == Angular2AttributeType.PROPERTY_BINDING
+            || type == Angular2AttributeType.BANANA_BOX_BINDING;
+   }
+
+   private static Stream<DuplicateAttribute> toDuplicateAttribute(
+         @NotNull final Angular2AttributeDescriptor descriptor,
+         @NotNull final XmlAttribute xmlAttribute) {
+      final String attributeName = descriptor.getInfo().name;
+
+      // Check if the attribute name represents a Directive selector.
+      // If it represents a Directive (or more than one) then enhance the Stream
+      // with an item per each Directive name.
+      // If it matches none, then a single item will be returned, using the name of the attribute
+      return StreamEx.of(descriptor.getSourceDirectives())
+                     .filter(directive -> matchesAttributeName(directive, attributeName))
+                     .map(Angular2Directive::getName)
+                     .distinct()
+                     .ifEmpty(attributeName)
+                     .map(targetName -> {
+                        final boolean isBinding =
+                              descriptor.getDeclarations()
+                                        .stream()
+                                        .anyMatch(d -> d instanceof JSElement);
+                        return new DuplicateAttribute(
+                              isBinding,
+                              targetName,
+                              isBinding ? descriptor.getInfo().name : targetName,
+                              xmlAttribute.getNameElement()
+                        );
+                     });
+   }
+
+   private static boolean matchesAttributeName(
+         @NotNull final Angular2Directive directive,
+         @NotNull final String attributeName) {
+      return StreamEx.of(directive.getSelector().getSimpleSelectors())
+                     .flatCollection(Angular2DirectiveSimpleSelector::getAttrNames)
+                     .anyMatch(a -> a.equalsIgnoreCase(attributeName));
+   }
+
+   private static void registerProblems(
+         @NotNull final ProblemsHolder problemsHolder,
+         @NotNull final List<DuplicateAttribute> duplicateAttributes) {
+      final int size = duplicateAttributes.size();
+      final boolean uniform =
+            duplicateAttributes.stream().map(a -> a.isBinding).distinct().count() == 1;
+
+      // All bindings            -> problem on all of them
+      // All selectors           -> problem only for last one
+      // Selectors and bindings  -> problem on all but last
+      if (uniform) {
+         final DuplicateAttribute first = duplicateAttributes.get(0);
+         doRegisterProblems(
+               problemsHolder,
+               first.isBinding
+                     ? duplicateAttributes
+                     : duplicateAttributes.subList(size - 1, size),
+               null,
+               first.isBinding
+                     ? "angular.inspection.template.duplicate-binding"
+                     : "angular.inspection.template.redundant-selector"
+         );
+      } else {
+         final Collection<DuplicateAttribute> subList = duplicateAttributes.subList(0, size - 1);
+         doRegisterProblems(
+               problemsHolder,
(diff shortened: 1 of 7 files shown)


lppedd:
I thing I just forgot to consider is a Directive can have multiple selectors assigned. 
Probably it's better to show a message in the form of `The selector {selectorName} for Directive {className} is already specified`.

piotrtomiak:
@lppedd Thanks for this PR, it looks very solid. As far as the code itself is concerned I don't have any remarks. I am not sure it is the best solution to distinguish between selectors and property bindings though. After all a particular attribute can come from a selector and also be an input to a directive or component or a regular attribute. Here is my proposition for the messages:
```
<my-comp
  foo-bar        // Duplicate attribute foo-bar
  foo-bar        // Duplicate attribute foo-bar 

  foo="2"        // Duplicate attribute foo
  foo="bar"      // Duplicate attribute foo
  [foo]="'baz'"  // Duplicate binding to foo
 
  bar="bar"       // Duplicate binding to bar
  [bar]="'baz'"   // Duplicate attribute [bar]
  [bar]="'bag'"   // Duplicate attribute [bar]
  [(bar)]="fff"   // Duplicate binding to bar
></my-comp>
```
What do you think? In case of attributes with exactly the same name the regular inspection is used (does it always work like that?) first and in other cases the "Duplicate binding {0}". 

There are also some edge cases in case of regular HTML attributes, simply checking `key` is not enough; for instance:
```
<my-comp
    innerhtml="<div>"
    [innerHTML]="foo"
    [attr.innerHtml]="<foo>"
/>
```
Please note that if someone makes a selector on `innerHtml` attribute, the attribute will have `Angular2AttributeDescriptor`. I am not sure how to properly detect this without doing some map for standard HTML attributes and their property counterparts, so I guess if there is an standard HTML source in the list of descriptor sources it should be ignored for checking. In such a case:
```
<img src="foo" [src]="bar"/>
```
would still not yeld any inspection errors.

AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	private static void registerProblem(
+			@NotNull final ProblemsHolder problemsHolder,
+			@NotNull final Collection<Entry<Angular2AttributeDescriptor, XmlAttribute>> entries) {
+		for (final Entry<Angular2AttributeDescriptor, XmlAttribute> entry : entries) {
+			final Angular2AttributeDescriptor descriptor = entry.getKey();
+			final XmlAttribute attribute = entry.getValue();
+			final String messageKey =
+					descriptor.getDeclaration() instanceof Angular2DirectiveSelectorPsiElement
```
piotrtomiak: Descriptor can have multiple declarations.

lppedd:
> What do you think? In case of attributes with exactly the same name the regular inspection is used (does it always work like that?)

Imho the standard HTML inspection should be reserved for standard HTML attributes (I'd say it's even misleading having it for Angular ones as of now).  
In the end
```
bar="bar"
[bar]="'baz'"
[bar]="'bag'"
[(bar)]="fff"
```
are all bindings to a class property (see also [one time string initialization](https://angular.io/guide/template-syntax#one-time-string-initialization) for the first one).

-----

> a particular attribute can come from a selector and also be an input to a directive

Yeah this is an edge case, but being we are binding to a property, I'd consider it as a binding.
Also, differentiating between selectors and bindings is useful for such cases

```
@Directive({
   selector: '[foo-bar-one], [foo-bar-two]'
})
export class MyDirective {}

...

<my-comp
  foo-bar-one   // Duplicate selector for MyDirective
  foo-bar-two   // Duplicate selector for MyDirective
></my-comp>
```

What do you think? Sounds reasonable?

AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
lppedd: @piotrtomiak Thanks!
Wouldn't all of them be of type `Angular2DirectiveSelectorPsiElement` anyway? Or am I missing something?

piotrtomiak: @lppedd No, the attribute declaration can come from a selector, input/output or regular HTML element.

piotrtomiak:
>  Imho the standard HTML inspection should be reserved for standard HTML attributes

I'll need to think about this one and talk to others on the team :)

> Also, differentiating between selectors and bindings is useful for such cases

Hmm, the selectors might be duplicated for the particular directive, but what if there are other directives one needing `foo-bar-one` and the other `foo-bar-two`.

lppedd:
> but what if there are other directives one needing foo-bar-one and the other foo-bar-two.

How I got what you mean.  
If there are other Directive(s) which use the same names (`foo-bar-one` or `foo-bar-two`), than they're not really duplicates, and all would be applied to the element.
We should know which Directive those selectors refer to, so that

```
@Directive({ selector: '[foo-bar-one], [foo-bar-two]' })
export class MyDirective {}

@Directive({ selector: '[foo-bar-two]' })
export class MyDirectiveTwo {}

...

<my-comp
  foo-bar-one   // Duplicate selector for MyDirective
  foo-bar-two   // Duplicate selector for MyDirective
  foo-bar-two   // OK!
></my-comp>
```
We are free to choose which ones to highlight.

> I'll need to think about this one and talk to others on the team :)

That's ok! Let me know.

AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
lppedd: @piotrtomiak ahhh yes! That's right.

piotrtomiak:
@lppedd I just talked with my team mate. My proposition is to apply following rule: if a regular attribute works as a binding (i.e. one-time-binding), the we use `binding`, otherwise `attribute`. In case of bindings we use `binding` always.

If you insist on having some information about the origin of attribute, than the only solution I see is to have a message like: `Duplicate attribute foo-bar specified in selector`. I don't like this, as to be 100% consisted in some case we would have to do something like:
`Duplicate attribute foo-bar specified in a selector, defined as input to 3 directive and coming from standard HTML element specification` 🤣

As far as directive with multiple selectors is concerned; IMHO the purpose of the inspection is to highlight potential conflicts. I think that if there are two attributes matched by one directive selector and specified only there, there is no conflict at all, as directives are matched and applied only once to each tag. However, it might be worth to underline such situations, so I propose to have a message `Redundant match on {directive} selector` 

So to summarize. I would see following validation results:
```
// foo-bar  - attribute from selector
// foo-bar2 - attribute from the same selector as foo-bar
// foo-bar3 - attribute from the same selector as foo-bar and a binding
// foo - a binding and an attribute from selector
// bar - a binding
<my-comp
  foo-bar        // Duplicate attribute foo-bar
  foo-bar        // Duplicate attribute foo-bar 
 
  foo-bar2       // Redundant match on MyImportantDirective selector

  foo="2"        // Duplicate binding to foo
  foo="bar"      // Duplicate binding to foo
 
  bar="bar"       // Duplicate binding to bar
  [bar]="'baz'"   // Duplicate binding to bar
  [bar]="'bag'"   // Duplicate binding to bar
  [(bar)]="fff"   // Duplicate binding to bar
></my-comp>
<my-comp
   foo-bar2        // Redundant match on MyImportantDirective selector
   [foo-bar3]='12'
>
```
What do you think?

lppedd:
> Duplicate attribute foo-bar specified in a selector, defined as input to 3 directive and coming from standard HTML element specification

hahaha yeah that would be quite convulted!

-----

Ok so I'm trying to be a hundred percent sure on what you meant.
I've placed questions marked with `Q:` in the example below
```
<my-comp
  foo-bar        // Duplicate attribute foo-bar
  foo-bar        // Duplicate attribute foo-bar 
 
  foo-bar2       // Redundant match on MyImportantDirective selector 
                    Q: So this one is a duplicate in respect of above ^ foo-bar?

  foo="2"        // Duplicate binding to foo
  foo="bar"      // Duplicate binding to foo
 
  bar="bar"       // Duplicate binding to bar
  [bar]="'baz'"   // Duplicate binding to bar
  [bar]="'bag'"   // Duplicate binding to bar
  [(bar)]="fff"   // Duplicate binding to bar
></my-comp>
<my-comp
   foo-bar2        // Redundant match on MyImportantDirective selector
                      Q: And this one is a duplicate in respect to below [foo-bar3]='12'?
   [foo-bar3]='12'
>
```

Anyway, I think it's reasonable to be as "simple" as possible, and I agree with your proposal.
Thanks again ;)

piotrtomiak:
@lppedd Yes to both questions :)

>Anyway, I think it's reasonable to be as "simple" as possible, and I agree with your proposal.
Thanks again ;)

Awesome! I am looking forward for code changes :)

lppedd:
@piotrtomiak I pushed some code as preview so you can see how I'm doing it.
I think the code might be improved so if you have any suggestion it's very welcome 😄 

And I've got another doubt. Is this case correct?
```
<my-comp
   [foo-bar3]='12'   // Redundant match on MyImportantDirective selector
   foo-bar2
>
```

piotrtomiak:
> And I've got another doubt. Is this case correct?
> 
> ```
> <my-comp
>    [foo-bar3]='12' // Redundant match on MyImportantDirective selector
>    foo-bar2
> ></my-comp>
> ```

I don't think so, the binding is significant

AngularJS/test/org/angular2/inspections/data/template/duplicate-attribute.html
```diff
@@ -0,0 +1,6 @@
+<my-comp
```
piotrtomiak: Please use the full example from the PR thread. We need to ensure that all corner cases are covered.

AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
```diff
@@ -0,0 +1,164 @@
...
+   }
+
+   private static boolean isInspectable(@NotNull final Angular2AttributeDescriptor descriptor) {
+      final Angular2AttributeType type = descriptor.getInfo().type;
+      return type == Angular2AttributeType.REGULAR
+            || type == Angular2AttributeType.PROPERTY_BINDING
+            || type == Angular2AttributeType.BANANA_BOX_BINDING;
+   }
+
+   private static Stream<DuplicateAttribute> toDuplicateAttribute(
```
piotrtomiak: Logic is slightly convoluted here and it's hard for me to follow it.

I think it would be simpler if you handle it as two separate cases. One would be redundant selector, which might actually be worth moving to a separate inspection and the other is duplicate binding.

In case of selectors you could group attrs by directives, from which originates selector. For each directive with multiple attributes representing redundant message would appear on those attributes who have single source of origin (selector) and are not bindings.

lppedd:
@piotrtomiak yeah that's right. `[foo-bar3]='12'` always "wins" apparently.

AngularJS/src/org/angular2/inspections/AngularDuplicateAttributeInspection.java
lppedd: Sounds reasonable to split the inspection!

piotrtomiak:
Closing the PR due to lack of activity — feel free to create a new PR with the suggested fixes.

PR #700, opened 2019-09-18
Additional work toward the Dartfix analysis server protocol implemented

jwren:


Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/DartfixProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+    this.consumer = consumer;
+  }
+
+  public void process(JsonObject resultObject, RequestError requestError) {
+    if (resultObject != null) {
+      try {
+        List<DartFixSuggestion> suggestions = DartFixSuggestion.fromJsonArray(resultObject.get("suggestions").getAsJsonArray());
+        List<DartFixSuggestion> otherSuggestions = DartFixSuggestion.fromJsonArray(resultObject.get("otherSuggestions").getAsJsonArray());
+        boolean hasErrors = resultObject.get("hasErrors").getAsBoolean();
+        List<SourceFileEdit> edits = SourceFileEdit.fromJsonArray(resultObject.get("edits").getAsJsonArray());
```
alexander-doroshko: @jwren Would it be better if the server returned `SourceChange` instead of `List<SourceFileEdit>`?

alexander-doroshko:
Merged

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/processor/DartfixProcessor.java
jwren: This will ultimately be up to the dart server devs, when they figure out the UX for the migration.

bwilkerson: Maybe. The protocol was originally designed for use by the command-line dartfix tool, and there wasn't any interest in using the other fields so we didn't include them. But if other clients want to make use of any or all of those fields, then we can certainly change the protocol.

PR #701, opened 2019-09-18
WEB-41422 Angular - consider null type as any and thus as valid one-time binding

lppedd:
This PR fixes the issue described in https://youtrack.jetbrains.com/issue/WEB-41422.

### DIFF
--- a/AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
+++ b/AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
@@ -466,7 +466,7 @@ public class Angular2AttributeDescriptor extends BasicXmlAttributeDescriptor imp
     if (ONE_TIME_BINDING_EXCLUDES.contains(property.getName())) return false;
     if (property.isVirtual()) return true;
     JSType type = property.getType();
-    if (type == null) return false;
+    if (type == null) return true;
 
     Map<Angular2DirectiveProperty, Boolean> cache = CachedValuesManager.getCachedValue(property.getSourceElement(), () ->
       CachedValueProvider.Result.create(new ConcurrentHashMap<>(), PsiModificationTracker.MODIFICATION_COUNT));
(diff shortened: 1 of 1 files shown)


AngularJS/src/org/angular2/codeInsight/attributes/Angular2AttributeDescriptor.java
```diff
@@ -465,15 +465,27 @@ else if (types.size() > 1) {
   public static boolean isOneTimeBindingProperty(@NotNull Angular2DirectiveProperty property) {
     if (ONE_TIME_BINDING_EXCLUDES.contains(property.getName())) return false;
     if (property.isVirtual()) return true;
-    JSType type = property.getType();
-    if (type == null) return false;
```
piotrtomiak: I think that the only change needed here is to return `true` if the type is `null` to fix the issue. I recall that I purposefully returned `false` here, but I can't recall why 😃 Since `any` type is assignable to `string` this should indeed return `true` here.

lppedd: @lppedd ok! I choosed to manually assign a `JSAnyType` to be sure I didn't lose anything which come after.
I'll do it tonight when I arrive home. Thanks!

lppedd:
@piotrtomiak done. I still maintained the code formatting.
Fun fact is I spent like 15 minutes to find the problematic line of code, and two hours debugging trying to understand how to calculate the implicit `JSAnyType` from a `TypeScriptField`, only to discover it basically does what I did 🤣

piotrtomiak:
@lppedd Thank you for the contribution! I've allowed myself to merge the first commit containing actual change and skip the formatting changes (I know the code within the method is harder to read, but on the other hand it's easier to find particular method - less lines - and in large files I prefer to keep the code "compressed"). In a separate commit I've updated some tests to cover the new behavior (I still need to get the tests working outside of full IDEA project). Github repo state should get updated within an hour or so.

lppedd:
@piotrtomiak that's ok. In the end it was just a single word change :)

piotrtomiak:
Cool :) Thanks again for the contribution :)

piotrtomiak:
@lppedd GitHub will automatically close this PR once commit gets into repo :)

PR #702, opened 2019-09-19
Dart refactoring- verify SourceChange object returned in the ServerRefactoringDialog is not null before calling applySourceChange

jwren:
I have seen a few reports of nulls coming back from the DAS in this refactoring dialog only.


alexander-doroshko:
@jwren Any chance to reproduce? I want to make sure we won't shadow a bigger problem.

jwren:
Agreed.  It also occurs to me that we should guard against all such non-null assertions from refactoring.

What about guarding against the `null` and at the same time reporting the issue either via the log or as a dialog to the user?

alexander-doroshko:
Ok, I've pushed null check with some logging. Let's hope it'll help or someone will submit STR when something doesn't work.

PR #706, opened 2019-09-24
Cleanup and refactoring in DartAnalysisServerService.java

jwren:
- Introduce `logTookTooLongMessage` to have the `LOG.info`s printed out be more consistent and include versioning information
- Have all Dart Analysis Server queries consistently check `AnalysisServer server = myServer; if (server == null) { return ...; }` before work starts in the method, this makes the methods more consistent and is a small optimization
- Refactor any file path references to be named `filePath` to be consistent

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -1072,13 +1072,13 @@ public class DartAnalysisServerService implements Disposable {
 
   @NotNull
   public List<HoverInformation> analysis_getHover(@NotNull final VirtualFile file, final int _offset) {
-    final String filePath = FileUtil.toSystemDependentName(file.getPath());
-    final List<HoverInformation> result = Lists.newArrayList();
-
     final AnalysisServer server = myServer;
     if (server == null) {
       return HoverInformation.EMPTY_LIST;
     }
+
+    final List<HoverInformation> result = Lists.newArrayList();
+    final String filePath = FileUtil.toSystemDependentName(file.getPath());
 
     final CountDownLatch latch = new CountDownLatch(1);
     final int offset = getOriginalOffset(file, _offset);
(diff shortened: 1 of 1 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2244,6 +2314,23 @@ private String getShortErrorMessage(@NotNull String methodName, @Nullable String
...
+                                     @Nullable String filePath) {
+    StringBuilder builder = new StringBuilder();
+    builder.append(methodName).append("() took longer than ").append(timeout).append("ms");
+    if (contextID != null) {
+      builder.append(", for context id ").append(contextID);
+    }
+    if (filePath != null) {
+      builder.append(", for file ").append(filePath);
+    }
+    builder.append(", IntelliJ product name: ").append(getClientId()).append(", ").append(getClientVersion());
```
alexander-doroshko: IDE version is always in logs. I'll remove this line if you don't mind.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2244,6 +2314,23 @@ private String getShortErrorMessage(@NotNull String methodName, @Nullable String
            ", error code = " + error.getCode() + ": " + error.getMessage();
   }
 
+  private void logTookTooLongMessage(@NotNull final String methodName,
+                                     final long timeout,
+                                     @Nullable final String contextID,
```
alexander-doroshko: Looks like there's only one call where `contextID` is not null, and having it in log won't help much, so I removed this arg for the sake of simplicity.

alexander-doroshko:
merged

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: SGTM

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
jwren: I went back and forth, I didn't feel strongly either way.

jwren:
Thanks Alex

PR #712, opened 2019-10-04
Update the Dart VM API (vm_service) to the latest protocol, version 3.27.0

jwren:
This ensures that the Dart Plugin 2019.3.0 will have the latest recent Dart VM protocol.


jwren:
cc @alexander-doroshko @devoncarew @jacob314

PR #714, opened 2019-10-09
Add hook to allow subscribing for debug logs from analysis server

lambdabaa:


lambdabaa:
Update: Working to have this handle analysis server restarts.

alexander-doroshko:
There's already a universal `DartAnalysisServeService.sendRequest()` method, which is already used in `FlutterDartAnalysisServer.sendSubscriptions()`. Maybe that's enough for now?

lambdabaa:
> There's already a universal `DartAnalysisServeService.sendRequest()` method, which is already used in `FlutterDartAnalysisServer.sendSubscriptions()`. Maybe that's enough for now?

I think that could work if we didn't also need to make sure that we subscribe to this additional topic after analysis server restart. Please see my updated changes.

alexander-doroshko:
Thanks, merged. I've added a parameter to `setServerLogSubscription(boolean subscribeToLog)`, just to have more universal API.

PR #717, opened 2019-10-17
Update the Dart Web run configuration to have a working directory input

jwren:
This is not yet supported by webdev but gets ready for the additional support coming soon, see https://github.com/dart-lang/webdev/issues/744


alexander-doroshko:
Do we really need a separate field for the working directory (aka dart project root)?
We used to calculate it by HTML file path (`DartCommandLineRunnerParameters.suggestDartWorkingDir()`), is anything wrong with this approach?

Also, having the HTML file path, we can calculate the directory to serve and pass it to the `webdev` as an argument (currently `web` is hardcoded).

devoncarew:
> We used to calculate it by HTML file path (DartCommandLineRunnerParameters.suggestDartWorkingDir()), is anything wrong with this approach?

That sounds good to me - asking for the html file to launch, and using it to infer the serving directory.

jwren:
Sounds good to me.  I'll go review `suggestDartWorkingDir()` and update the command line generation to not hardcode the `web`.

PR #718, opened 2019-10-18
Update the Dart Web launch configuration with the new --launch-app input

jwren:
`webdev` change in this PR: https://github.com/dart-lang/webdev/pull/748

This PR also removes the hardcoded `web` in the construction of the `webdev` command line.


alexander-doroshko:
Thanks, merged. Now we need to update `DartWebdevParameters#check()` accordingly.

PR #721, opened 2019-10-24
Optimization with Dart Remote Debugger- don't always try to guess the project root

jwren:


alexander-doroshko:
Why no chance for pubspecless projects? `myCurrentWorkingDirectory` is manually set in the run configuration, so we have all the info to guess the project root correctly.
Also even if `"dart.projects.without.pubspec"` is set to `true`, I think these users may occasionally open standard pubspec-based Dart apps. I don't think we can win a lot with this optimization, but I think we can lose some functionality.

jwren:
Agreed, your point about not enough of an optimization for the potential loss of functionality is true.  Closing out this PR.

PR #725, opened 2019-10-29
Add Dart setting to control passing --enable-completion-model to analysis server

lambdabaa:
Screenshot from 2019-10-29 14-04-07

/cc @alexander-doroshko @devoncarew @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -254,8 +254,7 @@ analysis.server.settings.title=Dart Analysis Server Settings
 analysis.server.settings.sdk.settings=Dart plugin settings
 analysis.server.settings.diagnostics=View analyzer diagnostics
 analysis.server.settings.package.analysis=Scope analysis to the current package
-analysis.server.settings.ml.completion=ML Code Completion:
-analysis.server.settings.ml.completion.enable=Enable ML-based autocompletion
+analysis.server.settings.ml.completion=Enable machine learning code completion
 warning.old.sdk.ml.completion=Warning\: ML-based autocompletion only available starting at Dart SDK version 2.5.0.
 
 dart.sort.members.action.name=Sort Members in Dart File
--- a/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
+++ b/Dart/src/com/jetbrains/lang/dart/DartBundle.properties
@@ -254,6 +254,9 @@ analysis.server.settings.title=Dart Analysis Server Settings
 analysis.server.settings.sdk.settings=Dart plugin settings
 analysis.server.settings.diagnostics=View analyzer diagnostics
 analysis.server.settings.package.analysis=Scope analysis to the current package
+analysis.server.settings.ml.completion=ML Code Completion:
+analysis.server.settings.ml.completion.enable=Enable ML-based autocompletion
+warning.old.sdk.ml.completion=Warning\: ML-based autocompletion only available starting at Dart SDK version 2.5.0.
 
 dart.sort.members.action.name=Sort Members in Dart File
 dart.sort.members.action.name.ellipsis=Sort Members in Dart files...
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -71,6 +71,7 @@ import com.jetbrains.lang.dart.ide.actions.DartPubActionBase;
 import com.jetbrains.lang.dart.ide.errorTreeView.DartFeedbackBuilder;
 import com.jetbrains.lang.dart.ide.errorTreeView.DartProblemsView;
 import com.jetbrains.lang.dart.ide.template.postfix.DartPostfixTemplateProvider;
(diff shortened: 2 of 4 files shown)


Dart/src/com/jetbrains/lang/dart/sdk/DartConfigurable.java
```diff
@@ -410,6 +418,11 @@ private String getErrorMessage() {
       }
     }
 
+    String sdkVersion = DartSdk.getDartSdk(myProject).getVersion();
```
jwren: The Dart SDK version should be fetched from the dialog settings higher in the dialog itself so that if a higher version of the SDK is actively selected, users won't see the error message.

lambdabaa: Done, thanks!

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2045,6 +2046,9 @@ else if (!dasSnapshotFile.canRead()) {
       catch (MissingResourceException e) {
         // NOP
       }
+      if (DartConfigurable.getMLCodeCompletionEnabled(myProject)) {
```
devoncarew: I'm wondering how this will interact with the analysis server default for ML completion, if we at some point choose to change it. Should we pass in `--no-enable-completion-model` if the box is unchecked? Or, how would user indicate to disable ML completion if the analysis server later enables it by default?

lambdabaa: One idea to address this issue is to never enable it by default in analysis server and to instead enable it by default across clients. But I'm not sure if that's the best rollout strategy.

devoncarew: That's feasible, and avoid timing issues with rolling it out between the analysis server and clients. We may consider some other way to enable these flags in the future - to have a set of settings in the analysis server that we can query and modify.

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
```diff
@@ -254,6 +254,9 @@ analysis.server.settings.title=Dart Analysis Server Settings
 analysis.server.settings.sdk.settings=Dart plugin settings
 analysis.server.settings.diagnostics=View analyzer diagnostics
 analysis.server.settings.package.analysis=Scope analysis to the current package
+analysis.server.settings.ml.completion=ML Code Completion:
+analysis.server.settings.ml.completion.enable=Enable ML-based autocompletion
```
DanTup: As discussed in email - it'd make sense to have these labels match (or at least be similar) with VS Code. I was trying to avoid using just "ML" since it might not be clear to everyone what that is. Do you have any preferences? (the current text here also uses both "code completion" and "autocompletion", I'm not sure if they should be consistent?)

The current VS Code wording is here:

https://github.com/Dart-Code/Dart-Code/pull/2076

Though re-reading, I'd probably change "machine learning" to "machine learned", since if I understand correctly we're training the model and shipping it, rather than learning on the users machine?

(I don't feel at all strongly about this though, so I'm happy to update it to match whatever you go with here!)

alexander-doroshko:
I'm leaning towards the UI similar to what @DanTup has done (but with sentence capitalization).
image
A more detailed explanation of the checkbox will be in the IDE online help (accessible by F1).

Dart/src/com/jetbrains/lang/dart/DartBundle.properties
devoncarew: The name for this (at tech preview) was `ML Complete`. The abbreviation 'ML' is I think well enough understood now that we can use it in a label (through  tooltip text or help text may use the full text).

As bikeshedding for the labels, perhaps `'ML Completion:'` and `'Enabled ML based code completion'`, or, `'Enabled machine learning based code completion'`?

DanTup: I had an additional reason for not calling it "ML" in the title for VS Code... VS Code auto-generates the display labels from the settings name and would write "ML" as "Ml" 😔 (I've opened https://github.com/microsoft/vscode/issues/83663).

I'll copy whatever is done here though, with just that change (and any tweaks to make it consistent with how other settings appear in VS Code, such as being prefixed with "controls whether...").

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2045,6 +2046,11 @@ else if (!dasSnapshotFile.canRead()) {
       catch (MissingResourceException e) {
         // NOP
       }
+      String sdkVersion = DartSdk.getDartSdk(myProject).getVersion();
```
alexander-doroshko: IDE shows a warning in this line. I've reused `sdk` parameter available here instead.

alexander-doroshko:
@DanTup, finally, the checkbox label is going to be 
```
Enable machine learning based code completion
```

alexander-doroshko:
Pushed with minor updates. Github mirroring is under maintenance, the commit will appear here with some delay.

alexander-doroshko:
The most important addition made: restart the server on option change.

PR #727, opened 2019-10-30
Removed the Dart analyzed files subscription from the Dart plugin

jwren:


alexander-doroshko:
Pushed. Github mirroring is still under reconstruction (broken after a huge merge).
Curious, has this code ever been used? What was wrong with it and what is used now instead?

PR #728, opened 2019-10-31
Implement suggestedName() in DartWebdevConfiguration.java

jwren:


Dart/src/com/jetbrains/lang/dart/ide/runner/server/webdev/DartWebdevConfiguration.java
```diff
@@ -56,6 +59,28 @@ public RunProfileState getState(@NotNull final Executor executor,
...
+        htmlFilePath.startsWith(workingDir.getPath() + "/") ? htmlFilePath.substring(workingDir.getPath().length() + 1) : null;
+    }
+    catch (RuntimeConfigurationError ignore) {
+    }
+
+    if (StringUtil.isNotEmpty(htmlFilePathRelativeFromWorkingDir)) {
+      return htmlFilePathRelativeFromWorkingDir;
+    }
+
+    return StringUtil.isEmpty(htmlFilePath) ? null : PathUtil.getFileName(htmlFilePath);
```
alexander-doroshko: Returning not-null here enables some magic: the name of the manually created run configuration changes as user types in or selects an HTML file. But once the name is manually edited, it's not auto-updated anymore.
I've changed null to `"Dart web app"`.

alexander-doroshko:
Pushed.
We may think about the case when people have several dart projects within a single IDE project. They might want to create run configurations for `web/index.html` from different dart roots.
On the other hand, long generated run configuration names are also not great. We should think about the best balance for the most common scenarios.

PR #730, opened 2019-10-31
Additional logging in DartDocumentationProvider.java

jwren:
Additional logging in `DartDocumentationProvider.java` to try to get additional data for the `psiFile == null` exceptions


alexander-doroshko:
I'm not sure how to write a log message that would help us to understand and fix the exception. (And we still have zero reports in our exception tracker.) I see that some changes in `CtrlMouseHandler` have been done recently that might fix the issue. I suggest to wait for exception reports from 2019.3 and think once again if they will appear.

PR #731, opened 2019-11-04
Cache the client id and client version reported to the Dart Analysis Server

jwren:


alexander-doroshko:
I don't think we need to cache this. These methods are not used in loops and they don't contain heavy calculation. I've changed `replaceAll()` to `replace(char, char)` just not to use regexp where there's no need in it. Now both methods take less than a nanosecond, I think we are fine with this. This class already contains too many fields.

PR #733, opened 2019-11-05
Add Dart DiagnosticMessage action support in the Dart Analysis window

jwren:
This adds support for jumping to source locations indicated by Dart errors and warnings via the optional `List<DiagnosticMessage>` to see an example, include this function in a Dart file:

```
void f() {
  print(x);
  int x = 3;
}
```


alexander-doroshko:
Thanks, pushed.
Do you think we also want to expose this new information on hover? On the one hand, it's already loaded with information, on the other hand, it looks like a good place for diagnostics with a link to source. Current state:
image

jwren:
How would one make a `Location` navigable?  Also, this would violate your concern about the hover size even more.

alexander-doroshko:
>How would one make a Location navigable?

I thought about a link, just like `Open Documentation` but I'm not sure it's easy/possible to jump to source by such link.

> concern about the hover size even more

That's true. I think I want to consult our UX team regarding the tooltip rendering.

PR #734, opened 2019-11-05
Add Bazel support to DartTestFinder.java

jwren:


Dart/src/com/jetbrains/lang/dart/ide/testIntegration/DartTestFinder.java
```diff
@@ -32,14 +34,14 @@ public PsiElement findSourceElement(@NotNull final PsiElement from) {
   @NotNull
   @Override
   public Collection<PsiElement> findTestsForClass(@NotNull final PsiElement element) {
-    Project project = element.getProject();
-    VirtualFile vFile = element.getContainingFile().getVirtualFile();
-    String topLevelDirName = getTopLevelDirNameForDartFile(project, vFile);
+    final Project project = element.getProject();
```
alexander-doroshko: Funny story. I used to be a fan of `final` everywhere. Then at some point I noticed that one of my colleagues silently removes `final`s from my code with 'cleanup' commit message. I talked to him in person and we both realized that it's just a matter of preference. IntelliJ IDEA has a special highlighting (underline) for variables and fields that are reassigned somewhere. So, regardless of the `final` modifier, we do see which vars/fields are effectively final and which are not. See 'Reassigned local variable' in Java color scheme. Finally, we agreed that everyone writes code as they want and we don't 'cleanup' `final`s in existing code without any special reason.

jwren:
Thanks for landing.

Fair enough about the `final`s... it is a habit that I got from you that has stuck.   :)

PR #735, opened 2019-11-05
Include `analysis_options.yaml` as a valid analysis options file name in DartServerCompletionContributor

jwren:
See documentation at [dart.dev](https://dart.dev) showing this change: [dart.dev/guides/language/analysis-options](https://dart.dev/guides/language/analysis-options)


alexander-doroshko:
Thanks!
Was you able to verify it works? I wish I could but it doesn't seem to work on Windows =( 
https://github.com/dart-lang/sdk/issues/25114

jwren:
Nope... but it something that they intend to do still.  I just want the Dart Plugin to be compatible with the functionality when it becomes available.

PR #736, opened 2019-11-05
Dart annotation hover html improvements: bold the message, and break before the url.

jwren:


alexander-doroshko:
Dart error tooltip is already quite big, especially when there are several errors under cursor. I'm not sure we want it to make it even bigger and heavier. AFAIU error tooltips in other languages are not bold. I think I'll suspend this PR until we decide whether we want to add diagnostics to this tooltip. After that I'll also talk to my colleagues about the UI.

jwren:
Having the message and description look different makes it more readable.  The size doesn't change much, developers have big screens.  I agree that we should be consistent with other languages... Java/ other languages should make their main messages bold as well.  :)

Thanks Alex.

devoncarew:
Is this PR still active?

PR #737, opened 2019-11-06
Update the testData/sdk/ Dart sdk files to version 2.6.0-dev.8.2

jwren:
I didn't see any `Dart tests` failures with this update.

Copied from https://github.com/dart-lang/sdk/tree/2.6.0


alexander-doroshko:
Thanks. I noticed that tests require just a few files in the mock Dart SDK. So I left only required files just not to clutter the repository. Plus, updated 2 tests.

PR #741, opened 2019-11-10
change observatory folding text to 'debug service'

devoncarew:
- change observatory folding text to day `'Debug service available at ...'`

@alexander-doroshko @jwren


devoncarew:
Screen Shot 2019-11-09 at 7 31 48 AM

alexander-doroshko:
[Docs](https://dart.dev/tools/dart-vm) still say Observatory. Is it going to change? Will the message printed by the VM change?

devoncarew:
cc @kwalrath and @sfshaza2 - we may want to update those docs (https://dart.dev/tools/dart-vm) to have a reference to DevTools.

 > Will the message printed by the VM change?

It will change eventually, but that'll be a breaking change for a whole host of tools. The VM has added a flag to write the service protocol info to a file once the port is available. We'd want to move tools over to using that and only change the text of the message after some (long) period of time.

```
--write-service-info=<file_name>
  Outputs information necessary to connect to the VM service to the
  specified file in JSON format. Useful for clients which are unable to
  listen to stdout for the Observatory listening message.
```

jwren:
LGTM

PR #743, opened 2019-11-10
re-work the dart analysis server error handling

devoncarew:
Re-work the dart analysis server error handling; we're moving to a model where most analysis server crashes are automatically reported by the analysis server. We no longer need to rely on manually reported crashes (but still want to provide the user some insight into the state of the analysis server).

- we no longer show ballon notifications, but instead reply on logging into the IntelliJ event log
- we now show all exceptions, up to the first 100 (including duplicates)
- crash details are available from the event log (viewable as temp editor files)
- functionality from `DartAnalysisServerService.java` moved into `DartAnalysisServerErrorHandler.java`
- `DartFeedbackBuilder.java` becomes a no-op interface (to be deleted in the future once downstream plugins are updated)
- delete `DefaultDartFeedbackBuilder.java`
- we continue to allow manual feedback from `AnalysisServerFeedbackAction.java`, but this no longer includes crash info

Screen Shot 2019-11-09 at 12 32 26 PM

Screen Shot 2019-11-09 at 12 32 31 PM

@alexander-doroshko, @jwren, @MichaelRFairhurst


devoncarew:
Sample file attached for when the user clicks on `view details` in the event log.

[issue-1.txt](https://github.com/JetBrains/intellij-plugins/files/3828004/issue-1.txt)

alexander-doroshko:
LGTM.
Also /cc @stevemessick, the initial author if this code.

`DartFeedbackBuilder` seems to be not used by plugins and can be removed right away.

devoncarew:
We have one internal use of it - I'd prefer to deprecate, remove that use, and then delete the class.

jwren:
LGTM

alexander-doroshko:
> We have one internal use of it 

Ok, then is it safe to delete public `DartFeedbackBuilder.showNotification()`?

`dart.feedback.error` and `dart.feedback.file` are unused now in `DartBundle.properties`. Is it ok to delete them?

I also removed `<extensionPoint name="feedbackBuilder" interface="com.jetbrains.lang.dart.ide.errorTreeView.DartFeedbackBuilder" />` from `plugin.xml`. It's safe even if it's used in other plugins.

devoncarew:
> Ok, then is it safe to delete public `DartFeedbackBuilder.showNotification()`?

Yup - no uses of it (just the single subclass of `DartFeedbackBuilder`).

> `dart.feedback.error` and `dart.feedback.file` are unused now in `DartBundle.properties`. Is it ok to delete them?

Yup!

> I also removed `<extensionPoint name="feedbackBuilder" interface="com.jetbrains.lang.dart.ide.errorTreeView.DartFeedbackBuilder" />` from `plugin.xml`. It's safe even if it's used in other plugins.

Ah, cool. I didn't know if that would cause problems - good to know it won't.

PR #745, opened 2019-11-17
fix a null param issue from DartDocumentationProvider.java

devoncarew:
- fix a null param issue from DartDocumentationProvider.java
- unrelated, but remove two unused fields from `DartDocUtil.java`

It looks like it's possible to get a null value for a PSI element's containing file if the element is a directory or package.

Here's the exception we're seeing:

```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'psiFile' of com/jetbrains/lang/dart/ide/documentation/DartDocumentationProvider.getSingleHover must not be null
    com.jetbrains.lang.dart.ide.documentation.DartDocumentationProvider.$$$reportNull$$$0(), DartDocumentationProvider.java:-1
    com.jetbrains.lang.dart.ide.documentation.DartDocumentationProvider.getSingleHover(), DartDocumentationProvider.java:-1
    com.jetbrains.lang.dart.ide.documentation.DartDocumentationProvider.getSingleHover(), DartDocumentationProvider.java:182
    com.jetbrains.lang.dart.ide.documentation.DartDocumentationProvider.generateDoc(), DartDocumentationProvider.java:33
    com.intellij.lang.documentation.CompositeDocumentationProvider.generateDoc(), CompositeDocumentationProvider.java:136
    com.intellij.codeInsight.documentation.DocumentationManager$MyCollector.lambda$getDocumentation$2(), DocumentationManager.java:1176
    com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(), ApplicationImpl.java:1106
    ...
```

@alexander-doroshko @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -28,8 +28,6 @@ public class DartDocUtil {
 
   public static final String SINGLE_LINE_DOC_COMMENT = "///";
   private static final String NBSP = "&nbsp;";
-  private static final String GREATER_THAN = "&gt;";
-  private static final String LESS_THAN = "&lt;";
 
   public static String generateDoc(final PsiElement element) {
     if (!(element instanceof DartComponent) && !(element.getParent() instanceof DartComponent)) {
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocumentationProvider.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocumentationProvider.java
@@ -177,7 +177,11 @@ public class DartDocumentationProvider implements DocumentationProvider {
   @Nullable
   private static HoverInformation getSingleHover(final PsiElement element) {
     if (element != null) {
-      final PsiFile psiFile = element.getContainingFile();
+      @Nullable final PsiFile psiFile = element.getContainingFile();
+      // PsiElement.getContainingFile() can return null if the PSI element is a directory or package.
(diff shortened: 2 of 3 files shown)


alexander-doroshko:
I remember Jaime has already wrote about this exception but we still have no idea how it could happen. Do you have any ideas how to reproduce? We haven't received such exceptions at all. Is it happening for internal users only? What's the latest Dart plugin version where it happens? In my understanding, `PsiDirectory` or `JavaPackage` can't reach `DartDocUtil`. `DartDocumentationProvider` is registered with `language='Dart'` and directories can't have `Dart` language.

On the one hand, it's easy to add this `if` and forget about the exception. On the other hand, I'm very curious about the reason of this exception. Maybe it's shadowing a bigger problem.

devoncarew:
We're seeing it in `2019.2.3` (and in `2019.1.4`). I don't have a user repro for it, but we are seeing lots of exceptions from it - it's in the top three of reported exceptions.

From reading the API, I think the only way this could happen is if dartdocs are being requested for a directory. If we want more data about the issue, instead of this fix, we could land a change which logged the type of object we were seeing (`LOG.error()` calls will be reported via our internal instrumentation).

alexander-doroshko:
Well, I managed to reproduce it after rewriting `DartFileReference.Resolver` to return `PsiDirectory` instead of `PsiFile`. Still no idea how it can happen in real life. And not sure what to write to log to understand what happens. Ok, since the fix is harmless, I've pushed it. Thanks.

devoncarew:
> Ok, since the fix is harmless, I've pushed it. Thanks.

Great, thanks!

jwren:
Thank you Alex!

PR #747, opened 2019-12-03
Remove the UI option to enable Dart ML code completion

jwren:
- This will be temporary, so I commented out the source instead of deleting it
- I took the liberty of doing a bit of refactoring before commenting it out to make the code more readable
- I also added some `TODO`s to resolve when it comes time (hopefully soon) to re-enable the option
- Related bug: https://github.com/flutter/flutter-intellij/issues/4143


jwren:
@alexander-doroshko @devoncarew @lambdabaa @bwilkerson

Alex- We are disabling the ability to easily enable the ML completion for the time being as it isn't worth the risk of additional users hitting the issue mentioned.  We are working on a fix and when we are convinced it is resolved, we will re-enable the disabled UI.

alexander-doroshko:
Thanks, pushed.
Though I'm not sure why `myDoEnableMLBasedCodeCompletion` field name is better than `myMLBasedCodeCompletion` =).

PR #748, opened 2019-12-10
Fix in DartDocUtil.formatSignature(String) for constructor signatures

jwren:
Fix in `DartDocUtil.formatSignature(String)` for constructor signatures with an implicit `new` or `const`.

This fixes https://github.com/flutter/flutter-intellij/issues/4195

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -70,10 +70,11 @@ public class DartDocUtil {
   }
 
   private static String formatSignature(@NotNull final String signature) {
-    final int offsetToOpenParen = signature.indexOf('(');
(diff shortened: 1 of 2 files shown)


Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
```diff
@@ -70,10 +70,11 @@ public static String generateDoc(final PsiElement element) {
   }
 
   private static String formatSignature(@NotNull final String signature) {
-    final int offsetToOpenParen = signature.indexOf('(');
+    // Match the first open paren, "(", but ignore a starting "(new)" or "(const)" patterns.
+    final int offsetToOpenParen = signature.startsWith("(") ? signature.indexOf('(', 2) : signature.indexOf('(');
```
alexander-doroshko: Looks like this could be as simple as `offsetToOpenParen = signature.indexOf('(', 1);`

alexander-doroshko:
Pushed, thanks.

PR #749, opened 2019-12-11
Support non-Dart files in the Dart analyzer

simolus3:
This PR adds support for non-Dart files in the Dart plugin. When assigning the custom file ending to Dart, the analysis server is still used. It looks like outline, autocomplete, quickfixes, highlights and errors are all working (see screenshot, which comes from an analyzer plugin).

grafik

Fixes [WEB-41424](https://youtrack.jetbrains.com/issue/WEB-41424).

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/DartFileListener.java
+++ b/Dart/src/com/jetbrains/lang/dart/DartFileListener.java
@@ -53,9 +53,7 @@ public class DartFileListener implements VirtualFileListener {
     if (VirtualFile.PROP_NAME.equals(event.getPropertyName())) {
       fileChanged(myProject, event.getFile());
 
-      if (event.getFile().isInLocalFileSystem() &&
-          (DartAnalysisServerService.isFileNameRespectedByAnalysisServer(event.getOldValue().toString()) ||
-           DartAnalysisServerService.isFileNameRespectedByAnalysisServer(event.getFileName()))) {
+      if (event.getFile().isInLocalFileSystem()) {
         updateVisibleFilesIfNeeded();
       }
     }
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -836,13 +836,10 @@ public class DartAnalysisServerService implements Disposable {
    */
   @Contract("null->false")
   public static boolean isLocalAnalyzableFile(@Nullable final VirtualFile file) {
-    if (file != null && file.isInLocalFileSystem()) {
-      return isFileNameRespectedByAnalysisServer(file.getName());
-    }
-    return false;
+    return file != null && file.isInLocalFileSystem();
   }
 
-  public static boolean isFileNameRespectedByAnalysisServer(@NotNull String _fileName) {
+  public static boolean isFileNameRespectedByAnalysisServerWithoutPlugins(@NotNull String _fileName) {
     // see https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/src/generated/engine.dart (class AnalysisEngine)
     // and AbstractAnalysisServer.analyzableFilePatterns
     String fileName = _fileName.toLowerCase(Locale.US);
--- a/Dart/src/com/jetbrains/lang/dart/highlight/DartSyntaxHighlighterColors.java
+++ b/Dart/src/com/jetbrains/lang/dart/highlight/DartSyntaxHighlighterColors.java
@@ -8,9 +8,9 @@ import com.intellij.openapi.editor.colors.TextAttributesKey;
 import static com.intellij.openapi.editor.colors.TextAttributesKey.createTextAttributesKey;
 
 public class DartSyntaxHighlighterColors {
-  public static final String DART_ERROR = "DART_ERROR";
-  public static final String DART_WARNING = "DART_WARNING";
-  public static final String DART_HINT = "DART_HINT";
+  private static final String DART_ERROR = "DART_ERROR";
+  private static final String DART_WARNING = "DART_WARNING";
+  private static final String DART_HINT = "DART_HINT";
 
   public static final String DART_ANNOTATION = "DART_ANNOTATION";
   public static final String DART_CLASS = "DART_CLASS";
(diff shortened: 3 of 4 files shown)


jwren:
The change looks good to me.  @alexander-doroshko - is risk introduced here by considering all files as potentially getting Dart Analysis Server warnings and completions?  If so, we could add some protocol on the DAS startup to indicate to the Dart Plugin which file extensions the server will contribute to.  Would this be sufficient @simolus3? Or is having an undetermined file extension part of the intention here?

cc @stereotype441

alexander-doroshko:
@simolus3 Have you submitted a contributor license agreement? If not, please follow the steps described at http://www.jetbrains.com/agreements/cla/ to sign it.

simolus3:
I just signed the CLA.

> is risk introduced here by considering all files as potentially getting Dart Analysis Server warnings and completions

Just for completeness: My intention was that only files that were explicitly marked as Dart files would get analyzed (Via right-click and "Associate with File Type..." in IntelliJ). From my local testing that seems to have worked, but I could have easily missed something.

>  Would this be sufficient @simolus3? Or is having an undetermined file extension part of the intention here?

As long as I can have `*.moor` I'm happy :) I'm not sure the server could easily compile a list of file endings on startup though, as I think it loads plugins afterwards. Hopefully it's not necessary though.

jwren:
I talked with the analysis server developers.  Allowing error and warning information from some random file extension, i.e. `bar` for some `foo.bar`, to be displayed to the user seems low risk.

What I don't see in the PR is how code completion and other integration points mentioned are working.  For example, code completion is integrated via `DartServerCompletionContributor` which explicitly calls out which file types are supported: `*.dart`, `*.html`, `.analysis_options`, etc.

simolus3:
> For example, code completion is integrated via `DartServerCompletionContributor` which explicitly calls out which file types are supported

It checks for the associated language in https://github.com/JetBrains/intellij-plugins/blob/9d319e6d427d3195cd28d2654da1d4c74d6e03f8/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java#L62-L63

I manually had to mark those files as Dart for this to work. This was acceptable for me since it only needs to be done once. Since an IntelliJ installation could have other plugins which may support that file ending, I think manually setting it to Dart is the best solution here.
grafik

Regarding the other integration points:
- Outline: Seems to be provided by `DartStructureViewModel` without further checks on file endings. I assume IntellIiJ uses that class for files associated with `DartLanguage.INSTANCE` only.
- Quickfixes: Checks for `file instanceof DartFile && DartAnalysisServerService.isLocalAnalyzableFile(file.getVirtualFile())`, which now includes all files that have been set to Dart manually.

alexander-doroshko:
Please expect some delay with this PR.

It is not absolutely trivial. The cost of supporting and maintaining the code that it modifies is not zero.

And the solution that it assumes is not 100% ideal. For example, marking non-Dart files as Dart ones has a lot of implications. Dart lexer, parser and tons of other language-specific features try to work for all files with Dart file type, but we know in advance that it makes no sense for your use case.

We have more than 300.000 users of the Dart plugin. We are happy to accept changes that make something better for a significant amount of users and don't make anything worse for anyone. We're also ok with PRs that improve something for just a few users given that such PRs are (a) trivial, (b) 100% safe, (c) 100% correct, (d) have zero maintenance cost.

This PR is simple but not absolutely trivial; probably safe but needs to be tested; assumes not 100% correct workflow with known downsides; not covered by tests; maintenance cost is not zero (each time when working with the touched code we'll have to keep your use case in mind).

So we need to weigh pros and cons before landing this. Are you the only developer who needs this? Maybe you could simply use you custom Dart plugin build?

simolus3:
Thanks for the transparency, I understand your concerns.

> Are you the only developer who needs this?

I'm using this for a Dart package I maintain, [moor](https://pub.dev/packages/moor_flutter). That's definitely not a significant amount of Dart plugin users, but I'd say it's fairly popular judging by the pub score and GitHub activity. I think most users of that package would also profit from this change.

> Maybe you could simply use your custom Dart plugin build?

That probably wouldn't work, but maybe I can publish a separate IntelliJ plugin for moor that depends on the Dart plugin and also uses the analysis server. I assume that's also what the Flutter plugin does, so maybe it could work. That would still need some changes to this plugin though, like some api on the `DartAnalysisServerService` that could mark a file as server analyzable without having an impact on other integrations. Would adding such api be acceptable to you? It likely has a lower risk (since the Dart scanner and parser isn't affected) and it could be easier to test and maintain.

simolus3:
Another idea I just had: Would it be acceptable to provide another language next to `DartLanguage` and `DartInHtmlLanguage`? Maybe something like `DartAnalyzerOnlyLanguage` that doesn't have a default extension and explicitly needs to be associated with a file ending. It would still use the analyzer for syntax highlighting, errors, completion, etc. 
__Pros__:
- it wouldn't interfere with the current lexer and parser at all, since it's a different language. 
- it's both safe and correct, since the analysis server supports that functionality and it doesn't affect any local analysis in IntelliJ
- authors of analyzer plugins don't have to write another IntelliJ plugin for their package

Any thoughts on this? I could close this PR and work on the alternative if you think it's more likely to be accepted, thanks!

alexander-doroshko:
Need to investigate this idea. Language might be not enough. You might need to register a different file type as well (and this is a user-visible entity and must have some default file ending).

A custom plugin based on the Dart plugin should be possible. |It could register Moor file type, language, annotator, completion contributor, etc. `DartAnalysisServerService` allows to register listeners that provide low-level info from server. Theoretically, that's enough but not convenient. We can accept some minor changes in Dart plugin that would allow to use not so low-level API. Again, need to investigate this deeper, it's not a trivial question.

piotrtomiak:
How about registering a new file type, which extends `DartFileType`, and which contents are of Dart language? I did something like that to support Angular SVG files. Than the only change in Dart plugin is that instead of direct comparision with `DartFileType.INSTANCE`, you need to perform `instanceof DartFileType`. The plugin could also contribute an additional file extension to Dart server to include in analysis.

alexander-doroshko:
Reusing `DartLanguage` for files, which are in fact have different language, is probably the biggest issue I'm concerned about. There're 35 entries of `language="Dart"` in the `plugin.xml` file, and probably most of the extensions shouldn't work for custom non-Dart language. Still thinking...

simolus3:
@alexander-doroshko I'm wondering if you have any update on if / how this could be addressed.

Recent activity on [WEB-41424](https://youtrack.jetbrains.com/issue/https://youtrack.jetbrains.com/issue/WEB-41424) indicates that I'm not the only one writing analyzer plugins for non-Dart files. If possible, I'd prefer the approach of having another language and file type, as it reduces maintainance work for me and others writing analyzer plugins (we wouldn't have to ship an IntelliJ plugin as well).

You mentioned that a file type is user visible and has to have a default ending. If we end up using this approach, we could just name it something like "non-Dart language handled by the Dart plugin" which makes this clear. A default file ending is harder, maybe we can just pick `.moor` or some bogus value that's unlikely to be encountered.
Of course, it's your decision - I just wanted to check if there's something new. I'd love to work on that approach to replace this PR if it's ok, but either way is fine for me. Thanks!

alexander-doroshko:
Sorry for the delay. Given that the interest for this feature grows, I should have another look. I'll revisit when I'm done with current tasks.

simolus3:
Hi @alexander-doroshko, just a friendly reminder :)

I'd still love to help here, and I still prefer the approach of a second language and file type that users would have to assign. But I'm fine with any approach to get this working.

alexander-doroshko:
Given that https://youtrack.jetbrains.com/issue/WEB-41424 keeps getting votes, we'll eventually come up with some solution. I'm not sure that the solution will be based on this PR, so I'm closing it. Please watch/vote/comment in https://youtrack.jetbrains.com/issue/WEB-41424 to get updated when it's fixed.

PR #750, opened 2019-12-13
Improvements in the Dart hover documentation results, WEB-9993

jwren:
I almost gave up looking for a solution before proposing we write our own little markdown processor or importing a different java library when I found `MarkdownUtil`.  See the comment on the method `MarkdownUtil.replaceHeaders()`.

The `MarkdownProcessor` is providing value, we just need additional changes for things code blocks, headers and lists.

Further improvements here probably mean additions and bug fixes to `MarkdownUtil`:

- `<pre><code>` are being inserted, but not all of the ``` are being removed, also in the popup I don't see the text in a different mono-space font.  Is there some other html block that the hover popup expects?
- Headers work better, but are not perfect.  See screenshot below, this shows that not all headers are created equal.

- `[ref]` and `[: some code :]` are not respected yet.  We probably want these formatted with `<code></code>` blocks?  The fix here is probably a new utility method on `MarkdownUtil`.  We should be careful not to modify if the `[]` symbols are in source code being documented.

step_one

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -180,14 +180,14 @@ public class DartDocUtil {
     builder.append("<br>");
     builder.append("</code>\n");
     if (docText != null) {
-      final MarkdownProcessor processor = new MarkdownProcessor();
-      final String markdownProcessorOutput = processor.markdown(docText);
-      List<String> lines = Arrays.asList(StringUtil.splitByLines(markdownProcessorOutput));
+      List<String> lines = StringUtil.split(docText, "\n");
       MarkdownUtil.replaceCodeBlock(lines);
       MarkdownUtil.removeImages(lines);
       MarkdownUtil.replaceHeaders(lines);
       MarkdownUtil.generateLists(lines);
-      builder.append(StringUtil.join(lines, "\n"));
+
+      MarkdownProcessor processor = new MarkdownProcessor();
+      builder.append(processor.markdown(StringUtil.join(lines, "\n")));
     }
     // done
     return builder.toString().trim();
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -181,7 +181,7 @@ public class DartDocUtil {
     builder.append("</code>\n");
     if (docText != null) {
       final MarkdownProcessor processor = new MarkdownProcessor();
-      final String markdownProcessorOutput = processor.markdown(docText.trim());
+      final String markdownProcessorOutput = processor.markdown(docText);
       List<String> lines = Arrays.asList(StringUtil.splitByLines(markdownProcessorOutput));
       MarkdownUtil.replaceCodeBlock(lines);
       MarkdownUtil.removeImages(lines);
--- a/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/documentation/DartDocUtil.java
@@ -10,6 +10,7 @@ import com.intellij.psi.tree.IElementType;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.util.ArrayUtil;
 import com.intellij.util.SmartList;
+import com.intellij.util.text.MarkdownUtil;
 import com.jetbrains.lang.dart.DartTokenTypesSets;
 import com.jetbrains.lang.dart.psi.*;
 import com.jetbrains.lang.dart.util.DartGenericSpecialization;
@@ -21,6 +22,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


alexander-doroshko:
Please check `DartServerDocUtilTest` and add needed tests.

I think we may fix `MarkdownUtil` according to our needs. I'll make sure that other users of this class are not broken. There's `MarkdownUtilTest` to check and add tests.

devoncarew:
Just a quick comment - the `[: foo :]` style is meant to represent code - it's identical to backticks:

```
` foo `
```

The `[: foo :]` is a bit of a historical artifact now; I think all user code now uses backticks, and the only real `[: foo :]` uses may be in dart: code.

(the `[ foo ]` style is separate, and is meant to be an identifier reference - the analyzer might offer to hyperlink that)

http://dartdoc.takyam.com/articles/doc-comment-guidelines/#inline-styles

jwren:
Tests in `DartServerDocUtilTest` as well as `DartDocUtil` now all pass.  I also added some tests to verify explicitly that the utilities in the `MarkdownUtil` class are called.

jwren:
Thanks for that context Devon.  We won't prioritize `[: foo :]` syntax.

alexander-doroshko:
Thanks for the update (and sorry for delay). A few questions.
- I noticed that since SDK 2.8-edge, server includes `containingClassDescription` in `getHover()` for a class (and it's this class itself). Does it make any sense? Looks like a server bug for me. Class doesn't contain itself. I think it just adds a useless line to the hover tooltip. Should I file an issue in the SDK issue tracker?
image
- We seem to have lost understanding of an empty line as a paragraph separator. For example, see changes in `testClassMultilineDoc1`. We used to have `doc4` as a separate paragraph (surrounded with `<p>`), now it's not. 
- Just curious, explicit string trimming - can it make any user visible change? Or it's just for more accurate HTML output?

I can land it as is if you want, as a WIP intermediate result.

Idea for further improvements: go through the SDK docs and make sure that our rendering is the same as on https://api.dartlang.org. For example, at the moment we fail to show [Object](https://api.dartlang.org/stable/dart-core/Object-class.html) class docs. Looks like line-by-line handling doesn't always work fine. May be we could see how dartdoc tool handles markdown?
image

jwren:
Not sure about `containingClassDescription`, I'll ask.

Multi-paragraphs, yes, I have some additional work in the `MarkdownUtil` to fix this.

Explicit string trimming-- just general hardening to ensure that the HTML output doesn't include additional white space that should be deleted.

Dartdoc -- good idea, I'll ask about this as well.

In-line code references, i.e. `[`references`]`, should and can be handled better as well, I have started this as well.

As for whether to land, I'd say to land this  now so that we can iterate on additional pieces.  Keeping changes small is generally good when it doesn't come at a cost.

alexander-doroshko:
Ok, pushed, but without those extra `Containing class X` in the expected test data. I tend to think it's a server bug that will be fixed. Even if it's not fixed in server, I think we should filter out duplicating information in our doc popups.

jwren:
The problem illustrated with the hover over the `Object` type is that `[code]` and `[:code:]` aren't handled.  For some `[code]` (that isn't followed by `(http://example.com)`), they should be replaced with `<code>` tags, i.e.:

`[some in-line example code] --> <code>some in-line example code</code>`

Options:

1. Change in `MarkdownUtil` -- the input and output of an array of lines makes sense for the other parsing tasks it is doing, but would make this change hard to read/maintain, but could be done.
2. Or, it could be a change we push into `MarkdownProcessor`.
3. We could have an additional utility in `MarkdownUtil` that takes the content as a single `String` that we take and return a result for.

alexander-doroshko:
Thanks for the investigation. On the one hand, it's a problem in `com.petebevin.markdown.MarkdownProcessor`, it's not handling `[foo]` not followed by `(link)` correctly.
On the other hand, we need full support for Dart-specific `[code fragments]` anyway, including navigation info that we get from the Analysis Server. So I'm thinking about:
- Handling `[code fragments]` and `[:code fragments:]` right in the `DartDocUtil` before using `MarkdownProcessor`
   - if there are no regions with navigation data in `[code fragment]` then simply replace it with smth like `<code>code fragment</code>`
   - if there is `DartNavigationRegion` (like `[hashCode]` or `[operator ==]`) then replace it with a special link like `<a href='psi_element://path_to_target_file:offset'>hashCode</a>` (see `DocumentationManagerProtocol`)
- Implement `DartDocumentationProvider.getDocumentationElementForLink()` that would understand `path_to_target_file:offset` and return `DartResolver#getElementForNavigationTarget()`.

Looks like we can get great support for code links this way.
image

jwren:
Thanks for the information around `<a href='psi_element://path_to_target_file:offset'>hashCode</a>`.

I'll look to create a change in `DartDocUtil` in the coming few weeks.  Yes, these changes would be sufficient for the `Object` dartdoc case.

jwren:
Making progress:
Screen Shot 2019-12-31 at 12 13 31 AM

The `DartNavigationRegion` offsets won't map nicely into the char offsets we have in the document `String`.  When we get some documentation:

```
/**
 * Some
 * sentence
 * here.
 */
```

The document text that we are passed is `Some sentence here`, not `\**\n * Some\n * sentence ...`.

Is there some utility for this mapping that we already have, or would it be prudent to have these `DartNavigationRegion`s passed along in the Dart Analysis Server?  (I suspect the second option...?)

alexander-doroshko:
Yes, one way is to add navigation regions to the server's docs related functions.

Handling everything on the Dart plugin end might be doable as well. For example, before asking server for hover info, we could process all bracketed text fragments (not followed by `(link)`) and remember corresponding DartNavigationRegions, if any (see also `DartDocUtil#getDocumentationText`). Assuming that the text received from the server contains the same bracketed code fragments, we could map them to DartNavigationRegions.

PR #751, opened 2019-12-20
Append ` --debug` in the `webdev serve` folded message in the Dart Plugin

jwren:
Append ` --debug` in the `webdev serve` folded message in the Dart Plugin.  The flag `--debug` is required in the `serve` context


jwren:
@alexander-doroshko

PR #752, opened 2019-12-27
Update the Dart plugin with the change from github.com/JetBrains/intellij-community/pull/1279

jwren:


jwren:
@alexander-doroshko

PR #755, opened 2020-01-08
Modify the Dart remote debugger

jwren:
Modify the Dart remote debugger to be able to debug when auth codes are used to connect to the Dart VM.

The dialog title and message match the Dart DevTools connect page: https://github.com/flutter/devtools/pull/1513

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
@@ -2,6 +2,7 @@
 package com.jetbrains.lang.dart.ide.runner.server.vmService;
 
 import com.google.common.base.Charsets;
+import com.google.common.base.Predicate;
 import com.intellij.execution.ExecutionResult;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
@@ -12,6 +13,8 @@ import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ReadAction;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.ui.InputValidator;
+import com.intellij.openapi.ui.Messages;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.Pair;
@@ -66,8 +69,6 @@ public class DartVmServiceDebugProcess extends XDebugProcess {
 
   @Nullable private final ExecutionResult myExecutionResult;
   @NotNull private final DartUrlResolver myDartUrlResolver;
-  @NotNull private final String myDebuggingHost;
-  private final int myObservatoryPort;
 
   private boolean myVmConnected = false;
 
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Dart Remote Debug configuration has host and port fields. With this change we simply ignore them. First, we need to decide whether we want to keep (and use) the fields or remove completely. 

There are use cases when the auth token is not required and we can start debug process without bothering user with questions. I mean a) using previous versions of Dart SDK and b) using `--disable-service-auth-codes`. I'm not sure how important it is to support these scenarios (especially given that 
`--disable-service-auth-codes` is going to be removed).

Are there other use cases (maybe in Flutter) when we can proceed without questions?

jwren:
First, here is what the dialog looks like.  Note again that the text messages match Dart DevTools (PR: https://github.com/flutter/devtools/pull/1513).

Screenshot from 2020-01-09 10-25-41

Adding some Dart Devtools and Flutter tooling developers for their perspective.  @jacob314 @kenzieschmoll @devoncarew @bkonyi 

My understanding is that nearly all runs of the Dart VM with time will have an authentication code.  When there is no authentication code, entering the observatory URL still works by putting the URL into this new dialog.  Having a dialog **not** appear in some rare cases when the user wants to not have an authentication code and just set a single port, feels like a user experience hazard.  I am imagining the dialog not appearing because users accidentally mis-configured their run configuration. 

You are correct Alex that the host string as well as port should be removed from the Remote Debugger dialog.  I will follow up when if/ when we land this PR.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -157,8 +164,40 @@ public void stackFrameChanged() {
...
+
+            @Override
+            public boolean canClose(String inputString) {
+              return true;
+            }
+          });
+      if (isDebugUrl.test(httpDebuggerUrl)) {
+        assert httpDebuggerUrl != null;
+        final String wsToConnect;
+        if (httpDebuggerUrl.startsWith("http://")) {
```
jacob314: you might want to first trim the URL. We noticed DevTools users had trouble copying the URLs with leading and trailing spaces and so we stripped the spaces to avoid confusing failure to connect errors.

jwren: Good to know, PR updated.  Thanks Jacob

bkonyi:
> First, here is what the dialog looks like. Note again that the text messages match Dart DevTools (PR: [flutter/devtools#1513](https://github.com/flutter/devtools/pull/1513)).
> 
> Screenshot from 2020-01-09 10-25-41
> 
> Adding some Dart Devtools and Flutter tooling developers for their perspective. @jacob314 @kenzieschmoll @devoncarew @bkonyi
> 
> My understanding is that nearly all runs of the Dart VM with time will have an authentication code. When there is no authentication code, entering the observatory URL still works by putting the URL into this new dialog. Having a dialog **not** appear in some rare cases when the user wants to not have an authentication code and just set a single port, feels like a user experience hazard. I am imagining the dialog not appearing because users accidentally mis-configured their run configuration.
> 
> You are correct Alex that the host string as well as port should be removed from the Remote Debugger dialog. I will follow up when if/ when we land this PR.

FYI, there's also a `--write-service-info=<filename>` flag that can be used to write the service protocol connection information to a known location. If this plugin is used to launch Flutter applications, you could provide this flag and automagically have all the information you need to connect to a VM service which requires authentication codes.

jwren:
@bkonyi Good to know, I didn't know about `--write-service-info=<filename>`.  The Flutter Plugin will takes care of the Flutter users with their flutter attach configs,etc., so we can think just about pure-Dart here.  How highly should we prioritize adding the ability to specify this flag?

alexander-doroshko:
Thanks! I've pushed this PR with minor changes. In particular, I've moved a blocking call of `showInputDialog()` outside the constructor. Also, handled the case if user pressed Cancel in this modal dialog.

Some thoughts about the next iteration. Since we are going to remove `host` and `port` fields from the UI, user has no clues how the remote app should be started and where to get the URL IDE is asking for. I think I'd be confused when trying to debug a remote Dart VM app but asked for some "URL to a running Dart or Flutter application".

Maybe we could give clues right in the dialog that is asking for the URL? Something like this but less verbose and more clear =)

- to connect to the remote Dart VM app, start it with `--enable-vm-service --pause_isolates_on_start` VM options and paste the Debug service URL (Observatory URL) that will be printed in its console on start
- to connect to the remote Dart web app, start `webdev serve --debug` option and paste the DevTools URL that will be printed in the webdev's console
- to connect to the remote Flutter app/whatever else, ...

PR #757, opened 2020-01-17
Switch to HTTPS for maven dependency

robhurring:
I started seeing the following error when attempting to use the `error-prone` plugin:

>Error:Failed to download error-prone compiler JARs: Failed to download 'http://central.maven.org/maven2/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1.jar':
Request failed with status code 501

It looks like they switched over to HTTPS-only on 1/15/2020. 

* https://support.sonatype.com/hc/en-us/articles/360041287334
* https://stackoverflow.com/questions/59763531/maven-dependencies-are-failing-with-a-501-error

### DIFF
--- a/error-prone/resources/library/error-prone.xml
+++ b/error-prone/resources/library/error-prone.xml
@@ -8,6 +8,6 @@
     <item url="https://repo1.maven.org/maven2/com/google/errorprone/javac/9+181-r4173-1/javac-9+181-r4173-1.jar"/>
 
     <!-- this is a workaround missing dependency in error_prone_core-with-dependencies: https://github.com/google/error-prone/issues/1238 -->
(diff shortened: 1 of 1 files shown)


C-Otto:
Thank you. I raised an issue for that: https://youtrack.jetbrains.com/issue/IDEA-231506

chashnikov:
I've fixed the problem by upgrading Error Prone to 2.3.4 where the problem with missing dependency on failureaccess artifact is fixed. It has a different [missing dependency](https://github.com/google/error-prone/issues/1445) though, but I included it manually.

dmitry-fa:
A minor addition: switching to the new Idea version is not enough to make 'Failed to download error-prone compiler JARs' error disappear, one should **reinstall** the `error-prone` plugin as well.

PR #759, opened 2020-02-07
Remove the host and port from the Dart Remote Debugger configuration

jwren:


alexander-doroshko:
What's the difference between `--debug` and `--enable-vm-service`? 
`--debug` is not listed on https://dart.dev/tools/dart-vm, neither is it printed by `dart --help`.

alexander-doroshko:
Should we give hints not only for Dart VM apps but also for web and maybe Flutter?
I mean do we want to write something like:
- to connect to the remote Dart VM app, start it with `--enable-vm-service --pause_isolates_on_start` VM options and paste the Debug service URL (Observatory URL) that will be printed in its console on start
- to connect to the remote Dart web app, start `webdev serve --debug` option and paste the DevTools URL that will be printed in the webdev's console
- to connect to the remote Flutter app/whatever else, ...

jwren:
Yes, `--debug` was a mistake, thanks.  I discussed with Devon as a sounding board here as well.  He isn't convinced that having directions for each context (Dart VM, Dart Web, Fluter, etc.) is necessarily wise--- there might be different or additional flags in the future which would require us touching the UI.  I updated it with a new message, below, but the issue now is that I seemed to choose the wrong options or widget: `JTextPane`.

```
When launched, a dialog will take, and connect to, the Dart VM URL.

For Dart VM applications consider passing '--enable-vm-service' and possibly 
'--pause_isolates_on_start'.  For other contexts, such as Dart web apps, reference 
appropriate flags and options.
```

devoncarew:
Perhaps:

```
When run, this launch configuration will prompt for a Dart service protocol URL to connect to.

For Dart VM applications launch with '--enable-vm-service'.  For other contexts - such as Dart web apps - reference the appropriate flags and options.
```

alexander-doroshko:
What about this?

image

alexander-doroshko:
What about this?
image

alexander-doroshko:
What do you think about adding a bit more information / instructions to the dialog that pops up?
image

PR #760, opened 2020-02-11
New run producer for DartWebdevConfigurations

jwren:
@alexander-doroshko


alexander-doroshko:
Please add a check that HTML file belongs to a Dart project. Probably we also need to check that the SDK version is sufficient to run webdev tool. Or maybe we need to check webdev package presence among dependencies in pubspec?

jwren:
Added a minimum Dart SDK version of `2.6.0`

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -83,6 +83,11 @@
   public static final String MIN_SDK_VERSION = "1.12";
   private static final String MIN_MOVE_FILE_SDK_VERSION = "2.3.2";
 
+  // Webdev works going back to 2.6.0, future minimum version listed in the pubspec.yaml, link below, won't mean that 2.6.0 aren't
+  // supported.
+  // https://github.com/dart-lang/webdev/blob/master/webdev/pubspec.yaml#L11
+  private static final String MIN_WEBDEV_SDK_VERSION = "2.6.0";
```
alexander-doroshko: I noticed that we already have `DartWebdevParameters.MIN_DART_SDK_VERSION_FOR_WEBDEV`, though it's set to "2.5".  Will 2.5 be enough for Webdev run configuration? Let's merge 2 constants into 1 if so.

alexander-doroshko:
IDE projects may be huge and mix different technologies. We need to check that the HTML file is within a Dart project root (which we can detect by a `pubspec.yaml` file presence). Actually, that's what we already do in `DartWebdevParameters.check()`. Additionally, we check there that HTML file is not right in the Dart project root (webdev can't serve such files).
`DartWebdevConfigurationProducer` must be at least as strict as `DartWebdevParameters.check()` - to be sure that we don't produce red run configuration. We may make producer more strict if we want. For example, do you think we should limit the set of folders where `DartWebdevConfigurationProducer` works? I mean, we could ignore HTML files which are within `dart_root/packages/` or `dart_root/lib`.

jwren:
I just updated PR to not have the duplicated `DartWebdevParameters.MIN_DART_SDK_VERSION_FOR_WEBDEV`.

With regard to your last comment, I did consider verifying more information into the producer.  As you point out thought, `DartWebdevParameters.check()` verifies and prints appropriate warnings.  It seems to me that if the users is creating and wanting to run some html file, having the warning printed out (instead of not having the "Run index.html"), seemed like the better option.

alexander-doroshko:
I don't think it makes much sense to auto-create Dart-specific run configuration for an HTML file that is not related to Dart at all (e.g. it is in JavaScript part of an IntelliJ project). "Run index.html" will be there from JavaScript Debug producer. We want DartWebdevConfigurationProducer to work only where it is better than JavaScript Debug producer. Suggesting Run/Debug option that would immediately result in error doesn't sound like a good UX.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/webdev/DartWebdevConfigurationProducer.java
```diff
@@ -0,0 +1,79 @@
...
+                                            final @NotNull ConfigurationContext context) {
+    final VirtualFile htmlFile = getHtmlFileFromContext(context);
+    return htmlFile != null && htmlFile.getPath().equals(configuration.getParameters().getHtmlFilePath());
+  }
+
+  @Nullable
+  private static VirtualFile getRunnableHtmlFileFromContext(@NotNull final ConfigurationContext context) {
+    // return null if this is not a Dart project
+    final Project project = context.getProject();
+    if (project == null || !DartAnalysisServerService.getInstance(project).serverReadyForRequest()) {
```
alexander-doroshko: Just realized that `serverReadyForRequest()` tries to start the analysis server if it's not already running. And this line of code runs on any right click in any file in any project =).

I don't think we care about analysis server in RunConfigurationProducer. We care first of all about being able to start the produced run configuration right away.

jwren:
Thanks for the review Alex, working on updating the PR now.

alexander-doroshko:
Thanks, pushed. Sorry for delay.

PR #763, opened 2020-02-14
[Cucumber] Fix StringIndexOutOfBoundsException on CucumberJavaParamet…

sshuvalov:
…erTypeReference.getParameterTypeName()

Added additional annotation type check.


sshuvalov:
@avokin review this?

avokin2:
Could you provide the exception's stacktrace?

This check should be done in `CucumberJavaUtil#isCucumberStepAnnotation`. Also I suggest not to use class names from `CucumberJavaExtension#CUCUMBER_JAVA_STEP_DEFINITION_ANNOTATION_CLASSES`. It's classmarkers to detect cucumber-java library version (no guarantee that this list contains step definition annotations)

PR #765, opened 2020-02-20
Update OSMORC plugin to bnd 5.0.0

pkriens:
bnd 5.0 was recently released and some of our customers are asking for the updated OSMORC plugin. This patch is unfortunately not tested because I am not sure how to build and test this repository? Is it possible to streamline these updates? Just create a PR like this? Could we build a snapshot version of the plugin in our build?

There seem to be a number of companies interested in this plugin and it would be nice if we could work together to keep this plugin synchronized with the bnd release cycle.

Signed-off-by: Peter Kriens <Peter.Kriens@aqute.biz>

### DIFF
--- a/osmorc/osmorc-jps-plugin/intellij.osgi.jps.iml
+++ b/osmorc/osmorc-jps-plugin/intellij.osgi.jps.iml
@@ -14,40 +14,40 @@
     <orderEntry type="module" module-name="intellij.maven.jps" />
     <orderEntry type="module-library" exported="">
       <library name="bndlib" type="repository">
-        <properties maven-id="biz.aQute.bnd:biz.aQute.bndlib:4.2.0">
+        <properties maven-id="biz.aQute.bnd:biz.aQute.bndlib:5.0.0">
           <exclude>
             <dependency maven-id="org.slf4j:slf4j-api" />
           </exclude>
         </properties>
         <CLASSES>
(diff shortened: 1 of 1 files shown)


trespasserw:
Filing a YouTrack request remains the easiest - there is still no easy way to open the plugin sources as a project and run a test suite.

pkriens:
I deleted the request because I noticed that I had forgotten to update the osmorc/resources/META-INF/plugin.xml file, see previous patch https://github.com/JetBrains/intellij-plugins/commit/002c631d1ad802e189265e54aa98d9f25f8e9c3f#diff-8cd697e51e611b7dd1a3c6d3fea8a546

Do you prefer YouTrack or a PR?

trespasserw:
Ah! I thought you deleted it after noticing the [conflicting commit](https://github.com/JetBrains/intellij-plugins/commit/4a85ae70fdfd12712d887b9e5d8309a0b2aa98e1). Coincidentally, I already upgraded the lib a few minutes before getting the PR notification.

pkriens:
Ah, great minds think alike ? :-)

Next time, a YouTrack or PR? I've added an entry in our release process for this.

trespasserw:
Well, almost :) Got a [request](https://youtrack.jetbrains.com/issue/IDEA-232978) in YouTrack not long ago.

YouTrack is preferred (at least until there is a way to build and test the plugin independently).

PR #769, opened 2020-04-22
Enhancement proposal for Cloudformation templates

Pelirrojo:
Below are two small additions to the cloudformation base templates. These include:

* Example of Parameters
* Example of Outputs
* Example of Intrinsic Functions
* Example of Comments
* Visual interface improvements (AWS::CloudFormation::Interface)

The main improvement is that these templates are tested and work and help to generate modular templates that accept parameters and have easily reusable outputs


Pelirrojo:
Hello,

With this pull request I would like to offer an improvement in the cloudformation base templates, currently they only create one EC2 resource and I suggest that they have the four most common sections and an example of comments.

The same example is in the two accepted formats: **json** and **yaml**.

Do you find it interesting to merge it? Is there any contributing document that you have not seen or followed?

Thanks in advance,
Manuel

PR #771, opened 2020-05-01
Bugfix/idea 239899 parameter types and step hooks

guyv:
fix for [IDEA-239899](https://youtrack.jetbrains.com/issue/IDEA-239899)

Add glue for parameter types in separate package + scan for step hooks BeforeStep and AfterStep


avokin2:
Merged

PR #774, opened 2020-07-30
WEB-46098 Deno: Support custom DENO_DIR environment variable

Nerixyz:
This PR aims to support the `DENO_DIR` environment variable.

Deno itself also checks whether the variable is absolute or not [here](https://github.com/denoland/deno/blob/d619e3c7ac8618c5cd30fcc314db2ef994c604e4/cli/deno_dir.rs#L18-L22). I'm not really sure how the paths are handled inside of IntelliJ so I'll only mention it but this has to be checked.

### DIFF
--- a/Deno/src/com/intellij/deno/DenoUtil.kt
+++ b/Deno/src/com/intellij/deno/DenoUtil.kt
@@ -37,6 +37,10 @@ object DenoUtil {
   }
 
   private fun getDenoDirPath(): String {
+    val denoDir = System.getenv("DENO_DIR");
+    if (denoDir != null) {
+      return denoDir
+    }
     val userHome = SystemProperties.getUserHome()
     if (SystemInfoRt.isMac) {
       return "$userHome/Library/Caches/deno"


olemartinorg:
@Nerixyz Thanks for looking into this! I was about to start looking into how to change the plugin myself to fix this glaring issue.

While it's cool that the `DENO_DIR` environment variable now works in the plugin, I'm unsure if this covers all the needs. My impression was that with Deno projects you'll want to check in the project-specific `DENO_DIR` into source control (just like with `node_modules` for NodeJS). I hoped to be able to configure the Deno plugin to have the `DENO_DIR` in `$PROJECT_DIR$/deno` and have this configuration distributed via `.idea/deno.xml` so that other developers don't need to configure this manually.

Now, if there was an IDEA plugin to have project-specific environment variables, that would probably solve this, but I couldn't find any obvious candidates.

Nerixyz:
@olemartinorg I get what you're trying to solve, however I think this has to be done as a seperate feature. Also, this is only done at an IDE level so for example commands in the terminal won't respect this variable. You might want to open an issue on YouTrack.

This PR is more focused on supporting the variable at all. I myself for example have the directory on a different drive and the plugin won't recognize this directory that's why I want this to be supported.

anstarovoyt:
@Nerixyz Hi. Thank you for the pull request. I am ready to merge it, but have you submitted a contributor license agreement? If not, please follow the steps described at http://www.jetbrains.com/agreements/cla/ to sign it.

Nerixyz:
@anstarovoyt Thanks, I just submitted it.

anstarovoyt:
https://github.com/JetBrains/intellij-plugins/commit/ed8f67c409a16a1bb877f9e882940e567fcb3f41

anstarovoyt:
@Nerixyz I've merged the PR. Unfortunately, this repo just a mirror so the commit was merged in the original repo manually.

PR #732, opened 2019-11-05
Fix detection of Glue code

adel-chebbine:
Add support of the new package "io.cucumber.java" for "Before" and "After" annotations

PR #775, opened 2020-08-24
Cucumber-Groovy: Fix glue code detection for Groovy, too

KLBonn:
As in [this PR](https://github.com/JetBrains/intellij-plugins/pull/732), also the Cucumber for Groovy plugin should recognize the newer Cucumber base packages and create appropriate StepDefinition templates.

This would be very helpful for anyone using a recent Cucumber version with Groovy.


KLBonn:
Anyone available to check and possibly merge this one?

KLBonn:
< bump >
Do you see problems with this PR?
Or is there anything I should improve?

At our site, we do not use the ancient versions of Cucumber anymore, so the currently released Cucumber-Groovy plugin is not very supportive for us.

avokin2:
Merged in master

PR #788, opened 2020-12-07
Create docker-image.ymlhello world

j420247:


anstarovoyt:
@j420247 Hi! I don't see any reasons to add the docker files there. If I am not right please correct me.

PR #790, opened 2020-12-31
IDEA-138550 Add basic support for bnd-maven-plugin

ibauersachs:
This PR adds a new FacetImporter to detect the bnd-maven-plugin to add and configure the OSGI facet.

The generated manifest and jar will only be similar to the Maven plugin, not the same. The Maven model available during the import is not the same as the native Maven model, and the two-phase execution (parsing in the facet, Bnd-execution in the JPS build) is not the same as the everything-at-once in an actual Maven build. It is however sufficient for basic support and to enable all the existing code assistance.

Detection of the plugin is shown as a new UI option:
image

The embedded bndtools are updated to v5.2.0.

### DIFF
--- a/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/BndWrapper.java
+++ b/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/BndWrapper.java
@@ -44,6 +44,7 @@ import java.io.FileInputStream;
 import java.util.*;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
+import java.util.jar.Manifest;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -55,6 +56,7 @@ import static org.jetbrains.osgi.jps.OsgiJpsBundle.message;
  * @author <a href="mailto:janthomae@janthomae.de">Jan Thomä</a>
  */
 public class BndWrapper {
+  private static final Logger LOG = Logger.getInstance(BndWrapper.class);
   private final Reporter myReporter;
 
   public BndWrapper(Reporter reporter) {
@@ -190,6 +192,45 @@ public class BndWrapper {
       builder.setPedantic(false);
       builder.setClasspath(classPath);
       builder.setSourcepath(srcPath);
+      doBuild(builder, outputFile);
+    }
+  }
+
+  public void build(@Nullable File bndFile,
+                    @NotNull Map<String, String> properties,
+                    @NotNull Properties mavenProperties,
+                    File @NotNull [] classPath,
+                    File @NotNull [] srcPath,
+                    @Nullable File base,
+                    @NotNull File moduleOutputDir,
+                    @NotNull File outputFile) throws Exception {
+    try (Builder builder = new ReportingBuilder(myReporter, new Processor(mavenProperties, false))) {
+      if (LOG.isDebugEnabled()) {
+        builder.setTrace(true);
+        LOG.debug("OSGi bundle bnd file: " + bndFile + ", properties: " + properties + ", Maven properties: " + mavenProperties);
+      }
+
+      builder.setProperties(base, OrderedProperties.fromMap(properties));
+      builder.setPedantic(false);
+      builder.setClasspath(classPath);
+      builder.setSourcepath(srcPath);
+
+      if (bndFile != null) {
+        // implicitly calls setBase
+        builder.setProperties(bndFile);
+      }
+      else {
+        builder.setBase(base != null ? base : new File(""));
+      }
+
+      // pack everything of the output directory, just like in maven-jar-plugin
+      // includes/filtering/etc. needs to be done elsewhere
+      if (moduleOutputDir.isDirectory()) {
+        Jar classesDirJar = new Jar("classes", moduleOutputDir);
+        classesDirJar.setManifest(new Manifest());
+        builder.setJar(classesDirJar);
+      }
+
       doBuild(builder, outputFile);
     }
   }
--- a/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/OsgiBuildSession.java
+++ b/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/OsgiBuildSession.java
@@ -158,6 +158,46 @@ public class OsgiBuildSession implements Reporter {
       }
       mySourceToReport = null;
     }
+    else if (myExtension.isUseBndMavenPlugin()) {
+      File bndFile = myExtension.getBundleDescriptorFile();
+      File base = null;
+      if (bndFile != null && bndFile.isFile()) {
+        // explicitly specified bndfile that exists
+        mySourceToReport = bndFile.getAbsolutePath();
+      }
+      else if (StringUtil.isNotEmpty(myExtension.getBndFileLocation())) {
+        // explicitly specified bndfile that does not exist - fail to report a build error
+        throw new OsgiBuildException(message("session.bnd.missing", myExtension.getBndFileLocation()));
+      }
+      else {
+        // no bndfile specified, try to fallback to the default bnd.bnd file next to the pom.xml
+        File mavenProjectPath = OsgiBuildUtil.getMavenProjectPath(myContext, myModule);
+        if (mavenProjectPath != null) {
+          bndFile = new File(mavenProjectPath.getParentFile(), "bnd.bnd");
+          if (bndFile.isFile()) {
+            mySourceToReport = bndFile.getAbsolutePath();
+          }
+          else{
+            bndFile = null;
+            // default not found, use the pom.xml location as the base and error source
+            mySourceToReport = mavenProjectPath.getAbsolutePath();
+            base = mavenProjectPath.getParentFile();
+          }
+        }
+      }
+
+      Properties mavenProperties = OsgiBuildUtil.getMavenProjectProperties(myContext, myModule);
+      try {
+        myBndWrapper.build(bndFile, myExtension.getAdditionalProperties(), mavenProperties,
+                           myClasses, mySources, base, myModuleOutputDir, myOutputJarFile);
+      }
+      catch (Exception e) {
+        throw new OsgiBuildException(message("session.unknown.error"), e, null);
+      }
+      finally {
+        mySourceToReport = null;
+      }
+    }
     else if (myExtension.isUseBundlorFile()) {
       String bundlorPath = myExtension.getBundlorFileLocation();
       File bundlorFile = OsgiBuildUtil.findFileInModuleContentRoots(myModule, bundlorPath);
--- a/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/OsgiBuildSession.java
+++ b/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/OsgiBuildSession.java
@@ -172,17 +172,19 @@ public class OsgiBuildSession implements Reporter {
       else {
         // no bndfile specified, try to fallback to the default bnd.bnd file next to the pom.xml
         File mavenProjectPath = OsgiBuildUtil.getMavenProjectPath(myContext, myModule);
-        if (mavenProjectPath != null) {
-          bndFile = new File(mavenProjectPath.getParentFile(), "bnd.bnd");
-          if (bndFile.isFile()) {
-            mySourceToReport = bndFile.getAbsolutePath();
-          }
-          else{
-            bndFile = null;
-            // default not found, use the pom.xml location as the base and error source
-            mySourceToReport = mavenProjectPath.getAbsolutePath();
-            base = mavenProjectPath.getParentFile();
-          }
+        if (mavenProjectPath == null) {
+          throw new OsgiBuildException(message("session.pom.missing"));
+        }
+
+        bndFile = new File(mavenProjectPath.getParentFile(), "bnd.bnd");
+        if (bndFile.isFile()) {
+          mySourceToReport = bndFile.getAbsolutePath();
+        }
+        else{
+          bndFile = null;
+          // default not found, use the pom.xml location as the base and error source
+          mySourceToReport = mavenProjectPath.getAbsolutePath();
+          base = mavenProjectPath.getParentFile();
         }
       }
 
--- a/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/ReportingBuilder.java
+++ b/osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/ReportingBuilder.java
@@ -25,6 +25,7 @@
 package org.jetbrains.osgi.jps.build;
 
 import aQute.bnd.osgi.Builder;
+import aQute.bnd.osgi.Processor;
 import com.intellij.openapi.diagnostic.Logger;
 import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
@@ -39,7 +40,7 @@ public class ReportingBuilder extends Builder {
     myReporter = reporter;
   }
 
-  public ReportingBuilder(@NotNull Reporter reporter, @NotNull Builder parent) {
+  public ReportingBuilder(@NotNull Reporter reporter, @NotNull Processor parent) {
     super(parent);
     myReporter = reporter;
     use(parent);
--- a/osmorc/src/org/osmorc/facet/OsmorcFacetConfiguration.java
+++ b/osmorc/src/org/osmorc/facet/OsmorcFacetConfiguration.java
@@ -26,6 +26,7 @@ package org.osmorc.facet;
 
 import aQute.bnd.header.OSGiHeader;
 import aQute.bnd.header.Parameters;
+import com.intellij.compiler.server.BuildManager;
 import com.intellij.facet.FacetConfiguration;
 import com.intellij.facet.ui.FacetEditorContext;
 import com.intellij.facet.ui.FacetEditorTab;
@@ -300,6 +301,10 @@ public final class OsmorcFacetConfiguration implements FacetConfiguration, Modif
 
   public boolean isUseBndFile() {
     return getManifestGenerationMode() == ManifestGenerationMode.Bnd;
+  }
+
+  public boolean isUseBndMavenPlugin() {
+    return getManifestGenerationMode() == ManifestGenerationMode.BndMavenPlugin;
   }
 
   /**
--- a/osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
+++ b/osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) 2007-2009, Osmorc Development Team
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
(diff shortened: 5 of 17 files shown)


ibauersachs:
@trespasserw you're assigned to this bug in Youtrack.

trespasserw:
Thanks for the patch!

Tell please, is it BndLib 5.2.0 critically important for the feature? If it isn't, I'd rather leave it at 5.1.2 for now.

ibauersachs:
I don't think it is, but I haven't tried with the old version, which was at 5.0.0 (not 5.1.2).

trespasserw:
Right, it was 5.0.0 back then, but it was bumped to 5.1.2 just yesterday (which now creates a conflict). If 5.2.0 isn't critical for the PR, then please drop b163139 and I'll try to review the rest early next week (or maybe even later today).

ibauersachs:
I'll look at it over the weekend and rebase. The two commits are really independent, you should be able to review or comment anyway.

ibauersachs:
@trespasserw Any update?

trespasserw:
Not yet, sorry :(

trespasserw:
Have a conceptual question. Is there a need to add another UI option and execution branches to handle it - i.e. why not just set the manifest generation mode to `Bnd` in `OsmorcBndFacetImporter`? I see that there are differences in build routines, but can't estimate how important are they.

ibauersachs:
I thought about doing that, but the important difference is that there can be additional instructions in the (parent-) pom in addition to the bnd file. The "pure" Bnd-mode would also require instructions to include resources in the final jar, which would make it radically different to the Maven handling.

The Maven plugin, when executed by Maven, only generates the manifest and the maven-jar-plugin can pick that up. So a jar should actually not be created (there's another, outdated pull request, #663 that did this).
But without creating a jar all other OSGi handling (the framework startup etc.) would not work at all. So I thought the best option would be to simply create the jar anyway, but automatically with all content. This makes it work cleanly when Kotlin is in the game, contrary to the "old" maven-bundle-plugin that requires pulling stunts to include the `*.kotlin_module`.

trespasserw:
I see.

Anyway, the option doesn't make much sense UI-wise - it is set by the importer, not by a user. Let's keep `ManifestGenerationMode.BndMavenPlugin`, but threat it the same as `ManifestGenerationMode.Bnd` in the UI (this will make `OsmorcFacetGeneralEditorTab#apply` a bit more complex, but not much).

ibauersachs:
The option currently also blocks editing in the other two tabs. If he importer sets it to `Use Bnd and ...` but in the background sets `ManifestGenerationMode.BndMavenPlugin`, how can that be removed/changed to regular Bnd when the Maven plugin is removed from the pom, or when `Do not synchronize` is set? Is it an acceptable workflow that the facet needs to be removed/readded for such a case?

trespasserw:
It does block the tabs, but so does the plain `Bnd` option (in fact, you can just reuse the existing key).

There was an issue about automatically dropping corresponding facets when plugins are removed from a Maven project, but I don't remember whether it is fixed. But there is no need in removing/re-adding the facet, one can just choose some other option, press "OK" or "Apply" (depending on the implementation), then choose the option again.

Alternatively, you may add a boolean property to the facet [instead of the new enum constant] and present it as a checkbox in the UI (I'd also make it uneditable unless "Do not synchronize ..." is on).

ibauersachs:
By checkbox you mean something like this?

image

trespasserw:
Something like this, right.

ibauersachs:
I'll look at it, might take me a couple of days.

Apart from the UI / config transfer, is there anything else that needs attention?

trespasserw:
Yep. I writing my concerns as review comments, probably they'll become visible when I finish the review (in an hour, unless something really terrible happens :)

osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/BndWrapper.java
```diff
@@ -194,6 +196,45 @@ public void build(@NotNull Map<String, String> properties, File @NotNull [] clas
...
+                    @NotNull Properties mavenProperties,
+                    File @NotNull [] classPath,
+                    File @NotNull [] srcPath,
+                    @Nullable File base,
+                    @NotNull File moduleOutputDir,
+                    @NotNull File outputFile) throws Exception {
+    try (Builder builder = new ReportingBuilder(myReporter, new Processor(mavenProperties, false))) {
+      if (LOG.isDebugEnabled()) {
+        builder.setTrace(true);
+        LOG.debug("OSGi bundle bnd file: " + bndFile + ", properties: " + properties + ", Maven properties: " + mavenProperties);
```
trespasserw, line 210: `mavenProperties` is used only for logging - is that correct?

trespasserw (changes requested):

osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/BndWrapper.java
ibauersachs: They're passed to the constructor of `Processor` on line 207.

osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/ReportingBuilder.java
```diff
@@ -39,7 +40,7 @@ public ReportingBuilder(@NotNull Reporter reporter) {
     myReporter = reporter;
   }
 
-  public ReportingBuilder(@NotNull Reporter reporter, @NotNull Builder parent) {
+  public ReportingBuilder(@NotNull Reporter reporter, @NotNull Processor parent) {
```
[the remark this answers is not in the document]
ibauersachs: For BndWrapper.java#207

osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
```diff
@@ -0,0 +1,200 @@
...
+        UTF8Properties properties = new UTF8Properties();
+        properties.load(bndElement.getValue(), pomFile, null);
+        for (Map.Entry<?, ?> e : properties.replaceHere(pomFile.getParentFile()).entrySet()) {
+          props.put(e.getKey().toString(), e.getValue().toString());
+        }
+      }
+    }
+
+    // <bndfile> element
+    String bndFileName = DEFAULT_BND_FILENAME;
```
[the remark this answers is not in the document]
ibauersachs: I need to dive a bit deeper into the logic again to see what I can do about the import vs. build phases.

I'm not sure I understand the part with the subdirectory. If the bnd file is not next to the pom, then `<bndfile>` must be specified. Is `findChild(...)` not handling paths?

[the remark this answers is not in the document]
trespasserw: No, `findChild` doesn't, it's what `findFileByRelativePath` is for. But I'd not use VFS here anyway, `j.i.File` would suffice.

ibauersachs:
I'm trying to implement the auto-pickup of the bnd file if present. This is easily extended based on what you did back in 2015 ([IDEA-140496](https://youtrack.jetbrains.com/issue/IDEA-140496), 1c778358861ad9bcea0eea1085775e7849578159) if the file is added or it's contents changed. However, the logic in `org.jetbrains.osgi.jps.build.OsmorcBuildTarget#computeRootDescriptors` does not work if the file is renamed or deleted because then `extension.getBundleDescriptorFile()` just returns `null` and the `DirtyFilesHolder` passed to `OsmorcBuilder` knows nothing about a changed or removed file.
This affects not just the changes I'm trying to implement, but also the existing options. What's the best option to address this? Tracking all content roots as a `BuildRootDescriptor`?

trespasserw:
Must admit I didn't touch that code for a while and have rather vague memories of how it works :( If I got it right, you need to add a branch to the condition in `OsmorcBuildTarget#computeRootDescriptors` and add a descriptor for implicit `bnd.bnd` if a module is governed by `bnd-maven-plugin` (or maybe do the same in `JpsOsmorcModuleExtensionImpl#myDescriptorFile` calculation).

Re file renaming/deletion - I'm not sure I understand what's the problem is, and since you're saying it is not limited to this case I'd leave it for later.

ibauersachs:
I've updated the the PR with
- The checkbox vs. radio button in the UI as shown below
- The default bnd file is not set as the config file during import
- The default bnd file is picked up at build if created without re-importing
- Changes in the bnd file (default or configured with bndfile-tag) are picked up correctly without re-importing

image

While attempting builds, I did some small, but important, bugfixes that I could separate if necessary:
- For peformance optimization, the build system preloads a ready to use build process in the background. That process does not pick up facet config changes automatically and needs to be reset, just like when changing the Java or Kotlin language level (`org.osmorc.facet.OsmorcFacetConfiguration#writeExternal`)
- The incremental compiler did not detect facet configuration changes (`org.jetbrains.osgi.jps.build.OsmorcBuildTarget#writeConfiguration`)
- The Manifest Generation tab did not reset the changed state after apply, resulting in an always active apply button (`org.osmorc.facet.ui.OsmorcFacetManifestGenerationEditorTab#apply`)
- The facet UI shows only one validation error. If the OSGi core library is not on the classpath, that warning always came first and any real configuration error, e.g. an invalid bnd file path, was never visible (`org.osmorc.facet.OsmorcFacetConfiguration#createEditorTabs`)
- Bnd by default sets the last-modified time of the created jar to the time of the bnd-file (if present). If the jar was built because of a  class change or Maven re-import, the date was still in the past. This is confusing and e.g. IntelliJ with the Archive Browser plugin would not pick up the changes (`org.jetbrains.osgi.jps.build.BndWrapper#doBuild`)

osmorc/src/org/osmorc/facet/OsmorcFacetConfiguration.java
```diff
@@ -278,6 +283,13 @@ public void writeExternal(Element element) throws WriteExternalException {
       additionalJARContentsElement.addContent(entry);
     }
     element.addContent(additionalJARContentsElement);
+
+    // reset the background build process to pick up facet configuration changes
+    if (myModificationCount.get() > 0) {
+      BuildManager.getInstance().clearState(myFacet.getModule().getProject());
+    }
+
+    myModificationCount.set(0);
```
trespasserw, line 292: The value is used for cache invalidation and is expected to increase monotonically; resets may cause hard-to-reproduce problems.

osmorc/src/org/osmorc/facet/OsmorcFacetConfiguration.java
```diff
@@ -278,6 +283,13 @@ public void writeExternal(Element element) throws WriteExternalException {
       additionalJARContentsElement.addContent(entry);
     }
     element.addContent(additionalJARContentsElement);
+
+    // reset the background build process to pick up facet configuration changes
+    if (myModificationCount.get() > 0) {
+      BuildManager.getInstance().clearState(myFacet.getModule().getProject());
```
trespasserw, line 289: The call is correct, but it is better to move it into tabs' `apply()` methods.

trespasserw (changes requested):

trespasserw:
That "missing core" warning... It was added ages ago and looks rather useless to me - after all, many bundles don't care whether they're loaded by a framework or a bare JVM. What do you think, would it be missed if I just let it go?

As for the bugfixes - if you don't mind, I will commit them separately (this repository is a mirror, so Github's "Merge" isn't applicable anyway).

osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
```diff
@@ -0,0 +1,192 @@
...
+    if (configuration != null) {
+      Element bndElement = configuration.getChild("bnd");
+      if (bndElement != null) {
+        LOG.debug("Using headers from bnd-element in " + mavenProject.getPath());
+        File pomFile = new File(pomVirtualFile.getPath());
+        UTF8Properties properties = new UTF8Properties();
+        properties.load(bndElement.getValue(), pomFile, null);
+        for (Map.Entry<?, ?> e : properties.replaceHere(pomFile.getParentFile()).entrySet()) {
+          props.put(e.getKey().toString(), e.getValue().toString());
+        }
```
trespasserw, line 175: There's [one thing](https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/README.md#configuration-parameters) with these properties:
> If the projects has a `bndfile` configuration property or a file in the default location `bnd.bnd`, then this configuration element is ignored.

The first case is easy to address, but that default location is tricky and probably doesn't worth bothering.

ibauersachs:
> That "missing core" warning... It was added ages ago and looks rather useless to me - after all, many bundles don't care whether they're loaded by a framework or a bare JVM. What do you think, would it be missed if I just let it go?

I thought of removing it as well but didn't want to change too much. To me, that check makes absolutely no sense, a module can be a perfectly valid bundle without that dependency. It's only needed to use OSGi interfaces, e.g. a `BundleActivator` or `ServiceReference<T>`.

> As for the bugfixes - if you don't mind, I will commit them separately (this repository is a mirror, so Github's "Merge" isn't applicable anyway).

Sure, please go ahead. I'll look at the other comments later and rebase.

trespasserw:
_For peformance optimization, the build system preloads a ready to use build process in the background. That process does not pick up facet config changes automatically and needs to be reset, just like when changing the Java or Kotlin language level (`org.osmorc.facet.OsmorcFacetConfiguration#writeExternal`)_ - **obsolete** (implemented in tabs' `apply()` methods).

_The incremental compiler did not detect facet configuration changes (`org.jetbrains.osgi.jps.build.OsmorcBuildTarget#writeConfiguration`)_ - **merged**.

_The Manifest Generation tab did not reset the changed state after apply, resulting in an always active apply button (`org.osmorc.facet.ui.OsmorcFacetManifestGenerationEditorTab#apply`)_ - **merged**.

_The facet UI shows only one validation error. If the OSGi core library is not on the classpath, that warning always came first and any real configuration error, e.g. an invalid bnd file path, was never visible (`org.osmorc.facet.OsmorcFacetConfiguration#createEditorTabs`)_ - **obsolete** (the "Missing OSGi Core" validator is gone).

_Bnd by default sets the last-modified time of the created jar to the time of the bnd-file (if present). If the jar was built because of a class change or Maven re-import, the date was still in the past. This is confusing and e.g. IntelliJ with the Archive Browser plugin would not pick up the changes (`org.jetbrains.osgi.jps.build.BndWrapper#doBuild`)_ - **merged**.

ibauersachs:
> _For peformance optimization, the build system preloads a ready to use build process in the background. That process does not pick up facet config changes automatically and needs to be reset, just like when changing the Java or Kotlin language level (`org.osmorc.facet.OsmorcFacetConfiguration#writeExternal`)_ - **obsolete** (implemented in tabs' `apply()` methods).

The facet import also needs to reset the background process (regardless of which Maven plugin is used). How should that be handled? From within the facet importers?

trespasserw:
I believe resetting the build process on import from Maven happens automatically, because the whole project model is rebuild. Will check.

osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
ibauersachs: I'm not too savvy with how IntelliJ mirrors the Maven model, but in the end I think it's impossible to exactly mirror the plugins' behavior. Mainly because the plugin traverses the pom hierarchy and parses each pom.xml individually, while IntelliJ presses all POMs into one flat model.

I can of course skip reading the properties here if a `<bndfile>` tag is found, but then it all instructions from `<bnd>` tag from a parent pom are also ignored because I can't distinguish where the properties came from.

I'd favor keeping it as is because if a bnd-file is present, regardless of via tag or default, the instructions in there take priority over the ones in the `<bnd>` tag. For example:

```
parent
|-- pom.xml:
|    <bnd>Bundle-DocUrl: https://example.org/parent</bnd>
|-- child
|    |--pom.xml:
|    |   <bnd>Bundle-DocUrl: https://example.org/child/from-tag</bnd>
|    |--bnd.bnd
|    |   Bundle-DocUrl: https://example.org/child/from-file
```

results in a manifest with `Bundle-DocUrl: https://example.org/child/from-file`. This is consistent with the Maven plugin.

However, in IntelliJ the `<bnd>` tag in the child overrides all instructions from the parent, while in Maven such a configuration is ignored and the parent is applied. As this is an invalid configuration per the plugin spec, I'm not sure if or how to deal with it given IntelliJ's simpliefied Maven model.

ibauersachs:
> I believe resetting the build process on import from Maven happens automatically, because the whole project model is rebuild. Will check.
> 
> Upd: right, any meaningful change to the Maven file triggers either a "roots changed" event or a bytecode target update (or both).

What counts as "meaningful change"? If I change a value inside the `<bnd>` tag and reimport, the change is not recognized anymore.

ibauersachs:
@trespasserw ping...?

ibauersachs:
@trespasserw I assume you're busy, but I'd appreciate it if you could give a hint about to move this forward.

trespasserw:
Terribly sorry :( Will definitely look at it later today.

osmorc/osmorc-jps-plugin/src/org/jetbrains/osgi/jps/build/OsgiBuildSession.java
```diff
@@ -158,6 +158,46 @@ private void doBuild() throws OsgiBuildException {
...
+        mySourceToReport = bndFile.getAbsolutePath();
+      }
+      else if (StringUtil.isNotEmpty(myExtension.getBndFileLocation())) {
+        // explicitly specified bndfile that does not exist - fail to report a build error
+        throw new OsgiBuildException(message("session.bnd.missing", myExtension.getBndFileLocation()));
+      }
+      else {
+        // no bndfile specified, try to fallback to the default bnd.bnd file next to the pom.xml
+        File mavenProjectPath = OsgiBuildUtil.getMavenProjectPath(myContext, myModule);
+        if (mavenProjectPath != null) {
```
trespasserw, line 175: I'd throw a build exception - this is not supposed to happen.

ibauersachs: Done

osmorc/src/org/osmorc/maven/facet/OsmorcBndFacetImporter.java
trespasserw: I see, thanks. Okay, I believe it doesn't make sense to invest great efforts in exact replication of the plugin's behavior - standard cases seem to be covered, and for the rest there's a possibility to delegate build to Maven.

trespasserw (approved):

trespasserw:
> What counts as "meaningful change"? If I change a value inside the <bnd> tag and reimport, the change is not recognized anymore.

I'll check this again, but IIRC any change outside of comments and whitespaces.

trespasserw:
In my experiments, a change inside `//configuration/bnd` does lead to `BuildManager#cancelPreloadedBuilds` invocation - though for the change to be registered one has to reimport the project first (there's a small "reimport Maven changes" popup appearing in the top right corner of the editor).

ibauersachs:
I did the Maven reimport after the pom change, but let me check again. Anyway, this is a separate issue.

ibauersachs:
Seems okay now, sorry about that. I probably ran into glitches with either the [Archive Browser plugin](https://plugins.jetbrains.com/plugin/9491-archive-browser), the outdated timestamp from bnd or Layer 8.

trespasserw:
Right. Okay, then merging. Thanks for the patience!

PR #791, opened 2021-02-17
Update the Dart problem hover tooltips to include the additional context messages

jwren:
@alexander-doroshko 

See image, but ignore the change showing the text of the URL:

new_problem_hover

The new change is that the text "Variable 'i' could be null due to..."

This is an initial cut.  Is it possible with a URL in HTML blocks in IJ to have the location metadata link to a location in source?  You'll see in the PR that we have the offset and full file path for the intervening write.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
@@ -525,12 +525,14 @@ public final class DartServerData {
     private final @Nullable @NonNls String myCode;
     private final @NotNull @NlsSafe String myMessage;
     private final @Nullable @NlsSafe String myCorrection;
+    private final @NotNull List<DartDiagnosticMessage> myContextMessages;
     private final @Nullable @NonNls String myUrl;
 
     @Contract(pure = true)
     @NotNull
     public DartError asEofError(int fileLength) {
-      return new DartError(fileLength > 0 ? fileLength - 1 : 0, fileLength > 0 ? 1 : 0, mySeverity, myCode, myMessage, myCorrection, myUrl);
+      return new DartError(fileLength > 0 ? fileLength - 1 : 0, fileLength > 0 ? 1 : 0, mySeverity, myCode, myMessage, myCorrection,
+                           myContextMessages, myUrl);
     }
 
     private DartError(@NotNull final AnalysisError error, final int correctedOffset, final int correctedLength) {
@@ -539,6 +541,7 @@ public final class DartServerData {
       myCode = error.getCode() == null ? null : error.getCode().intern();
       myMessage = error.getMessage();
       myCorrection = error.getCorrection();
+      myContextMessages = DartDiagnosticMessage.generateDartDiagnosticMessageList(error.getContextMessages());
       myUrl = error.getUrl();
     }
 
@@ -548,12 +551,14 @@ public final class DartServerData {
                       @Nullable @NonNls String code,
                       @NotNull @Nls String message,
                       @Nullable @Nls String correction,
+                      @NotNull List<DartDiagnosticMessage> contextMessages,
                       @Nullable @NonNls String url) {
       super(offset, length);
       mySeverity = severity;
       myCode = code;
       myMessage = message;
       myCorrection = correction;
+      myContextMessages = contextMessages;
       myUrl = url;
     }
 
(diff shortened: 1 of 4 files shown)


jwren:
@devoncarew @stereotype441

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
```diff
@@ -167,10 +169,16 @@ private void ensureInitialized() {
...
 
   public static @NotNull @Nls String generateTooltipText(@NotNull @Nls String message,
+                                                         @NotNull List<DartServerData.DartDiagnosticMessage> contextMessages,
                                                          @Nullable @Nls String correction,
                                                          @Nullable @NonNls String url) {
-
     HtmlBuilder htmlBuilder = new HtmlBuilder().append(message);
+    if (!contextMessages.isEmpty()) {
+      for (DartServerData.DartDiagnosticMessage contextMessage : contextMessages) {
+        htmlBuilder.append(HtmlChunk.br()).append(HtmlChunk.br()).append(contextMessage.getMessage());
```
devoncarew, line 178: I think this is going to be turned into a hyperlink, but for the text of the link, instead of `(main.dart)`, could we do `main.dart:111`. I.e., file:line.

jwren: I had that thought as well, but I am not sure if it is possible/pragmatic with the IJ APIs and what we have sent with the analysis server.  The `DiagnosticMessage`s don't include the line number, but do contain the `offset` and `length`.  For open files where the PSIs have been created, i.e. we have a `Document` object, we could use the `Document.getLineNumber(int offset)` method, but this could lead to lags in generating the hover tooltips.

I figured having `file.dart:<offset>` would be misleading.

@alexander-doroshko Is there a way to quickly convert an offset to line number without the PSI construction?

alexander-doroshko: >  Is there a way to quickly convert an offset to line number without the PSI construction?

Either via Document, or via FileOffsetsManager. The latter is the right choice if the offset comes from the server (it contains the required data but may lack public API for this task). Both don't parse files (don't create PSI) but both do load file contents (at least once). I think this is fast enough.

alexander-doroshko:
Looks good to me. Agree, a link to the location would be great. I don't remember if such links are already supported. I'll take a closer look tomorrow. If not, we are free to add it to the IntelliJ Platform.

Dart/src/com/jetbrains/lang/dart/ide/errorTreeView/DartProblem.java
jwren: Thanks Alex.  I updated the PR to use the `Document` API, the numbers out of the `FileOffsetsManager` were all offsets, not line numbers.  LMK if there are any changes you want before we land it.

alexander-doroshko: `Document` API assumes all line breaks are single-character. Offsets given by server count real line break characters, as they are on disk. So, Document API may give incorrect results for CR-LF line endings.

That's why we wrote `FileOffsetsManager` for Dart. 

Sorry, I wasn't clear. I mean, `FileOffsetsManager` has enough data (in its fields) to calculate `offset-> line number` correctly. But you are right, it doesn't have such public method yet. If we want to show line number correctly we need to add it.

jwren: https://github.com/JetBrains/intellij-community/pull/1516

alexander-doroshko: Thanks! I'll merge both tomorrow.

alexander-doroshko: @jwren I just realized that we have `DiagnosticMessage.getLocation().getStartLine()` method. And we already se it in `DartProblemsViewPanel.addDiagnosticMessageActions()`. So we don't need any manual conversion at all! =)

alexander-doroshko:
> location metadata link to a location in source? 

Looks like we've already discussed it in https://github.com/JetBrains/intellij-plugins/pull/733 and at the moment there's no such API. But we are free to add it eventually.

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
```diff
@@ -577,11 +582,46 @@ public boolean isError() {
...
+    public @NotNull List<DartDiagnosticMessage> getContextMessages() {
+      return myContextMessages;
+    }
+
     public @Nullable @NonNls String getUrl() {
       return myUrl;
     }
   }
 
+  public static final class DartDiagnosticMessage extends DartRegion {
```
alexander-doroshko, line 594: There are 2 main reasons why we have classes like `DartRegion` and its inheritors instead of using classes from `org.dartlang.analysis.server.protocol.*` package directly:
- we need converted offsets and don't want to think about conversion each time we use offsets, so we store converted offsets in `DartRegion` object right away
- we need to update offsets/lengths on typing so that we have reasonable data before updated data comes from server

Neither is crucial for `DiagnosticMessage` so I updated the PR to store `org.dartlang.analysis.server.protocol.DiagnosticMessage` right in the `DartError` class, and removed `DartDiagnosticMessage` class.

alexander-doroshko:
Pushed with minor updates, thanks.
image

Dart/src/com/jetbrains/lang/dart/analyzer/DartServerData.java
jwren: Thanks Alex, I thought I checked for the start line, but missed it.  Thanks for merging.

PR #792, opened 2021-02-18
Update the Dart parser to allow triple-shift operators, and triple-shifts as a valid shift operator.

jwren:
The first of up to 3 parser changes: https://github.com/dart-lang/sdk/issues/44918


jwren:
@alexander-doroshko

alexander-doroshko:
Thanks! LGTM.

Could you add a test (`DartParsingTest`)? Updating an existing file would be enough (say, `Operators.dart`). Having something like `<List<List<List>>>`  in a test would be great - to make sure parser can distinguish generics vs operator. Symbol literal (`#>>>`) is also a good case for parsing test.

Also, I guess, new operator should be added to a couple of token sets in `DartTokenTypesSets`.

jwren:
@alexander-doroshko PTAL, from my commit: Add >>> tests into DartParsingTest, and updated DartTokenTypesSets to include >>>, note that I also added >> where it was missing.

alexander-doroshko:
Great! Thanks, merged.

PR #796, opened 2021-03-29
Remove the isFailure option in the Dart test event framework

jwren:
https://github.com/dart-lang/test/issues/1465


jwren:
@alexander-doroshko

alexander-doroshko:
Thanks. I posted my comments in dart-lang/test#1465. If Dart Test framework maintainers decide to remove the flag I'd suggest to remove not only the 'if' but the whole line completely from the Dart and Flutter plugins:
```
if (!getBoolean(obj, JSON_IS_FAILURE)) testError.addAttribute("error", "true");
```
In this case all failures will look like test failures. 

(If we keep `testError.addAttribute("error", "true")` then all failures will look like potential test code problems.)

Dart/src/com/jetbrains/lang/dart/ide/runner/test/DartTestEventsConverter.java
```diff
@@ -323,7 +322,7 @@ private boolean handleError(JsonObject obj) throws ParseException {
         }
       }
 
-      if (!getBoolean(obj, JSON_IS_FAILURE)) testError.addAttribute("error", "true");
+      testError.addAttribute("error", "true");
```
alexander-doroshko, line 325: As discussed, I removed this line.

alexander-doroshko:
Assuming that the "isFailure" flag is going to be removed, so I've merged this PR.
@jwren Please make sure to delete [this line](https://github.com/flutter/flutter-intellij/pull/5345/files#diff-9fe5010cea9dbc123c19766c4cb9f0e0ac7c51cee84993ab8730a012a8c36332R351) in the Flutter plugin.

PR #797, opened 2021-03-31
Add a bundled dictionary provider for Dart

parlough:
Introduce a bundled dictionary provider for Dart to prevent the typo inspection from triggering on many common words used in the Dart ecosystem. 

The included list is in no way exhaustive, but it includes ones I've hit in the past and I commonly found in a few Dart projects.

To keep the spelling functionality together, I created a new `spelling` package to consolidate it.


alexander-doroshko:
Merged, thanks for the contribution.

PR #799, opened 2021-04-02
Update dartfmt strings to the Dart formatter

parlough:
Open to suggestions here on the exact changes to the strings, but the name `dartfmt` is going to be less common as it has been replaced with `dart format` in 2.10.4. 

Refer to https://github.com/dart-lang/site-www/issues/3111#issuecomment-810843455 for context


alexander-doroshko:
Thanks! I'll take a pause to discuss the wording internally.

PR #801, opened 2021-05-12
add emacs tab support for makefile

foriequal0:
Original Emacs also simply inserts a tab character for Makefile.

Related: https://youtrack.jetbrains.com/issue/CPP-24135

(I've closed previous PR by mistake https://github.com/JetBrains/intellij-plugins/pull/800)


foriequal0:
I found that Shell script plugin also need a `EmacsProcessingHandler`.
I think it would be better `DefaultEmacsProcessingHandler` to insert a tab character like this...

tjarvstrand:
Any news on this?

PR #802, opened 2021-05-13
Fix set-over async call in the Dart VM debugger

jwren:
Fix from https://github.com/flutter/flutter-intellij/pull/5476


jacob314 (approved):
lgtm

alexander-doroshko:
Thanks, merged.

PR #804, opened 2021-06-26
add white-space after step keyword for Korean

boorownie:
For natural Korean sentences, a space is needed at the end of the step keyword of Korean. 
Currently, if I put a space after the step keyword, it cannot be recognized as the step keyword.

## with space: cannot be recognized as the step keyword
image

## without space: can be recognized as the step keyword
image


avokin2:
Thank you.

Merged

PR #813, opened 2021-07-28
Update rule localization according to specification

SteelAlex:
Rule localization for russian language is wrong. Current value was "Rule", but should be "Правило". Link to spec: https://cucumber.io/docs/gherkin/languages/ 

This mislocalization cause syntax highlight issue: 

image

Here is the sample:

```
# language: ru
@framework
Функция: пример

  Правило: некорректная подстветка

  Rule: корректная подстветка, некорректный язык
```


avokin2:
Thank you. 

Merged

PR #814, opened 2021-08-07
IDEA-138550: Update dependencies for Error Prone plugin

cushon:
Upgrade to Error Prone 2.7.1, and fix Checker Framework Dataflow
dependency to use the correct artifact and version.

The latest version of Error Prone at its dependencies are documented at
http://errorprone.info/docs/installation.


anstarovoyt:
Thank you for your contribution to the IntelliJ Plugins, and sorry for not getting back to you sooner. 

The problem was fixed in https://youtrack.jetbrains.com/issue/IDEA-275519 (I guess the assignee didn't know about the PR).

PR #816, opened 2021-08-11
Update the Dart thirdPartySrc vmServiceDrivers sources

jwren:
Sources copied from https://github.com/dart-lang/sdk/tree/master/pkg/vm_service


jwren:
@alexander-doroshko

PR #822, opened 2021-09-15
Use getContainerClass method as a fallback when PsiElement couldn't resolve class context

bay73:
Synthetic methods are represented by PSI elements which don't have parent context and so class can't be obtained from the context. These elements implement PsiMember directly and have getContainerClass method to accessing the linked class.


alice-ks:
Nikita, we only identified this issue as we had a test which failed in 2021.2. As far as I know, those tests (e.g. [PbJavaGotoDeclarationHandlerTest](https://github.com/jvolkman/intellij-protobuf-editor/blob/master/java/src/itest/java/idea/plugin/protoeditor/java/PbJavaGotoDeclarationHandlerTest.java)) were removed from this repository. You might want to reconsider this as those tests turned out to be pretty helpful.

NikitaKatkov:
thank you! the fix is merged.

regarding removed test: it used some black magic and didn't conform to our tests code requirements. I'll file an issue about it and will create tests for this functionality

PR #823, opened 2021-09-22
Fix WEB-52638, Prettier plugin doesn't take into account prettierrc when detecting if a file can be formatted

MasonD:
Addresses https://youtrack.jetbrains.com/issue/WEB-52638.

I'm not sure if the generated files are meant to be included in each commit, so if they are supposed to be committed, someone will need to edit this pull request before merging.


urbanex:
Hi, can anyone tell why it is not being merged? I really need this functionality to work again :(

alexander-doroshko:
Sorry for delay, I will review shortly.

PR #825, opened 2021-10-05
Remove excessive logging

bay73:
When using the NameGeneratorContributor extension a lot of identical messages spamming log file. It seems safe to use debug level for this message.


NikitaKatkov:
merged

PR #828, opened 2021-10-07
New extension in the Dart plugin: completionTimerExtension

jwren:
This change allows downstream plugins to record the end-to-end time for Dart code completions to be displayed to end users.


alexander-doroshko:
Ok, merged. Do you also want to call `dartCompletionEnd()` or some other special method in `DartAnalysisServerService...completion_getSuggestions()...onError()` (line 1465) just to make sure that each `dartCompletionStart()` has its pair?

PR #829, opened 2021-10-08
Add 'Generate constructor with named arguments'

Xylez01:
Added a "Generate" option to generate a constructor with named arguments.

Example:

https://user-images.githubusercontent.com/418977/136557355-6c35322e-c0ae-4456-bd78-ee627bd662e8.mov

Could not find any tests for the actions. If there are any, please point me in the right direction.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateConstructorWithNamedArgumentsFix.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateConstructorWithNamedArgumentsFix.java
@@ -1,0 +1,77 @@
+// Copyright 2000-2021 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+package com.jetbrains.lang.dart.ide.generation;
+
+import com.intellij.codeInsight.template.Template;
+import com.intellij.codeInsight.template.TemplateManager;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.NlsContexts;
+import com.jetbrains.lang.dart.DartBundle;
+import com.jetbrains.lang.dart.psi.DartClass;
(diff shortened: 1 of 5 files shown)


Dart/src/com/jetbrains/lang/dart/ide/generation/CreateConstructorWithNamedArgumentsFix.java
```diff
@@ -0,0 +1,77 @@
...
+        template.addTextSegment(",");
+      }
+    }
+    template.addTextSegment(",});");
+    template.addEndVariable();
+    template.addTextSegment(" "); // trailing space is removed when auto-reformatting, but it helps to enter line break if needed
+    return template;
+  }
+
+  private static Boolean isNullable(DartComponent component) {
```
JordyLangen, line 68: Did not find a way to check if a field is nullable, thus implemented this very crude "isNullable" method. Any feedback on this is welcome.

asashour:
FYI, there is a request to do something similar from the server-side, https://github.com/dart-lang/sdk/issues/47389.

However, it is not planned to find all the usage of the constructor(s) and add a default value based on the user input. I hope this limitation can be handled from the IDE side.

alexander-doroshko:
@JordyLangen Thank you for your contribution.

Most of the smart code insight features in Dart are powered by the Analysis Server, a tool from the Dart SDK. IDE only does what the server asks it to do. This includes code highlighting, completion, quick fixes, intention actions, refactorings.

For historical reasons, some features still remain implemented right in the Dart plugin. Code generation actions are among them. Dart plugin knows Dart code worse than the Analysis Server, so these features work not as great as they should.

We are planning to move code generation features to Analysis Server as well. I hope we'll start this work in the nearest months. So I don't think we should work on new code generation features in the Dart plugin.

asashour:
I am not part of the team, but some CLs were accepted, have a look at the `CONTRIBUTING` page.

PR #836, opened 2021-11-09
Dart plugin: add the new Java API for the updated code completion DAS protocol

jwren:
API added in the Dart Analysis Server in this PR: https://dart-review.googlesource.com/c/sdk/+/215643

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -605,7 +605,8 @@ public class DartServerCompletionContributor extends CompletionContributor {
       null,
       null,
       null,
-      null);
+      null,
+      0);
   }
 
   private static Icon getBaseImage(Element element) {
--- a/Dart/thirdPartySrc/analysisServer/com/google/dart/server/GetSuggestionDetailsConsumer2.java
+++ b/Dart/thirdPartySrc/analysisServer/com/google/dart/server/GetSuggestionDetailsConsumer2.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2021, the Dart project authors.
+ *
+ * Licensed under the Eclipse Public License v1.0 (the "License"); you may not use this file except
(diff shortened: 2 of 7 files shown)


asashour:
May I ask how these files are generated?

I tried to run `generate_all` in Dart analyzer packages, but the output doesn't contain the relevant sources.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -605,7 +605,8 @@ private static CompletionSuggestion createCompletionSuggestionFromAvailableSugge
       null,
       null,
       null,
-      null);
+      null,
+      0);
```
alexander-doroshko, line 609: According to the `libraryUriToImportIndex` field docs, I think it's better to pass `null` here instead of `0`. Would you agree?

It doesn't matter in this particular case (as it's not used), just for the consistent use of the API.

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/c752cad8c4d8be485664a27a6bbc392bf7ac88ac

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
jwren: Yes, `null` is fine, but it also doesn't matter for the time being.

> May I ask how these files are generated?
> 
> I tried to run `generate_all` in Dart analyzer packages, but the output doesn't contain the relevant sources.

@asashour the consumers aren't generated, they are done manually.  The output from `generate_all` is the correct location.

asashour: @jwren  Thanks for your reply.

I guess that the comment about the other PR https://github.com/JetBrains/intellij-plugins/pull/832#issuecomment-942475198, that all files should not be changed is not precise.

I would raise a CL for dart generation, then we come to the `intellij-plugins` later.

PR #838, opened 2021-11-10
Dart plugin: add the ability to override the DAS protocol version in the DartAnalysisServerService.java

jwren:
This is part of the ongoing work for the new code completion implementation in the Dart Plugin.


jwren:
@alexander-doroshko

alexander-doroshko:
Is this for development purposes?
I mean, when developing the Dart plugin, we can run it from sources, hence add whatever hacks we want without committing them.
Will end users need to hack server version when using official Dart plugin builds?

PR #840, opened 2021-11-19
Dart plugin: work in progress towards updated completion support

jwren:
@alexander-doroshko 

This PR:
1) removes the `HTML` support as the angular plugin has been turned down,
2) introduces the usage of the new completion protocol if the returned DAS server version is `1.33` (currently higher than is committed into the Dart SDK)

Status: the new protocol works, the completions are re-queried only when `isIncomplete` is `true`.  What I can't yet figure out is how implement and decide to re-query for completions if `isIncomplete` is `true`, to give users access to more than just the fixed 100 that I have hard coded.  Any input here would be helpful Alex.

FYI @bwilkerson & @scheglov

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -732,6 +732,11 @@ public final class DartAnalysisServerService implements Disposable {
   }
 
   @NotNull
+  public String getServerVersion() {
+    return myServerVersion;
+  }
+
+  @NotNull
   public Project getProject() {
     return myProject;
   }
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
In progress...

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -1481,6 +1486,53 @@ public void onError(@NotNull final RequestError error) {
...
+        for (DartCompletionTimerExtension extension : DartCompletionTimerExtension.getExtensions()) {
+          extension.dartCompletionError(StringUtil.notNullize(error.getCode()), StringUtil.notNullize(error.getMessage()),
+                                        StringUtil.notNullize(error.getStackTrace()));
+        }
+        // Not a problem. Happens if a file is outside the project, or server is just not ready yet.
+        latch.countDown();
+      }
+    });
+
+    awaitForLatchCheckingCanceled(server, latch, GET_SUGGESTIONS_TIMEOUT);
```
alexander-doroshko, line 1527: I guess we can use a timeout longer than 5 seconds for this case. This is a ProcessCancelledException-protected code that runs in a background thread so we can afford a longer 'in progress' spinner in code completion popup if server might need more time to calculate and send results.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -99,93 +99,114 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
+                 if (completionInfo2.myIsIncomplete) {
+                   updatedResultSet.restartCompletionOnAnyPrefixChange();
                  }
 
-                 updatedResultSet.addElement(lookupElement);
-               }, (includedSet, includedKinds, includedRelevanceTags, libraryFilePathSD) -> {
-                 if (includedKinds.isEmpty()) {
-                   return;
+                 for (DartCompletionTimerExtension extension : DartCompletionTimerExtension.getExtensions()) {
+                   extension.dartCompletionEnd();
```
alexander-doroshko, line 131: If you don't mind, I'll move this right to the DartAnalysisServerService#completion_getSuggestions2() method, closer to `extension.dartCompletionStart()`. This will fix a possible code execution path when dartCompletionStart() is called but dartCompletionEnd() is not (notice `return` statement few lines above). It's easier to maintain code when both dartCompletionStart() and dartCompletionEnd() are in the same method.

Well, this way we'll not include lookup element creation in the gathered metrics but I guess it's <1% of the measured time.

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -99,93 +99,114 @@ protected void addCompletions(@NotNull final CompletionParameters parameters,
...
-                   }
+
+               if (!das.getServerVersion().isEmpty() && StringUtil.compareVersionNumbers(das.getServerVersion(), "1.33") >= 0) {
+                 final DartAnalysisServerService.CompletionInfo2 completionInfo2 = das.completion_getSuggestions2(file, offset, 100);
+                 if (completionInfo2 == null || completionInfo2.mySuggestions.isEmpty()) {
+                   return;
                  }
 
-                 LookupElementBuilder lookupElement = null;
+                 final CompletionResultSet updatedResultSet = resultSet;
```
alexander-doroshko, line 116: There are a few tricks that fix replacement offset and replacement length in the old completion provider implementation. Sometimes we calculate offsets ourselves, sometimes use info from server. If there are no known issues with the info from server, maybe let's try to always use replacement offset and replacement length from server?

alexander-doroshko:
Thanks, merged!

alexander-doroshko (approved):

Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
bwilkerson: The offset and length ought to be correct as long as the completions were computed from the same state of the file as the state in which completion is being offered. But if server's view of the file (at the time suggestions are computed) can ever be different from IntelliJ's (at the time suggestions are presented to the user), then the offset and/or length could be wrong.

jwren:
Thanks for reviewing and landing, the next thing I am trying to figure out is the best way to populate the next `100 + N` completions? `duringCompletion()`?  Other ideas?

PR #844, opened 2021-12-09
[Cucumber] Add extension point for configuring language injection prefixes and suffixes

danic-wd:
This addresses the issue described here: https://youtrack.jetbrains.com/issue/IDEA-284539
It's a small change but adds flexibility for plugins that depend on the gherkin plugin by allowing prefixes and suffixes to be configured for language injection.


avokin2:
Merged

PR #846, opened 2021-12-22
Dart plugin: implement and use the `completion_getSuggestionDetails2` protocol

jwren:
Dart plugin: implement and use the completion_getSuggestionDetails2 protocol, updates to the Dart completion tests for Dart SDK version `2.15` as well as the latest SDK HEAD version with the new protocol

`testWithImportPrefix` - modified due to change in completion support (or in the inclusion of constants in the `dart:io` package)

`testFunctionAfterShow` - disabled due to https://github.com/dart-lang/sdk/issues/47993

`testNotYetImportedClass` - test updated to match the current functionality of the Dart completion for not yet imported classes

These are failing with the updated DAS API version `1.33`:

`testIncompleteTernary` - test failing with new protocol, even though the DAS is returning correct result and manual testing of this test do work, not sure how to debug further.  Specifically, the lookup panel doesn't exist, and a NPE is being thrown.

`testConstructorParens`, `testDoNotEatAwaitOnTab`, `testNotYetImportedClass` - manual test looks fine to me, but framework is throwing errors with these tests, attempts to add additional pauses, etc, did not work

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -100,6 +100,7 @@ public final class DartAnalysisServerService implements Disposable {
   private static final long STATEMENT_COMPLETION_TIMEOUT = TimeUnit.MILLISECONDS.toMillis(100);
   private static final long GET_SUGGESTIONS_TIMEOUT = TimeUnit.SECONDS.toMillis(5);
   private static final long GET_SUGGESTION_DETAILS_TIMEOUT = TimeUnit.MILLISECONDS.toMillis(100);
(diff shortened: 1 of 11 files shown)


alexander-doroshko:
`testIncompleteTernary` fails because `true` appears to be the only item (after case-sensivity filtering) and hence it is auto-inserted without showing a completion popup.

The test will pass if we shorten the prefix to `tr` (currently it's `tru`).

Interesting, the old protocol gives 4 matching items for this case:
image

alexander-doroshko:
Thanks, pushed. 

testConstructorParens, testDoNotEatAwaitOnTab, and testSorting fail. I'll have a closer look later.

I tried to simplify code a bit. I removed 2 overloaded createLookupElement() methods and extracted common code for InsertHandlers. Please have a look when it lands.

PR #847, opened 2022-01-06
Sync the Dart plugin with the latest Dart analysis server API changes

jwren:
Including: completionMode, invocationCount and timeout on the completion_getSuggestions2 protocol as well as server_cancelRequest

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
@@ -196,7 +196,8 @@ public class DartServerCompletionContributor extends CompletionContributor {
       maxResults = (invocationCount - 1) * 1000;
     }
     DartAnalysisServerService das = DartAnalysisServerService.getInstance(project);
-    DartAnalysisServerService.CompletionInfo2 completionInfo2 = das.completion_getSuggestions2(file, startOffsetInHostFile, maxResults);
+    DartAnalysisServerService.CompletionInfo2 completionInfo2 =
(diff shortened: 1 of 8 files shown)


Dart/src/com/jetbrains/lang/dart/ide/completion/DartServerCompletionContributor.java
```diff
@@ -196,7 +196,8 @@ private static void handleCompletion2(@NotNull Project project,
       maxResults = (invocationCount - 1) * 1000;
     }
     DartAnalysisServerService das = DartAnalysisServerService.getInstance(project);
-    DartAnalysisServerService.CompletionInfo2 completionInfo2 = das.completion_getSuggestions2(file, startOffsetInHostFile, maxResults);
+    DartAnalysisServerService.CompletionInfo2 completionInfo2 =
+      das.completion_getSuggestions2(file, startOffsetInHostFile, maxResults, CompletionMode.BASIC, invocationCount);
```
alexander-doroshko, line 200: AnalysisServer API docs say that `invocationCount` starts from 1, but `com.intellij.codeInsight.completion.CompletionParameters#getInvocationCount()` returns 0 in case of autopopup. Should Analysis Server support 0 or should we use `min(1, invocationCount)` here?

jwren: Yes, I didn't include the fact that getInvocationCount can return 0.  If/when we choose to use the information in the analysis server we can update the documentation, and have this distinction between 0 and 1.  For the time being the DAS doesn't do anything with the input, hence I just passed the invocation count as it is.

PR #848, opened 2022-01-06
Update code completion in the Dart to include the display URI

jwren:


alexander-doroshko:
As discussed by email, `libraryUriToImport` doesn't always work great for display purposes.

PR #849, opened 2022-01-11
Enhanced Enum support in the Dart.bnf parser for Dart 2.17

jwren:
https://github.com/dart-lang/sdk/issues/47869

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/Dart.bnf
+++ b/Dart/src/com/jetbrains/lang/dart/Dart.bnf
@@ -251,14 +251,15 @@ classDefinition ::= metadata* 'abstract'? 'class' componentName typeParameters? 
 mixinDeclaration ::= metadata* 'mixin' componentName typeParameters? onMixins? interfaces? classBody
 {mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartPsiClass" implements="com.jetbrains.lang.dart.psi.DartClass"}
 
-enumDefinition ::= metadata* 'enum' componentName '{' enumConstantDeclaration (',' enumConstantDeclaration)* ','? '}'
-{pin=3  mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartPsiClass" implements="com.jetbrains.lang.dart.psi.DartClass"}
+enumDefinition ::= metadata* 'enum' componentName typeParameters? mixins? interfaces? '{' enumConstantDeclaration? (',' enumConstantDeclaration)* ','? ';'? classMemberDefinition* '}'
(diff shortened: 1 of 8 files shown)


Dart/src/com/jetbrains/lang/dart/Dart.bnf
```diff
@@ -251,14 +251,15 @@ classDefinition ::= metadata* 'abstract'? 'class' componentName typeParameters?
...
 {mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartPsiClass" implements="com.jetbrains.lang.dart.psi.DartClass"}
 
-enumDefinition ::= metadata* 'enum' componentName '{' enumConstantDeclaration (',' enumConstantDeclaration)* ','? '}'
-{pin=3  mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartPsiClass" implements="com.jetbrains.lang.dart.psi.DartClass"}
+enumDefinition ::= metadata* 'enum' componentName typeParameters? mixins? interfaces? '{' enumConstantDeclaration? (',' enumConstantDeclaration)* ','? ';'? classMemberDefinition* '}'
+ {pin=3  mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartPsiClass" implements="com.jetbrains.lang.dart.psi.DartClass"}
 
-enumConstantDeclaration ::= componentName
+enumConstantDeclaration ::= metadata* componentName typeArguments? arguments? |
+                            metadata* componentName typeArguments? '.' componentName arguments
```
alexander-doroshko, line 258: Any input matches `arguments?`, and this means that the second branch of this rule (after `|`) will never be reached. To make it work as expected, I flipped the branches. Test updated as well.
```
enumConstantDeclaration ::= metadata* componentName typeArguments? '.' componentName arguments |
                            metadata* componentName typeArguments? arguments?
```

Dart/src/com/jetbrains/lang/dart/Dart.bnf
```diff
@@ -251,14 +251,15 @@ classDefinition ::= metadata* 'abstract'? 'class' componentName typeParameters?
...
 {
   mixin="com.jetbrains.lang.dart.psi.impl.AbstractDartComponentImpl"
   implements="com.jetbrains.lang.dart.psi.DartComponent"
-  methods = [getMetadataList]
+  methods= [getMetadataList]
```
alexander-doroshko, line 262: With the updated grammar, we don't need custom getMetadataList() method declaration anymore (otherwize the generated code contains 2 of them and doesn't compile). But now we need custom `getComponentName` method declaration. Added it here and in `DartPsiImplUtil`, so now the generated code compiles fine.

alexander-doroshko:
Thanks, pushed.

PR #856, opened 2022-02-16
Lookup *.proto files in additional libraries

fkorotkov:


fkorotkov:
@NikitaKatkov I finally was able to run tests when I imported the protobuf modules into the `intellij-community` (looked at [Dart's documentation](https://github.com/JetBrains/intellij-plugins/blob/master/Dart/README.md#developing-the-plugin)). Not sure why ac4560c was necessary.

PR #860, opened 2022-03-01
Bug fix in Dart Analysis Server completion support, don't call completion_setSubscriptions() in startServer().

jwren:
This PR changes this call to happen in `serverConnected()`, where the das version is sent back to the client.


jwren:
@scheglov 
@alexander-doroshko

alexander-doroshko:
Good catch! Pushed, thanks.

PR #861, opened 2022-03-01
Call the Dart Analysis Server via `dart language-server`

jwren:
This is instead of `…dart .../analysis_server.dart.snapshot`, starting in Dart SDK version 2.16

@devoncarew
@pq 
@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -83,6 +83,10 @@ public final class DartAnalysisServerService implements Disposable {
   // https://github.com/dart-lang/webdev/blob/master/webdev/pubspec.yaml#L11
   public static final String MIN_WEBDEV_SDK_VERSION = "2.6.0";
 
+  // The dart cli command (https://dart.dev/tutorials/server/get-started) provides a language server command, `dart language-server` which
+  // should be used going forward instead of `dart .../analysis_server.dart.snapshot`.
+  public static final String MIN_DARTDEV_LANG_SERVER_SDK_VERSION = "2.16.0";
+
   private static final long UPDATE_FILES_TIMEOUT = 300;
 
   private static final long CHECK_CANCELLED_PERIOD = 10;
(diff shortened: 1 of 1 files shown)


Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2060,7 +2072,7 @@ else if (!dasSnapshotFile.canRead()) {
         vmArgsRaw = "";
       }
 
-      @NonNls String serverArgsRaw = "";
+      @NonNls String serverArgsRaw = useDartdevLangServerCall ? "--protocol=analyzer" : "";
       // Note that as of Dart 2.12.0 the '--useAnalysisHighlight2' flag is ignored (and is the
       // default highlighting mode). We still want to pass it in for earlier SDKs.
       serverArgsRaw += " --useAnalysisHighlight2";
```
devoncarew, line 2078: It's not necessary to pass this in for the 'language-server' command.

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -83,6 +83,10 @@
   // https://github.com/dart-lang/webdev/blob/master/webdev/pubspec.yaml#L11
   public static final String MIN_WEBDEV_SDK_VERSION = "2.6.0";
 
+  // The dart cli command (https://dart.dev/tutorials/server/get-started) provides a language server command, `dart language-server` which
+  // should be used going forward instead of `dart .../analysis_server.dart.snapshot`.
+  public static final String MIN_DARTDEV_LANG_SERVER_SDK_VERSION = "2.16.0";
```
devoncarew, line 88: I think this can be `2.14.4` (see https://github.com/Dart-Code/Dart-Code/blob/3fa9e24e0486785cab67e3d69e32e5026f50fca5/src/shared/capabilities/dart.ts#L23).

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Thanks, updated before pushing.

devoncarew:
> This is instead of `…dart .../analysis_server.dart.snapshot`, starting in Dart SDK version 2.16

Also - good to see this PR :)

Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
alexander-doroshko: Jaime says:
> the Dart SDK of 2.16 is a bit arbitrary, and I intentionally left support for older SDK as we have done in the past.

I personally have no opinion. I guess it doesn't matter.

jwren: It shouldn't, I tested back to `2.16.0`.

PR #862, opened 2022-03-08
The Dart plugin should use `.dart_tool/package_config.json`, instead of `.packages`

jwren:
https://youtrack.jetbrains.com/issue/WEB-54703

@alexander-doroshko

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/util/DotPackagesFileUtil.java
+++ b/Dart/src/com/jetbrains/lang/dart/util/DotPackagesFileUtil.java
@@ -1,6 +1,10 @@
 // Copyright 2000-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 package com.jetbrains.lang.dart.util;
 
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.util.Key;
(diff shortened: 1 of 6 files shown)


alexander-doroshko:
Merged, many thanks!

If you have time, it would be great to add some tests. We seem to have some for .packages (search for .packages in test source files), I guess we could write similar ones for package_config.json

Dart/src/com/jetbrains/lang/dart/util/DotPackagesFileUtil.java
```diff
@@ -15,15 +19,149 @@
...
+            jsonObjectPackage.get("rootUri") != null &&
+            jsonObjectPackage.get("packageUri") != null
+        ) {
+          final String packageName = jsonObjectPackage.get("name").getAsString();
+          final String rootUriValue = jsonObjectPackage.get("rootUri").getAsString();
+          final String packageUriValue = jsonObjectPackage.get("packageUri").getAsString();
+          // need to protect '+' chars because URLDecoder.decode replaces '+' with space
+          final String encodedUriWithoutPluses = StringUtil.replace(rootUriValue + "/" + packageUriValue, "+", "%2B");
+          final String uri = URLUtil.decode(encodedUriWithoutPluses);
+          final String packageUri = getAbsolutePackageRootPath(packageConfigJsonFile.getParent().getParent(), uri);
```
alexander-doroshko, line 154: @jwren Not sure `packageConfigJsonFile.getParent().getParent()` is correct here. See https://youtrack.jetbrains.com/issue/WEB-55363. I guess it should be `packageConfigJsonFile.getParent()`.

PR #865, opened 2022-04-22
Fix for Dart folding of longer methods and getters implemented with the fat arrow (=>) syntax. 

jwren:
Fixes issue WEB-35980

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
+++ b/Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
@@ -297,8 +297,16 @@ public class DartFoldingBuilder extends CustomFoldingBuilder implements DumbAwar
                                        @NotNull final PsiElement dartComponentOrOperatorDeclaration) {
     final DartFunctionBody functionBody = PsiTreeUtil.getChildOfType(dartComponentOrOperatorDeclaration, DartFunctionBody.class);
     final IDartBlock block = functionBody == null ? null : functionBody.getBlock();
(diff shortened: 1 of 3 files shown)


Dart/src/com/jetbrains/lang/dart/folding/DartFoldingBuilder.java
```diff
@@ -297,9 +297,17 @@ private static void foldFunctionBody(@NotNull final List<FoldingDescriptor> desc
                                        @NotNull final PsiElement dartComponentOrOperatorDeclaration) {
     final DartFunctionBody functionBody = PsiTreeUtil.getChildOfType(dartComponentOrOperatorDeclaration, DartFunctionBody.class);
     final IDartBlock block = functionBody == null ? null : functionBody.getBlock();
+    // Handle bodies defined with {...}
     if (block != null && block.getTextLength() > 2) {
       descriptors.add(new FoldingDescriptor(functionBody, block.getTextRange()));
     }
+    // Handle bodies defined with the fat arrow: =>
+    // A text length of 5 is used for cases as short at "=> 0;"
```
alexander-doroshko, line 305: Does it make sense to collapse single-line functions? Maybe collapse only multi-line ones (check for the "\n" symbol)?

jwren: Either choice would make sense, we currently fold long blocks even if they are on a single line, the current functionality keeps it consistent.

alexander-doroshko:
Thanks, pushed

PR #866, opened 2022-05-08
[Dart] Extended code folding (WEB-54993)

Gustl22:
see: https://youtrack.jetbrains.com/issue/WEB-54993/Dart-Support-Code-Folding-for-Arrays-Lists-anonymous-functions-f

Added code folding for:
- For-, While-, Do-While-Loops
- Anonymous functions / expression bodies
- Lists / Arrays
- Fix `asnyc` keyword in fat arrow functions

Removed code folding for:
- Tests with `group` and `test`. Explanation:
  - the test methods use [anonymous / lambda functions](https://dart.dev/guides/language/language-tour#anonymous-functions), which now can be folded anywhere not just in these specific test cases. So it wouldn't make sense to have both options, just to save two braces. I readded the "test"-tests, which are still valid to the `FunctionsExpressionBody` section.
  
Not yet implemented, but that's stuff for another PR, not this one:
- Declaration of signatures over multiple lines
- Call of signatures over multiple lines

Edit: Need to check
- [x] do while loops
- [x] function (expression) bodies with `async` keyword


jwren (approved):
Changes LGTM.  Thank you for the PR!  I have always been a fan of improved folding IntelliJ.

alexander-doroshko:
Thank you for the contribution and sorry for the delay. I've merged the PR. The commit will appear here shortly.

PR #867, opened 2022-05-22
Add Makefile location next to its name in files tree

fan-tom:
This PR adds file path next to the Makefile file name in makefiles panel
Before:
изображение
After:
изображение

This is useful when there are multiple makefiles with the same name in the same project, possibly even with the same set of actions

This PR also adds gradle build script to be able to build, run and test this plugin without need to have IDEA sources and copying file there, as described [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006411779/comments/360000859159)

To be able to run tests I modified `BASE_TEST_DATA_PATH` constant in `MakefileTestUtils.kt` to calculate it in similar way it is done in `handlebars` plugin


makefile/src/com/jetbrains/lang/makefile/toolWindow/MakefileFileNode.kt
```diff
@@ -32,4 +34,9 @@ class MakefileFileNode(val psiFile: PsiFile, private val targets: List<MakefileT
   override fun getIndex(node: TreeNode) = targets.indexOf(node)
 
   override fun getAllowsChildren() = true
+
+  override fun renderName(renderer: ColoredTreeCellRenderer) {
+    super.renderName(renderer)
+    renderer.append("  ${psiFile.virtualFile.path}", SimpleTextAttributes.GRAYED_ITALIC_ATTRIBUTES, false)
```
fan-tom, line 40: Is there any way to make this fragment right-aligned? I tried different values for `padding` of this and filename fragments together with `alignment = RIGHT` - no effect.

makefile/src/com/jetbrains/lang/makefile/toolWindow/MakefileTreeNode.kt
```diff
@@ -9,4 +11,17 @@ abstract class MakefileTreeNode(@NlsSafe val name: String) : TreeNode {
   object Comparator : kotlin.Comparator<MakefileTreeNode> {
     override fun compare(a: MakefileTreeNode, b: MakefileTreeNode): Int = a.name.compareTo(b.name, ignoreCase = true)
   }
+
+  protected open fun renderIcon(renderer: ColoredTreeCellRenderer) {
+    renderer.icon = this.icon
+  }
+
+  protected open fun renderName(renderer: ColoredTreeCellRenderer) {
+    renderer.append(name, SimpleTextAttributes.REGULAR_ATTRIBUTES, true/*, name.length + 10, SwingConstants.LEFT*/)
```
fan-tom, line 20:
```suggestion
    renderer.append(name, SimpleTextAttributes.REGULAR_ATTRIBUTES, true)
```

fan-tom:
Seems this PR is superceeded by 2ebe18dc7a51e5206f1f7085dab4982cf5b4a2fe

PR #868, opened 2022-06-03
Parse web format of files with 'packages/' 

helin24:
This addresses https://github.com/flutter/flutter-intellij/issues/6131. The flutter issue occurs because the stdout for flutter web differs from other platforms. We investigated whether we should change the stdout to be more consistent but decided it made more sense to parse the file names differently (see https://github.com/dart-lang/sdk/issues/49147).

CC @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/DartPositionInfo.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/DartPositionInfo.java
@@ -30,6 +30,8 @@ public class DartPositionInfo {
   public final int line;
   public final int column;
 
+  public static final String PACKAGE_WITH_SLASH = "packages/";
+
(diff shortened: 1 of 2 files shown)


jwren:
LGTM

alexander-doroshko:
@helin24 @jwren I'm awfully sorry, I somehow haven't noticed this PR earlier.
I've pushed it with minor tweaks (I used a trick that allowed to reuse existing method `parseUrlStartAndEnd()`).
Let me know if this PR is not needed anymore and I need to revert it.

helin24:
@alexander-doroshko no worries, can you post the link to your change?

alexander-doroshko:
The link has been posted by intellij-monorepo-bot. (It appears with up to 1 hour delay after a commit to the original repository.)

PR #869, opened 2022-06-04
Enable support for fields that end with `$` and `_`

andreasavg:
We are using a field naming methodology were our class fields end in '_'. Currently, the tapestry plugin only supports fields that start with a '$' or '_'. 

Extend the functionality so that components can pick-up fields that end with '_' and not report them as missing.


anstarovoyt:
Tapestry plugin was moved to obsolete https://github.com/JetBrains/intellij-obsolete-plugins/tree/master/tapestry

PR #870, opened 2022-06-23
WEB-41595 call hierarchy: jump to caller position

hpoul:
When showing caller hierarchy, jump to the actual location of the invocation, not to the enclosed function.

If there are multiple invocations inside a function all locations will be highlighted, similar to how the java plugin works.

Screen Shot 2022-06-24 at 00 08 00

For some reason I have a few test failures.. but when reverting to the original master, I have the same failures.. so I don't think my changes caused them. Can someone reproduce this? thanks.

https://youtrack.jetbrains.com/issue/WEB-41595/Dart-plugin-Call-hierarchy-should-navigate-to-call-side-not-surrounding-function


alexander-doroshko:
Thank you for the PR and sorry for a daly with the review.

First comment: would you agree that there's no need to change logic in 'callee' mode? I mean that clicking on `bar` in the tree should jump to `bar` declaration, not to the `bar()` call. This works the same way in Java and Kotlin. I think jumping to caller position makes sense only for a tree in 'caller mode'.

image

PR #871, opened 2022-06-24
Dart: allow quick fixes with multiple edits. and support .arb translation files.

hpoul:
When a analysis warning is produced with multiple fixes they were correctly displayed in the "Dart Analysis" tool window. But in the quick fix drop down only fixsets with a single fix were displayed.
This changes this behavior to allow multiple fixes. And it seems to work for me.

In addition i've added `.arb` files to the list of analysis files.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -892,6 +892,7 @@ public final class DartAnalysisServerService implements Disposable {
     return fileName.endsWith(".dart") ||
            fileName.endsWith(".htm") ||
            fileName.endsWith(".html") ||
+           fileName.endsWith(".arb") ||
            fileName.equals(".analysis_options") ||
            fileName.equals("analysis_options.yaml") ||
            fileName.equals("pubspec.yaml") ||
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
@hpoul I'm very sorry about the delay with your PR.
Can you give an example of a quick fix that has multiple SourceFileEdits?

Are arb files supported by the Dart Analysis Server? I haven't found it in the Dart SDK source code so far.

hpoul:
@alexander-doroshko well.. the dart analysis server supports plugins.. and I've written one to show quick fixes for externalizing strings.. which requires multiple sourceedits to

1. Create an every in the .arb fine
2. Change the .dart file to reference the name used in the arb file

See also https://github.com/dart-lang/sdk/issues/46722#issuecomment-1162860274

alexander-doroshko:
@hpoul Thanks. Allowing multiple source changes looks reasonable to me.

Is there a real need in adding `.arb` to the list of analyzed files?

hpoul:
@alexander-doroshko iirc the main reason for my change was that the assist list checks whether the assist intention is part of an analysable file, and only shows it when that's the case:

https://github.com/JetBrains/intellij-plugins/blob/c412ed4fe07162a6e8347effa64cdb027e6bd471/Dart/src/com/jetbrains/lang/dart/assists/DartQuickAssistIntention.java#L76

https://github.com/JetBrains/intellij-plugins/blob/c412ed4fe07162a6e8347effa64cdb027e6bd471/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java#L881-L900

so I added the `arb` file. But since `arb` files are now part of dart/flutter projects i guess it would make sense that they will also be analysed one day 🤔

alexander-doroshko:
I've merged the multi-SourceFileEdits part of the PR.

> assist list checks whether the assist intention is part of an analysable file, and only shows it when that's the case:

As I understand, only the file in the editor is checked using `isLocalAnalyzableFile()`. So it should work if your quick fix is registered for a `.dart` file, even if it includes changes in `.arb` file as well.

PR #874, opened 2022-08-03
Update error-prone.xml

Kers1990:
update error-prone version from 2.8.1 to 2.14.0


chashnikov:
I've updated error-prone to 2.15.0.

PR #875, opened 2022-10-05
Nicca3006 patch 1

Nicca3006:
When creating step definitions with Alt+Enter in intelliJ right now we won´t get any Exception so our tests get green even if no code ever was implemented. As the template is hard coded I was trying to fix this. Probably it´s not as easy as I hoped for, so everyone feel free to add to this PR.
Thanks a lot!


bartekpacia:
Hi @Nicca3006, sorry for a late response and thanks a lot for your contribution.

This problem is tracked in [IDEA-326113](https://youtrack.jetbrains.com/issue/IDEA-326113), and I recently fixed it, before I found this contribution of yours.

PR #878, opened 2022-11-08
Fix WEB-57881, Prettier plugin doesn't take into account overrides config

gertmuller88:
Addresses https://youtrack.jetbrains.com/issue/WEB-57881


alexander-doroshko:
Merged, thank you for the contribution.

PR #879, opened 2022-11-24
IDEA-306713: Upgrade dependencies for Error Prone plugin

rickie:
Upgrade to Error Prone 2.16, and fix Checker Framework Dataflow
dependency to use the correct artifact and version.

[IDEA-306713](https://youtrack.jetbrains.com/issue/IDEA-306713)


chashnikov:
I've merged your pull request to master, thank you for the contribution! I've also [updated](https://github.com/JetBrains/intellij-plugins/commit/b4012e5f7968691ff03e68c5a29ae7ab7dd1dfc1) the version to ensure that IDE's will redownload the JARs.

PR #882, opened 2022-12-22
IDEA-306713: Unable to use Error Prone Compiler

vkravets:
Fix of IDEA-306713: Unable to use 'Error Prone Compiler' due to NoClassDefFoundError

Incorrect dataflow artifact is using, should be used dataflow-errorprone, since 3.14 version


chashnikov:
I've updated error-prone to 2.18.0 and fixed dependency on dataflow artifact.

PR #886, opened 2023-03-09
Dart: Update VM service protocol

stevemessick:
This is the replacement for #885, which was on a bad fork.

@alexander-doroshko


stevemessick:
Initial testing looks good. I stepped through both new and old styles of adding breakpoints. The error conditions have not been tested, but the successful path works.

The third commit removes some commented-out code.

PR #887, opened 2023-03-09
Dart: View record fields in the debugger

stevemessick:
This depends on #886.

Debugging some Dart code from the records spec:
Screenshot 2023-03-09 at 1 38 33 PM

I see the indexes have been fixed to start at one rather than zero, which is good.

@alexander-doroshko 

Fixes https://github.com/flutter/flutter-intellij/issues/6652


stevemessick:
Ready for review.

PR #888, opened 2023-04-06
Update the generated Dart Analysis Server java spec files to version 1.34.0

jwren:


alexander-doroshko:
I guess this PR is work in progress so I'm not merging it so far.

jwren:
@alexander-doroshko please take a look, PR has been touched and rebased.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/RemoteAnalysisServerImpl.java
```diff
@@ -29,7 +29,15 @@
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.jetbrains.lang.dart.DartBundle;
```
alexander-doroshko, line 37: Is it possible to keep separation of Dart Analysis Server concepts and IntelliJ-specific implementation?

I have 2 ideas.

The easiest would probably be to handle requests from the server in `processNotification()` and pass them to `AnalysisServerListener`. We have `AnalysisServerListener` implementation in `DartAnalysisServerService`, it will need to handle requests from the server.

Alternatively, it might be possible to introduce some new interface like `AnalysisServerRequestsHandler`, and pass requests to it. The client (`DartAnalysisServerService`) will need to register and implement such handler.

jwren: Abstracting things out and refactoring isn't bad, but I didn't want to over-architect such a small contribution.

jwren:
To reproduce, I patched and built the Dart SDK with this change:

```
+++ b/pkg/analysis_server/lib/src/legacy_analysis_server.dart
@@ -766,6 +766,8 @@ class LegacyAnalysisServer extends AnalysisServer {
     for (var driver in driverMap.values) {
       driver.priorityFiles = files;
     }
+    showUserPrompt(MessageType.warning, "jwren test message.",
+        <String>["ONE", "TWO", "thr33"]);
   }
 

```

alexander-doroshko:
Sorry for the delay with the PR. Pushed with some updates. I think I found the right way of implementing requests from the server. Please double-check. See comments in code.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/RemoteAnalysisServerImpl.java
```diff
@@ -570,6 +578,12 @@ public void server_getVersion(GetVersionConsumer consumer) {
     sendRequestToServer(id, RequestUtilities.generateServerGetVersion(id), consumer);
   }
 
+  // This is overridden due to the generated methods from the Dart SDK, but the implementation (if implemented) would be in #processRequests.
+  @Override
+  public void server_openUrlRequest(String url) {
+    // do nothing.
+  }
```
alexander-doroshko, line 585: I declared this `RemoteAnalysisServerImpl` class abstract and removed empty `server_openUrlRequest` and `server_showMessageRequest` implementations. This way I had to implement them right in the Dart plugin, where I create and instance of `RemoteAnalysisServerImpl`. So it allowed me not to use any IntelliJ-specific classes here, but to delegate the IntelliJ-specific implementation right to the Dart plugin.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/RemoteAnalysisServerImpl.java
```diff
@@ -715,6 +744,59 @@ else if (event.equals(LAUNCH_DATA_NOTIFICATION_RESULTS)) {
     return true;
   }
 
+  /**
+   * Attempts to handle the given {@link JsonObject} as a request. Return {@code true} if it
+   * was handled, otherwise {@code false} is returned.
+   *
+   * @return {@code true} if it was handled, otherwise {@code false} is returned
+   */
+  private boolean processRequests(final JsonObject response, final String idString) throws Exception {
```
alexander-doroshko, line 753: So, I rewrote `processRequests()` so that it calls `server_openUrlRequest` or `server_showMessageRequest` as needed

Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/RemoteAnalysisServerImpl.java
```diff
@@ -732,6 +814,12 @@ private void processResponse(JsonObject response) throws Exception {
     synchronized (consumerMapLock) {
       consumer = consumerMap.get(idString);
     }
+
+    // handle requests by calling processRequests, return if this was a request input type
+    if(consumer == null && processRequests(response, idString)) {
```
alexander-doroshko, line 819: Checking consumer for null doesn't seem to be correct. Consumer is found in the map by "id". It's a map of requests that the client has sent to the server.

Spec says: "The ids used by the server are independent of the ids used by the client. In other words, the ids sent by the server might be equal to ids already sent from the client, but are unrelated to the client request with that same id."

I removed `consumer == null` and moved processRequests() 3 lines up, right after `String idString = idJsonPrimitive.getAsString()`

PR #892, opened 2023-05-15
Remove use of Guava types from PbFile and related APIs.

mathewinwood:
This API can (and is) used from other plugins. Since each plugin can bundle their own version of guava, using a guava type in the return value is unsafe and results in LinkageErrors at runtime.

Switch to Map<X, Collection<Y>> in place of MultiMap<X, Y> and update related code accordingly to avoid such issues.


NikitaKatkov:
merged with additional fixes

PR #893, opened 2023-06-08
feat: Clarify Dart's closing label by changing it to an End symbol.

jinlongchen:
The original closing label in Dart uses a comment which may not be clear to developers. To improve clarity, I propose changing the closing label to an End symbol. This will make it more easily recognizable and reduce confusion in code.


alexander-doroshko:
Thank you for the pull request. Can you please add the screenshots to the PR description with the suggested change (both light and dark themes)?

Is ⦿ a well-known "end symbol"? 

I'm not sure if all fonts support this symbol nicely.

I'm also thinking about a different approach of addressing the confusion. I mean making it look not as a comment but as an inlay hint. Example from Kotlin: 
image

As a bonus, inlay hints have context menu that allows disabling them.

alexander-doroshko:
We'll fix it eventually in the scope of [IDEA-289260](https://youtrack.jetbrains.com/issue/IDEA-289260)

PR #894, opened 2023-07-29
serial-monitor: Support LF line-ending mode and fix split-up line-endings

Crementif:
This commit adds LF ("\n") line ending support, completing coverage for all line endings ("\r", "\n", "\r\n").

It also provides a robust solution for line endings split across chunks (the issue pointed out by the todo comment that links to https://bitbucket.org/dmitry_cherkas/intellij-serial-monitor/issues/1), replacing the previous temporary fix. Since the bitbucket repo is offline I had to use https://github.com/ribasco/intellij-serial-monitor/commit/1414af04540c3e28019a270aefe89b1e74d3da86 to find the details about this issue.


Crementif:
Would it help to have a bug tracker entry about the bug that this PR merged?

PR #896, opened 2023-08-30
fix: boolean typo

0o001:


hurricup:
Oh, nvm seems no test data bound to it.

PR #900, opened 2023-11-17
[error prone] update error prone to 2.23.0 and fix dependencies

rickie:
One thing to point out, now in Error Prone the `3.34.0-eisop1` is used instead of `3.34.0`, see here:
https://github.com/google/error-prone/blob/v2.23.0/pom.xml#L38C23-L38C36.

This means it is now a slightly different artifact, which is reflected in the changes.


rickie:
@dmitry-avdeev @piotrtomiak Hi, can you take a look at this PR?

PR #904, opened 2024-01-27
Add Dart tooling daemon startup and test request

helin24:
Dart is adding a tooling daemon for facilitating communication between services, e.g. DevTools and analysis server. (see protocol: https://github.com/dart-lang/sdk/blob/main/pkg/dtd_impl/dtd_protocol.md)

This change will cause the Dart plugin to start up the tooling daemon on project open and opens a websocket for communication.


helin24:
@alexander-doroshko @jwren @CoderDake

jwren:
@helin24 This looks like a great start to getting the Dart Tooling Daemon wired up with the Dart Plugin.  I think having a separate service is the right call, but will we have all of the context/metadata to pass into the DTD to initialize it from the Dart Analysis Server Service?

helin24:
@jwren I'm not sure yet how this will be passed into analysis server (have we clarified how analysis server will register DTD / vice versa?). I was thinking we would need a completer or something to notify when the websocket is ready, and the analysis server can wait on that to request information from DTD.

helin24:
@alexander-doroshko I think this could use a review. This isn't being used yet, but the next step will be having the Dart plugin start DevTools server by running `dart devtools` so we can connect to DevTools and pass in the DTD URI without waiting for an app to start running.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -0,0 +1,161 @@
...
+          if (method == "streamNotify") {
+            val params = json["params"].asJsonObject
+            for (listener in java.util.ArrayList(listeners)) {
+              listener.received(params["streamId"].asString, json)
+            }
+          }
+
+          val id = json["id"].asString
+          val consumer = consumerMap.remove(id)
+          consumer?.received(json)
```
alexander-doroshko, line 95: should this  be in `else` block?

alexander-doroshko:
Thanks, merged. The commit will appear here in several hours.

The call to `DartToolingDaemonService.getInstance(project).startService()` is commented out for now in `DartStartupActivity`, just because it's an early work in progress.

I applied some cleanup to my taste. If you are not sure about my updates or just interested in detailed comments, I'm happy to iterate.

As a next step, I think we need to polish the workflow regarding starting/stopping the DTD.

Is there a more canonical way of disconnecting and stopping the process than `dtdProcessHandler.killProcess()`?

Ideally, all these scenarios should be handled reasonably:
- initially, the project has the Dart support disabled, then a user enables the Dart support at some moment
- initially, the wrong Dart SDK path is configured, a user fixes it manually at some point
- initially, the wrong Dart SDK path is configured, but at some point, the SDK gets downloaded, and the path becomes valid without any manual configuration
- a user disables Dart support manually while working with the project
- a user changes the path to the Dart SDK manually (e.g., to switch version)
- a user kills the DTD process manually from the Terminal
- failure to start DTD because of missing permissions
- etc.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Oh yes, you're right. Or return after the for loop ends and listeners have been notified of `streamNotify` events.

alexander-doroshko: Ok, pushing the fix

PR #906, opened 2024-02-07
Dart Macros lang feature: start to map absolute paths to uri paths, work started on navigation targets

jwren:
The `lsp.notification`, as well as caching of the macro generated, in-memory files, is added as well in this change

@alexander-doroshko

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/com/google/dart/server/AnalysisServerListener.java
+++ b/Dart/thirdPartySrc/analysisServer/com/google/dart/server/AnalysisServerListener.java
@@ -222,4 +222,6 @@ public interface AnalysisServerListener {
    * @param existingImports a map from imported library uri onto the names declared by the imported library.
    */
   void computedExistingImports(String file, Map<String, Map<String, Set<String>>> existingImports);
+
+  public void lspTextDocumentContentDidChange(String url);
 }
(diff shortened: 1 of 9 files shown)


jwren:
Touched this PR with the comments from @alexander-doroshko and @DanTup.  Implemented the non-static `isDartSdkVersionSufficientForDartMacroSupport` but it is currently unused as the callers here are in static methods in RequestUtilities.

Dart/thirdPartySrc/analysisServer/com/google/dart/server/AnalysisServerListener.java
```diff
@@ -222,4 +222,6 @@ public void computedImplemented(String file, List<ImplementedClass> implementedC
    * @param existingImports a map from imported library uri onto the names declared by the imported library.
    */
   void computedExistingImports(String file, Map<String, Map<String, Set<String>>> existingImports);
+
+  public void lspTextDocumentContentDidChange(String url);
```
alexander-doroshko, line 226: I'll add a `default` empty implementation to this method right to this interface. Otherwise, we'd get Flutter plugin compatibility issues because its `FlutterAnalysisServerListener implements AnalysisServerListener`

PR #908, opened 2024-02-23
Start DevTools with DTD info

helin24:
Both the Dart and Flutter plugins make use of Dart DevTools, and now that Dart DevTools will have more static tooling, it makes sense to start this on project open instead of associating it with a running application.

DTD now has a `--machine` mode which is a little easier to parse, so I converted to using this. The DTD address (if retrieved successfully) will be registered to DevTools.

Upcoming work:
- Convert Dart plugin app startup to pass in existing DevTools address instead of starting up a new instance of DevTools for each app run
- Convert Flutter plugin to use Dart plugin's DevTools instance instead of starting its own


helin24:
CC @alexander-doroshko @jwren

jwren (commented):
LGTM

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsInstance.kt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2020 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
```
alexander-doroshko, line 5:
```
// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
```

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsInstance.kt
```diff
@@ -0,0 +1,8 @@
...
+package io.flutter.run.daemon
```
alexander-doroshko, line 6: `package com.jetbrains.lang.dart.ide.devtools`?

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsInstance.kt
helin24: Sorry, I copied this uncarefully from the flutter plugin

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsInstance.kt
helin24: Done

helin24:
@alexander-doroshko this is ready for another review

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsService.kt
```diff
@@ -0,0 +1,87 @@
...
+    result.addParameter("devtools")
+    result.addParameter("--machine")
+    if (dtdUri != null) {
+      result.addParameter("--dtd-uri=$dtdUri")
+    }
+
+    processHandler = ColoredProcessHandler(result)
+    processHandler.addProcessListener(object : ProcessListener {
+      override fun onTextAvailable(event: ProcessEvent, outputType: Key<*>) {
+        val text = event.text.trim { it <= ' ' }
```
alexander-doroshko, line 45: Is this comment valid for this service as well?
```
// The first line of text is the command issued, which can be ignored.
```

helin24: Yes, also valid for the DevTools service.

alexander-doroshko: Ok, I'll add some `if` tomorrow, like in DTDService

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -46,16 +53,38 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
+        if (details == null) {
+          dtdInfoFuture.completeExceptionally(Exception("No tooling daemon details found"))
+          return
+        }
+
+        val uri = details["uri"].asString
+        val secret = details["trusted_client_secret"].asString
+        dtdProcessHandler.removeProcessListener(this)
+        dtdInfoFuture.complete(DtdInfo(uri, secret))
+        connectToDtdWebSocket(uri)
```
alexander-doroshko, line 87: In my experiments, `uri` looks like `ws://127.0.0.1:65323/`. And the `connectToDtdWebSocket()` implementation is
```
private fun connectToDtdWebSocket(address: String) {
    webSocket = WebSocket(URI("ws://$address/ws"))
  ...
```
So the `URI` passed to `WebSocket()` constructor is like `ws://ws://127.0.0.1:65323//ws`

Should I change `connectToDtdWebSocket()` implementation to 
```
    webSocket = WebSocket(URI("${address}ws"))
```
?

In this example I removed the `ws://` prefix and the slash after `${address}`. Otherwize, there were two schema prefixes and two slashes in the end. Is the proposed change safe or we should introduce more complex string manipulation logic?

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsService.kt
```diff
@@ -0,0 +1,87 @@
...
+    if (sdk == null || !isDartSdkVersionSufficient(sdk)) {
+      return null
+    }
+    val result = GeneralCommandLine().withWorkDirectory(sdk.homePath)
+    result.charset = StandardCharsets.UTF_8
+    result.exePath = FileUtil.toSystemDependentName("dart")
+    result.addParameter("devtools")
+    result.addParameter("--machine")
+    if (dtdUri != null) {
+      result.addParameter("--dtd-uri=$dtdUri")
```
alexander-doroshko, line 39: I'm getting an error `Could not find an option named "dtd-uri"`
Dart SDK 3.4.0-194.0.dev built 5 Mar 2024

Was this option introduced later?

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsService.kt
alexander-doroshko: done

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Oops, sorry about that! I know I ran into this somewhere and didn't realize I changed it in the wrong PR.

Dart/src/com/jetbrains/lang/dart/ide/devtools/DevToolsService.kt
helin24: This should be included, e.g. I have `Dart SDK version: 3.4.0-171.0.dev` and can run `dart devtools --machine --dtd-uri=something`. Are you also getting this error when you run the command on your command line?

alexander-doroshko: Got it.
The code that starts DevTools is
```
val result = GeneralCommandLine().withWorkDirectory(sdk.homePath)
result.exePath = FileUtil.toSystemDependentName("dart")
```
`sdk` is correct for me, it is 3.4.0-194.0.dev. But the exePath is just "dart", so `dart` executable is taken from `$PATH` (having `withWorkDirectory(sdk.homePath)` doesn't help). For me it is `/opt/homebrew/bin/dart` which is some old SDK version.

I fixed it by using 
```
commandLine.exePath = FileUtil.toSystemDependentName(DartSdkUtil.getDartExePath(sdk))
```
like in DartToolingDaemonService.


BTW, does process home directory matter (for DTD or DT)?

PR #910, opened 2024-02-29
Use DevTools URI from project service [draft]

helin24:
Depends on #908

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/runner/server/OpenDartObservatoryUrlAction.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/runner/server/OpenDartObservatoryUrlAction.java
@@ -9,6 +9,7 @@ import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.DumbAwareAction;
 import com.intellij.openapi.util.Computable;
 import com.jetbrains.lang.dart.DartBundle;
+import com.jetbrains.lang.dart.ide.devtools.DevToolsService;
 import icons.DartIcons;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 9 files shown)


Dart/src/com/jetbrains/lang/dart/ide/runner/server/OpenDartObservatoryUrlAction.java
```diff
@@ -34,20 +36,25 @@ public void setUrl(@NotNull final String url) {
...
-    e.getPresentation().setEnabled(myUrl != null && myIsApplicable.compute());
+    e.getPresentation().setEnabled(myWebsocketUrl != null && myIsApplicable.compute());
   }
 
   @Override
   public void actionPerformed(@NotNull final AnActionEvent e) {
-    if (myUrl != null) {
-      openUrlInChromeFamilyBrowser(myUrl);
+    if (myWebsocketUrl != null) {
+      final String url = e.getProject().getService(DevToolsService.class).getDevToolsInstance().getUrl() + "/?uri=" + myWebsocketUrl;
```
alexander-doroshko, line 56: the implementation of the `getDevToolsConsumer` is
```
devToolsFuture.get()
```
This means that if DevTools process has not been started for some reason (we discussed these reasons by email), this call freezes the IDE. 

My thoughts are:
- not to use Future in DevToolsService (BTW, I renamed this class to DartDevToolsService, I hope it's fine)
- getDevToolsInstance() to be nullable and return the current result instantly
- keep compatibility with older SDKs, for which we don't start DTD+DT, so I mean keep using old behavior of the OpenDartObservatoryUrlAction when the new behaviour is not applicable

BTW, maybe renale DevToolsInstance to DartDevToolsInfo? It's just a data class for host and port.

Dart/src/com/jetbrains/lang/dart/ide/runner/server/webdev/DartWebdevRunningState.java
```diff
@@ -38,6 +38,7 @@ public class DartWebdevRunningState extends CommandLineState {
 
   @NotNull protected DartWebdevParameters myDartWebdevParameters;
 
+  // TODO: Is this used?
```
alexander-doroshko, line 41: Hmm, looks like not. Suspicious.

alexander-doroshko:
Merged, thanks

PR #911, opened 2024-03-04
Add and additional null and type checks in Dart's BoundField class.

jwren:
@alexander-doroshko


alexander-doroshko:
So, `getDecl()` is nullable now? I found two usages of this method, both could throw NPE, so I updated them. Please review when the link to the commit appears here.

PR #912, opened 2024-03-25
Add copy DTD URI action

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/7242

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
+++ b/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
@@ -48,6 +48,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
   private lateinit var webSocket: WebSocket
   var webSocketReady = false
     private set
+  var uri: String? = null
 
   private var secret: String? = null
 
@@ -84,6 +85,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
   private fun onServiceStarted(uri: String?, secret: String?) {
     serviceRunning = true
     this.secret = secret
+    this.uri = uri
     DartDevToolsService.getInstance(project).startService(uri)
     uri?.let { connectToDtdWebSocket(it) }
   }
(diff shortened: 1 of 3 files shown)


jwren (commented):
LGTM

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -84,6 +85,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
   private fun onServiceStarted(uri: String?, secret: String?) {
     serviceRunning = true
     this.secret = secret
+    this.uri = uri
```
alexander-doroshko, line 88: Should we set `uri` back to `null` in `processTerminated()`, where `serviceRunning` becomes `false`?

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -48,6 +48,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
   private lateinit var webSocket: WebSocket
   var webSocketReady = false
     private set
+  var uri: String? = null
```
alexander-doroshko, line 51: I'll add `private set` to make sure that the value of this property can't be accidentally changed from elsewhere

alexander-doroshko:
merged

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Ah yes, that would make sense. Same for in `onClose`?

alexander-doroshko:
https://github.com/JetBrains/intellij-plugins/commit/3c0a035e4c2abfb416249edd434be0ea3446b58a

helin24:
@alexander-doroshko do you know what version of the Dart plugin this will end up being in or when it will be released?

PR #913, opened 2024-04-03
Add basic editions support to intellij protobuf plugin.

shaod2:
This updates the protobuf grammar to include the following changes in edition 2023:
- Addition of `edition` keyword
- Reserved fields and enum values are now specified as identifiers instead of string literals

This also adds some basic validation checks:
- Enforce that only supported editions are allowed
- Enforce reserved fields and enum values obey the file's syntax
- Prohibit the removed `optional`, `required`, and `group` keywords
- Prohibit the `features` option from being used outside editions

As an immediate follow up we will audit all the gencode -> proto utilities exposed by the plugin and at least add some editions tests.

Future changes could actually handle feature cascading and provide some of the additional checks that protoc will apply during builds.  For example, every feature specifies the type of descriptors it's allowed to be specified on, and certain combinations of features are prohibited.

### DIFF
--- a/protobuf/protoeditor-core/src/com/intellij/protobuf/bnf/proto.bnf
+++ b/protobuf/protoeditor-core/src/com/intellij/protobuf/bnf/proto.bnf
@@ -274,7 +274,8 @@ private meta BlockEntry ::= !'}' <<p>> { pin = 1 }
 /*
  * Syntax statement.
  */
-SyntaxStatement ::= syntax '=' StringValue ';'
+SyntaxType ::= syntax | edition
+SyntaxStatement ::= SyntaxType '=' StringValue ';'
 {
   pin = 1
   implements = [
@@ -282,6 +283,9 @@ SyntaxStatement ::= syntax '=' StringValue ';'
     'com.intellij.protobuf.lang.psi.PbStatement'
   ]
   mixin = 'com.intellij.protobuf.lang.psi.impl.PbSyntaxStatementMixin'
+  methods = [
+    syntaxType = 'SyntaxType'
+  ]
 }
 
 /*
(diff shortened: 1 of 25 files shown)


mkruskal-google:
@NikitaKatkov

protobuf/protoeditor-core/src/com/intellij/protobuf/bnf/proto.bnf
```diff
@@ -274,14 +274,18 @@ private meta BlockEntry ::= !'}' <<p>> { pin = 1 }
 /*
  * Syntax statement.
  */
-SyntaxStatement ::= syntax '=' StringValue ';'
+SyntaxType ::= syntax | edition
```
NikitaKatkov, line 277: Is it necessary to introduce a separate PSI element representing a keyword? If you don't mind, I'd rather make this rule private and make SyntaxStatement#getSyntaxType method return enum. Currently it would have two options: syntax, edition. WDYT?

protobuf/protoeditor-core/src/com/intellij/protobuf/lang/psi/PbSyntaxStatementBase.java
```diff
@@ -21,4 +21,7 @@ interface PbSyntaxStatementBase extends PbElement {
 
   @Nullable
   SyntaxLevel getSyntaxLevel();
+
+  @Nullable
+  String getEdition();
```
NikitaKatkov, line 26: The only thing I find a bit confusing here is that there is a `getEdition` method that might return nothing in case it is an old style file with syntax statement. What if we extend the SyntaxLevel enum so that it would be a sealed class with several implementations. Traditional ones would be Proto2 and Proto3, and the new would be Edition class with custom version text in it?

The idea is identical to what you've proposed, just a bit different implementation

protobuf/protoeditor-core/src/com/intellij/protobuf/bnf/proto.bnf
mkruskal-google: No, it shouldn't be necessary.  And making it private sounds good, since under editions nobody should really be checking syntax/edition.  Long-term, you'd want to look at the features to truly understand the files

protobuf/protoeditor-core/src/com/intellij/protobuf/lang/psi/PbSyntaxStatementBase.java
mkruskal-google: Yea that sgtm

PR #918, opened 2024-07-09
Add register service method for DTD

helin24:
This is for enabling the Flutter plugin to register a service method for navigating to code, something that can be requested by a user from Flutter DevTools. More info on service methods: https://github.com/dart-lang/sdk/blob/main/pkg/dtd_impl/dtd_protocol.md#service-methods

Example DTD interactions in logs (comments for clarifying two-client interaction):
```
INFO - #c.j.l.d.i.d.DartDevToolsService - Starting Dart DevTools, sdk 3.5.0-321.0.dev
INFO - #c.j.l.d.i.t.DartToolingDaemonService - Connected to DTD successfully
FINE - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"FileSystem.setIDEWorkspaceRoots","id":"1","params":{"roots":["file:///Users/helinx/IdeaProjects/hello_flutter_external/android","file:///Users/helinx/IdeaProjects/hello_flutter_external"],"secret":"qzgLfkLrkpov0FXC"}}
// client 1 registers service
FINE - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"registerService","id":"2","params":{"service":"Test","method":"testMethod"}}
FINE - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"type":"Success"},"id":"1"}
FINE - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"type":"Success"},"id":"2"}

// client 2 calls service method
FINE - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"Test.testMethod","id":"3","params":{"param1":"1"}}

// client 1 receives service method request
FINE - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","method":"Test.testMethod","id":0,"params":{"param1":"1"}}

// client 1 sends service method response
FINE - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","id":"0","result":{"success":true}}

// client 2 receives response
FINE - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"success":true},"id":"3"}
```

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
+++ b/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
@@ -60,6 +60,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
 
   private val nextRequestId = AtomicInteger()
   private val consumerMap: MutableMap<String, DartToolingDaemonConsumer> = mutableMapOf()
+  private val servicesMap: MutableMap<String, DartToolingDaemonServiceConsumer> = mutableMapOf()
 
   private val eventDispatcher: EventDispatcher<DartToolingDaemonListener> = EventDispatcher.create(DartToolingDaemonListener::class.java)
 
@@ -102,6 +103,24 @@ class DartToolingDaemonService private constructor(private val project: Project)
     }
     catch (e: Exception) {
       logger.error("Failed to connect to Dart Tooling Daemon, uri: $uri", e)
+    }
+  }
+
+  fun registerServiceMethod(service: String, method: String, consumer: DartToolingDaemonServiceConsumer) {
+    val params = JsonObject()
+    params.addProperty("service", service)
+    params.addProperty("method", method)
+    sendRequest("registerService", params, false) {response ->
+      val result = response.getAsJsonObject("result")
+      if (result == null) {
+        logger.error("No result from attempt to register service $service.$method", service, method)
+      }
+      val type = result.getAsJsonPrimitive("type")
+      if (type == null || "Success" != type.asString) {
+        logger.error("Failed to register service $service.$method", service, method)
+      }
+
+      servicesMap["$service.$method"] = consumer
     }
   }
 
@@ -237,6 +256,21 @@ class DartToolingDaemonService private constructor(private val project: Project)
       //  println("received response from streamListen")
       //  println(response)
       //}
+
+      // Example request to test registering a service method
+      registerServiceMethod("Test", "testMethod") {request ->
+        println(request)
+        val params = JsonObject()
+        params.addProperty("success", true)
+        params
+      }
+
+      // Try using the service method
+      val methodParams = JsonObject()
+      methodParams.addProperty("param1", "1")
+      sendRequest("Test.testMethod", methodParams, false) {response ->
+        println(response)
+      }
     }
 
     override fun onMessage(message: WebSocketMessage) {
@@ -256,6 +290,12 @@ class DartToolingDaemonService private constructor(private val project: Project)
         val params = json["params"].asJsonObject
         val streamId = params["streamId"].asString
         eventDispatcher.multicaster.received(streamId, json)
+      } else if (servicesMap.contains(method)) {
+        val params = json["params"].asJsonObject
+        val consumer = servicesMap.get(method)
+        val result = consumer?.received(params)
+        println(result)
+        // Send result back to DTD - how?
       }
       else {
         val id = json["id"].asString
--- a/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
+++ b/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
@@ -274,8 +274,8 @@ class DartToolingDaemonService private constructor(private val project: Project)
       //}
 
       // Example request to test registering a service method
-      registerServiceMethod("Test", "testMethod") {request ->
-        println(request)
+      registerServiceMethod("Test", "testMethod") {requestParams ->
+        println(requestParams)
         val params = JsonObject()
         params.addProperty("success", true)
         params
--- a/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
+++ b/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
@@ -147,6 +147,22 @@ class DartToolingDaemonService private constructor(private val project: Project)
     webSocket.send(requestString)
   }
 
+  fun sendResponse(id: String, result: JsonObject) {
+    if (!webSocketReady) {
+      logger.warn("sendResponse(\"$id\", $result) called when the socket is not ready")
+      return
+    }
+
+    val response = JsonObject()
+    response.addProperty("jsonrpc", "2.0")
+    response.addProperty("id", id)
+    response.add("result", result)
+
+    val responseString = response.toString()
+    logger.debug("--> $responseString")
+    webSocket.send(responseString)
+  }
+
   fun addToolingDaemonListener(listener: DartToolingDaemonListener, parentDisposable: Disposable) =
     eventDispatcher.addListener(listener, parentDisposable)
 
@@ -292,10 +308,14 @@ class DartToolingDaemonService private constructor(private val project: Project)
         eventDispatcher.multicaster.received(streamId, json)
       } else if (servicesMap.contains(method)) {
         val params = json["params"].asJsonObject
+        val id = json["id"].asString
         val consumer = servicesMap.get(method)
         val result = consumer?.received(params)
-        println(result)
-        // Send result back to DTD - how?
+        if (result == null) {
+          logger.warn("No response from request consumer for message: $text")
+        } else {
+          sendResponse(id, result)
+        }
       }
       else {
         val id = json["id"].asString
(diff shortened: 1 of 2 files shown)


helin24:
I'm not sure how the response gets sent back to DTD yet - so this is just an early draft for @alexander-doroshko to leave comments on (whenever convenient).

alexander-doroshko:
LGTM. Let me know when it's production-ready. 

Note that I gonna be offline between July 20 and August 11.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -237,6 +272,21 @@ class DartToolingDaemonService private constructor(private val project: Project)
       //  println("received response from streamListen")
       //  println(response)
       //}
+
+      // Example request to test registering a service method
```
helin24, line 276: This part should be commented out and only used if we want to test this out

helin24:
@alexander-doroshko I fixed this up and it's production ready now. I added an example of logs to the original comment, let me know if you have any questions!

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -105,6 +106,24 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
   }
 
+  fun registerServiceMethod(service: String, method: String, consumer: DartToolingDaemonServiceConsumer) {
+    val params = JsonObject()
+    params.addProperty("service", service)
+    params.addProperty("method", method)
+    sendRequest("registerService", params, false) {response ->
+      val result = response.getAsJsonObject("result")
+      if (result == null) {
+        logger.error("No result from attempt to register service $service.$method", service, method)
```
alexander-doroshko, line 116: I removed the second and the third arguments because both `$service` and `$method` are already in the message (the first argument)

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -105,6 +106,24 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
+      }
```
alexander-doroshko, line 117: Should there be `return@sendRequest` right after logging an error?

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -105,6 +106,24 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
+      val type = result.getAsJsonPrimitive("type")
+      if (type == null || "Success" != type.asString) {
+        logger.error("Failed to register service $service.$method", service, method)
```
alexander-doroshko, line 120: The same. I removed the second and the third arguments because both $service and $method are already in the message (the first argument)

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -105,6 +106,24 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
+      }
```
alexander-doroshko, line 121: The same. Should there be return@sendRequest right after logging an error?

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -128,6 +147,22 @@ class DartToolingDaemonService private constructor(private val project: Project)
     webSocket.send(requestString)
   }
 
+  fun sendResponse(id: String, result: JsonObject) {
```
alexander-doroshko, line 150: I made this function private. It is ok?

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
alexander-doroshko: Ok, pushed as commented out

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -256,6 +306,16 @@ class DartToolingDaemonService private constructor(private val project: Project)
...
+      } else if (servicesMap.contains(method)) {
+        val params = json["params"].asJsonObject
+        val id = json["id"].asString
+        val consumer = servicesMap.get(method)
+        val result = consumer?.received(params)
+        if (result == null) {
+          logger.warn("No response from request consumer for message: $text")
+        } else {
+          sendResponse(id, result)
+        }
```
alexander-doroshko, line 318: I rewrote it in a way that neither `consumer`, nor `result` are nullable inside the `else if` branch:
```
      val method = json["method"]?.asString
      val serviceConsumer = servicesMap[method]
      if (method == "streamNotify") {...}
      else if (serviceConsumer != null) {
        val params = json["params"].asJsonObject
        val id = json["id"].asString
        val result = serviceConsumer.received(params)
        sendResponse(id, result)
      }
      else {...}
```
Is it ok?

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonServiceConsumer.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2000-2024 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license.
+package com.jetbrains.lang.dart.ide.toolingDaemon
+
+import com.google.gson.JsonObject
+
+fun interface DartToolingDaemonServiceConsumer {
+  fun received(request: JsonObject): JsonObject
```
alexander-doroshko, line 7: Maybe rename `received` to something like `handleRequest`?

The implementations can't fail and always return valid not null `JsonObject`, right?

alexander-doroshko:
Pushed, thanks. Please have a look at the comments.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Yes, it would make sense to return after the error.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Yes, same here!

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Yes it should only be used in this class

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
helin24: Looks great! Thanks

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonServiceConsumer.kt
helin24: 1. `handleRequest` sounds better! change to `RequestHandler` class name also fine
2. Yes, if there is an error the resulting `JsonObject` should reflect that.

helin24:
> One more thought. Can it become a problem that `DartToolingDaemonServiceConsumer#received` is called synchronously in `DtdWebSocketEventHandler#onMessage`?
> 
> As I understand, messages processing by DtdWebSocketEventHandler is blocked until `DartToolingDaemonServiceConsumer` responds. OTOH, if all `DartToolingDaemonServiceConsumer` impolementations are going to respond instantly, it shouldn't cause issues.

Hmmm... good point. I don't think we should assume that all implementations would respond instantly. What are implications for making this async? Sending the response to DTD later should be reasonable.

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
alexander-doroshko: done

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonServiceConsumer.kt
alexander-doroshko: done

alexander-doroshko:
> Sending the response to DTD later should be reasonable.

Ok, I moved potentially long calculations to a different thread:
```
      else if (serviceConsumer != null) {
        val params = json["params"].asJsonObject
        val id = json["id"].asString
        ApplicationManager.getApplication().executeOnPooledThread {
          val result = serviceConsumer.handleRequest(params)
          sendResponse(id, result)
        }
      }
```

I didn't test but I hope it works.

BTW, we may want to review the whole class implementation, how ready it is for concurrent multi-threaded usage

alexander-doroshko:
> how to test this

I'm not sure about real testing. But we can carefully analyze all the usages of the shared data (DartToolingDaemonService class fields) and think what would happen if many parallel threads executed the same (where applicable) or different functions of this class and read/wrote the shared data in the most unlucky moments

PR #920, opened 2024-08-22
Add capabilities and modified error response for DTD service methods

helin24:
We did some work to standardize how DTD service methods and their responses should look in: https://github.com/dart-lang/sdk/blob/main/pkg/dtd_impl/dtd_common_services.md#responses

This change is to make the Dart plugin compatible with those changes when registering service methods.

Oddly, it seems I get a `data` section (with original request info) in the response, but I'm not sure where that's being added in the code:

```
{
  "jsonrpc": "2.0",
  "error": {
    "code": 144,
    "message": "This is an error",
    "data": {
      "request": {
        "jsonrpc": "2.0",
        "method": "Test.testErrorMethod",
        "id": "5",
        "params": {
          "param1": "1"
        }
      }
    }
  },
  "id": "5"
}
```


helin24:
@alexander-doroshko

PR #921, opened 2024-09-14
Dart plugin config fix

Lootwig:
I'm not really sure if there's another intended way how to get this project to build that I'm missing, but following the current instructions alone fails during the build step.

After adding the plugin to classpath, the IDE still marks the overrides as invalid - although auto-generating them results in identical signatures, so I guess it's just related to where the types referenced in each are defined.

### DIFF
--- a/Dart/Dart-community.iml
+++ b/Dart/Dart-community.iml
@@ -69,4 +69,6 @@
       </library>
     </orderEntry>
   </component>
+  <orderEntry type="module" module-name="intellij.javascript.debugger" />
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
Thank your for the PR. I'll either merge it or find a different way to fix compilation.

Meanwhile, the easiest way to fix compilation is to exclude `DartWebDebugAware` and `DartFileUrlMapper` from Compile:
image

alexander-doroshko:
README updated in https://github.com/JetBrains/intellij-plugins/commit/101edb4dd5522a559af8a2b67cebb5226ac677b9, so closing

PR #923, opened 2024-10-03
Allow using "@" (at sign) in Gherkin step names

zenovich:
This will allow steps like 'And @Group is a child of @ParentGroup' for those who use such custom syntax for variables in tests.
Note: It doesn't contradict with Gherkin specifications as Gherkin tags (which have the similar syntax) are not allowed inside test steps.


zenovich:
Hi @bartekpacia,
The corresponding ticket in YouTrack: https://youtrack.jetbrains.com/issue/IDEA-362706/
By the "custom syntax for variables in tests" I meant using custom (non-standard) variables in cucumber tests starting with '@' sign. We have some in our test system (for example, [here](https://github.com/France-ioi/AlgoreaBackend/blob/master/app/api/threads/list_threads.feature#L13)).

bartekpacia (approved):
lgtm, started a merge request internally.

this PR will be closed automatically once it lands.

PR #925, opened 2024-10-08
[Dart] Update spelling dictionary for Dart related terms

parlough:
Adds some more words related to Dart or surrounding tooling that I've found popping up in Dart projects as not being recognized by IntelliJ.

Also removes some terms that are outdated or not that Dart specific.


jwren:

alexander-doroshko:
`overriden` looks like a real typo in the list :)

Some other words don't look like Dart-related. They are good to have in a personal dictionary but I'm not sure about shipping them right with the Dart plugin. (Truly, some of the already added words also don't seem to be Dart-specific; maybe I should clean them up eventually.)

Would you agree to keep only Dart/Flutter-related words in this PR?

parlough:
> Would you agree to keep only Dart/Flutter-related words in this PR?

Sure! I went ahead and removed a few outdated terms and some terms too broad for just a Dart dictionary. Let me know what you think.

If I wanted to propose some additions to the platform's default dictionary, do you know where I could do that?

alexander-doroshko:
Cool, thanks, merged.

> If I wanted to propose some additions to the platform's default dictionary, do you know where I could do that?

Maybe this one: https://github.com/JetBrains/intellij-community/blob/master/spellchecker/resources/com/intellij/spellchecker/jetbrains.dic

PR #926, opened 2024-10-09
[Dart] Use Object.hash and Object.hashAll if supported

parlough:
Resolves https://youtrack.jetbrains.com/issue/IDEA-249561

\cc @jwren

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateEqualsAndHashcodeFix.java
+++ b/Dart/src/com/jetbrains/lang/dart/ide/generation/CreateEqualsAndHashcodeFix.java
@@ -6,6 +6,7 @@ import com.intellij.codeInsight.template.TemplateManager;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.NlsContexts;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiElement;
 import com.jetbrains.lang.dart.DartBundle;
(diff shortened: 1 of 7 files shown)


jwren:

Dart/src/com/jetbrains/lang/dart/ide/generation/CreateEqualsAndHashcodeFix.java
```diff
@@ -32,6 +35,9 @@ public CreateEqualsAndHashcodeFix(@NotNull final DartClass dartClass) {
   public void beforeInvoke(@NotNull Project project, Editor editor, PsiElement file) {
     super.beforeInvoke(project, editor, file);
     mySuperclassOverridesEqualEqualAndHashCode = doesSuperclassOverrideEqualEqualAndHashCode(myDartClass);
+
+    final DartSdk sdk = DartSdk.getDartSdk(project);
+    supportsObjectHashMethods = sdk != null && StringUtil.compareVersionNumbers(sdk.getVersion(), "2.14") >= 0;
```
alexander-doroshko, line 40: I think it's safe to assume a fresh Dart SDK by default so I changed it to 
```
sdk == null || StringUtil.compareVersionNumbers(sdk.getVersion(), "2.14") >= 0
```
It also simplifies getting DartGenerateActionTest to pass, sdk is null in it

Dart/testData/generate/EqualsAndHashCode20.txt
```diff
@@ -0,0 +1,53 @@
...
+  int? n;
+  int? o;
+  int? p;
+  int? q;
+  int? r;
+  int? s;
+  int? t;
+
+  @override
+  bool operator ==(Object other) =>
```
alexander-doroshko, line 25: Formatting result is a bit different to me in this and in the ...21.txt file. It's because `dart format` doesn't run in this test after code generation. I updated the expected test data to make the test green. In real life `dart format` does run and the result is different.

alexander-doroshko:
Merged, thank you for the contributions!

PR #927, opened 2024-10-09
[Dart] Add 'Sort Members in Dart Code' to project view action menu

parlough:
Working on Dart projects that rely on member sorting, this enables much quicker access to the sort member functionality from the project view's action menu.

This PR also adjusts the text to "Sort Members in Dart code" rather than files as it can show up when selecting one or many files. It also aligns better with the the formatting action's text.

\cc @jwren


jwren:

alexander-doroshko:
I'll take a pause to think how natural the `Sort Members` action feels in the already overloaded Project View context menu...

By the way, it's probably a good time to remove `Reformat Code with 'dart format'` action from all action groups. It's currently in 3. These 3 groups already contain `Reformat Code`, which is effectively the same as `Reformat Code with 'dart format'`. Would you agree?

By the way, you can configure actions in the Project View context menu (and in all other menus and toolbars). Just add zSort Members` there in Settings | Appearance & Behavior | Menus and Toolbars

PR #929, opened 2024-10-14
[Dart] Remove some now unnecessary usages of guava methods

parlough:
Some of the usages were unused completely, so I removed them instead of replacing them.

While some files in this directory are generated, the ones updated here are not.

### DIFF
--- a/Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/BroadcastAnalysisServerListener.java
+++ b/Dart/thirdPartySrc/analysisServer/com/google/dart/server/internal/BroadcastAnalysisServerListener.java
@@ -14,7 +14,6 @@
 
 package com.google.dart.server.internal;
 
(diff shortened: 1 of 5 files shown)


parlough:
> Some files in Dart/thirdPartySrc/analysisServer are generated (and use guava even where it doesn't make much sense). Are the files that you touched generated or hand-written?

The files I updated in this PR aren't generated. I will separately update the generator in the SDK to avoid the usages in the generated files.

> I wonder if makes sense to cleanup the code in this folder?

I think it does. Originally it was copied, but now as the only client I think it makes sense to start stripping away what's not actually used, so it's easier to determine where changes are needed.

alexander-doroshko:
Thanks, merged.

The bot will close this PR in a few hours with a link to the commit.

Sorry for the delay with the other PRs. I'll eventually handle them all.

PR #930, opened 2024-10-16
Update i18n.json

martijn1985:
Replaced the dutch translation of Rule to Regel. This was updated in gherkin version 27 (https://github.com/cucumber/gherkin/releases/tag/v27.0.0)


jvsanten:
Would be nice to have this fix merged - still running into red highlighting when using Regel (and cucumber breaks when falling back to 'Rule' with Gherkin >= 27)

PR #931, opened 2024-11-20
[Dart] Increase min SDK to last minor Dart 1 release

parlough:
On the road to explicitly requiring Dart 2.

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/Dart/src/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -79,7 +79,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
 public final class DartAnalysisServerService implements Disposable {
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
Thanks,
I'm in contact with @jwren on how to increase it and when

jwren:
Thanks @parlough and @alexander-doroshko.  At the start of the year we will have the freedom and flexibility to remove versions before Dart SDK `3.0.0` (Flutter SDK version `3.10.0`, mapping here: https://docs.flutter.dev/release/archive).

Change in the Flutter Plugin for IntelliJ nudging users who are on such an old version: https://github.com/flutter/flutter-intellij/commit/451654ba00b11132bf78e47fb450f4224ead7904

@itsjustkevin

parlough:
Thank you both! Now that the Flutter notification and policy have been out for a bit, I went ahead and opened https://github.com/JetBrains/intellij-plugins/pull/942 instead of this PR. It takes a bit of a more conservative approach for now while still getting many of the benefits.

PR #932, opened 2024-11-22
adds `hashicorp/assert` to list of official providers

ksatirli:
Hey JetBrains folks 👋🏼 

Kerim here, big fan of the Terraform plugin inside GoLand.

I wanted to update this list with the Assert provider @bschaatsbergen and others maintain at https://registry.terraform.io/providers/hashicorp/assert/latest.

Hope this PR makes sense!


bschaatsbergen:
Thanks for this @ksatirli. Would be great if we could add this provider to the IntelliJ-HCL plugin 👍🏼.

batya239:
Thanks for contribution, unfortunately it's already added.

PR #933, opened 2024-11-23
[WIP] [Dart] Update the pub inspector to look up the directory structure to support the Dart Analysis workspaces feature

jwren:
Documentation to be added before this is ready to land.


jwren:
Thanks @alexander-doroshko.

> @jwren Thanks for the PR. Is [this document](https://docs.google.com/document/d/1UEEAGdWIgVf0X7o8WPQCPmL4LSzaGCfJTM0pURoDLfE) the single source of truth or there are other sources of information about how it works?

Yes, I believe so.

> 
> > [WIP]
> 
> So should I apply it as is or wait for updates?

I added some comments, yes, land it and push it to older IDEA platform versions if possible.  I am still able to push updated Flutter IJ versions to 2023.3, but 2024.1 and forward is sufficient.

> 
> BTW, do you think it's important for the Dart plugin to look into the `pubspec.yaml` file contents and have different logic depending on the `resolution` and `workspace` tags presence?

I think it is fine as-is, while we could possibly not go up a directory speeding up the check, I'd rather not be in parsing the yaml files, the fewer number of checks (existence of files) is deep enough in the weeds of what `pub` is doing.


> 
> Are there other features in the Dart plugin that should be aware of the workspaces feature?
> 
> For example, should we keep showing the bar with `Pub Get`/`Pub Upgrade`/`Pub Outdated` actions in the `pubspec.yaml` files with the `resolution` tag?

Possibly, something like this would be a reason to read and parse the yaml files.

I will take and action item to see if we are missing anything, need to add anything, etc.  (But lets not wait on landing this in the mean-time.)

alexander-doroshko:
Merged, thanks

gabrielgarciagava:
Is this part of latest version (dec 2nd)? I still face the issue so maybe it was not released yet and when can we expect it to be released?

alexander-doroshko:
@gabrielgarciagava Yes, the Dart plugin version 243.23177 includes this commit.

gabrielgarciagava:
I'm not sure why, but the latest available for me is 242.24931

I use IntelliJ IDEA 2024.2.4 (Community Edition)

alexander-doroshko:
The latest IntelliJ IDEA version is 2024.3.1.1.

However, the Dart plugin build 242.24931 for IntelliJ IDEA 2024.2.x also includes this commit.

PR #937, opened 2025-02-20
[deno] update plugin icon

KotlinIsland:
as we can see from: https://github.com/denoland/docs/pull/957/files there is no dark/light variant


anstarovoyt:
@web-devel

PR #938, opened 2025-03-11
[Dart] Send dart/connectToDtd request to the Analysis Server

elliette:
Fixes https://github.com/flutter/flutter-intellij/issues/7669

`dart/connectToDtd` request documented here: https://github.com/dart-lang/sdk/blob/bc9def5da7804243a9cce531e3512ab94a569fcd/pkg/analysis_server/tool/lsp_spec/README.md?plain=1#L280

### DIFF
--- a/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
+++ b/Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
@@ -68,7 +68,7 @@ class DartToolingDaemonService private constructor(private val project: Project)
 
   private val eventDispatcher: EventDispatcher<DartToolingDaemonListener> = EventDispatcher.create(DartToolingDaemonListener::class.java)
 
-  val analysisServerScope = DartAnalysisServerService.getInstance(project).serviceScope
+  private val analysisServerScope = DartAnalysisServerService.getInstance(project).serviceScope
 
   @Throws(ExecutionException::class)
   fun startService() {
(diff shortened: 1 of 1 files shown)


elliette:
FYI @alexander-doroshko @jwren

Dart/src/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -93,7 +96,12 @@ class DartToolingDaemonService private constructor(private val project: Project)
     this.uri = uri
     this.secret = secret
     DartDevToolsService.getInstance(project).startService(uri)
-    uri?.let { connectToDtdWebSocket(it) }
+    uri?.let { it ->
+      connectToDtdWebSocket(it)
+      CoroutineScope(Dispatchers.IO).launch {
```
elliette, line 101: I'm not sure this is the right way to call a `suspend` function

alexander-doroshko:
According to the [spec](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md?plain=1#L280), `dart/connectToDtd` is a request that the IDE should send to the Dart Analysis Server. This PR tries to send it to the Dart Tooling Daemon. I guess DTD doesn't need to connect to DTD :).

The scope of this subtask is to implement `DartAnalysisServerService.connectToDtd(String url)` method. It will be similar to `lspMessage_dart_textDocumentContent()` and it will require DAS protocol files update (`com.google.dart.server.generated.AnalysisServer` and friends). I think @jaime can help with it.

PR #942, opened 2025-06-28
[Dart] Require Dart 2.12 or later

parlough:
Increase the minimum supported Dart SDK to 2.12, which is the release which added support for null safety and finalized the consolidation on the combined `dart` CLI tool. This simplifies plugin development as a lot of tooling changed before the 2.12 release, requiring a lot of conditional behavior that is relatively hard to maintain.

---

### Justification and background

Dart 2.12 [was released on 2021-03-04](https://github.com/dart-lang/sdk/releases/tag/2.12.0), so over 4 years ago now. With the release of Dart 3 (May 2023), the Dart team stopped supporting language versions before 2.12 and [highlighted most packages support null safety](https://medium.com/dartlang/announcing-dart-3-53f065a10635#5b2a).

Beyond this, the Flutter IntelliJ plugin has implemented a [support policy](https://github.com/flutter/flutter-intellij?tab=readme-ov-file#flutter-sdk-compatibility) only supporting SDKs from the previous two years. As of the most recent release, they only support [Flutter 3.13 or later](https://github.com/flutter/flutter-intellij/blob/dd99416eed7cceea032acfa730d94fee00ecabc5/flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java#L30), which comes with Dart 3.1.

In this PR I set 2.12 as the minimum instead of 3.0 or 3.1 for two reasons:

- Incrementally migrating to an increased minimum, for easier review, fewer opportunities for regressions, and an additional notice period for developers.
- Continue supporting the few Dart developers that might not be running their apps with sound null safety, therefore needing an SDK before 3.0. I expect the minimum SDK version can later be increased to 2.19 while still supporting this group, but eventually support for SDKs before 3.0 should be dropped as well.

### Included changes

- Increased the minimum SDK version to "2.12.0".
- Removed checks for versions before 2.12 and if there was one, removed the branch for earlier versions.
- Removed support for the long-removed pub serve and `pub build` command, in favor of webdev.
  - Some related files, classes, methods, etc. should eventually be renamed due to this cleanup.
- Removed support for stagehand for project templates, only supporting `dart create`.
  - Some related files, classes, methods, etc. should eventually be renamed due to this cleanup.
- Consolidated the "asserts enabled or checked mode" setting to just "asserts enabled" now that Dart 1 is no longer supported.
- Remove messages from the bundle that are no longer used due to removed messages and forms.

### DIFF
--- a/Dart/resources/messages/DartBundle.properties
+++ b/Dart/resources/messages/DartBundle.properties
@@ -109,7 +109,6 @@ runner.web.app.configuration.name=Dart Web App
 command.line.run.config.label.dart.file=Dart &file:
 command.line.run.config.label.vm.options=&VM options:
 command.line.run.config.checkbox.enable.asserts=E&nable asserts
-command.line.run.config.checkbox.checked.mode=Checked mode
 command.line.run.config.label.program.arguments=Program a&rguments:
 command.line.run.config.label.working.directory=&Working directory:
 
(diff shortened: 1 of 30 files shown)


pq (commented):
This is fantastic. (I don't have access to approve but LGTM!)

PR #943, opened 2025-07-24
[clion, platformio] Update the main.c to compliment with the latest ESP-IDF

harriebird:
This changes the function name `main` to `app_main` in the initial `main.c` file for ESP-IDF projects.

This solves the `undefined reference to 'app_main'` issue as shown in the screenshot below:

app-main-error


elmot:
I don't think  this is a good solution. 
missing `main()` will break other types of projects.

PR #950, opened 2026-04-30
IDEA-306313 Java. Tests. Cucumber: Treat DataTable cells as strings

nlbuescher:
This fixes a bug where DataTable cells are treated as identifiers by the analyzer, which would in turn make IntelliJ warn about using non-ascii characters in cell values.

The non-ascii character inspection is configured to warn on identifiers by default.


ammarhantash (commented):
seems ok

nlbuescher:
Fixed in f94c94c277339e3023842c5e8fdf7d955ba14472

